PR #38, opened 2016-04-26
Arrange doc menu, header, contents with flexbox

begriffs:
Pages such as [What is Citus](https://www.citusdata.com/docs/citus/5.0/aboutcitus/what_is_citus.html) currently get cropped:

screen shot 2016-04-26 at 2 29 43 pm
^^^ *Menu just stops*

This pull request replaces complicated CSS positioning with flexbox.

@lfittl you're a CSS ninja, can you review my changes?


lfittl:
@begriffs If it works in all browsers you can easily test with, I'd say ship it - looks good from a quick glance at least :+1:

begriffs:
Good idea.

Chrome and Firefox good...Safari is horrible though.

lfittl:
You might still need vendor prefixes for Safari 8: http://caniuse.com/#feat=flexbox

lfittl:
Also not sure if you have the `display: flex` in all places you need it (anytime you use a `flex: ...` command, afaik)

begriffs:
OK that fixes Safari. Tried a new IE out on Browerstack and that looks good as well.

craigkerstiens:
@begriffs Mobile view seems to mangle things pretty badly on this one - 

citus_documentation_ _citus_5_0_0_documentation

begriffs:
Good catch.

begriffs:
Now it works in mobile except for a weird flicker in Safari at the media breakpoint width. Still have to look into that.

begriffs:
Resizing works perfectly in Safari now, while Chrome has a problem where at certain sizes the documentation body overflows its container. Might be a `box-sizing: border-box` and flexbox interaction.

begriffs:
I discovered the overflowing doc width happens only on pages which have code blocks.

begriffs:
The `pre` elements have `overflow: auto` so I would have thought they would shrink and add scrollbars. I tried [a](https://stackoverflow.com/questions/14299495/pre-tag-width100-or-max-width100-overflowauto-is-not-flexible) [few](https://stackoverflow.com/questions/10374171/how-to-make-twitter-bootstraps-pre-blocks-scroll-horizontally) [things](http://codepen.io/aj-foster/pen/emBYPW) but there's something complicating going on with our layout. I found that if I allow the words to wrap in the code blocks then it allows the documentation body to shrink.

craigkerstiens:
👍

begriffs:
I actually discovered more cases when the documentation body is unable to shrink properly. Found an entirely different and simpler way to fix the menu issue without using flexbox. Closing this PR and opening a new one.

PR #55, opened 2016-05-11
Restructure tutorials and installation instructions

begriffs:
Changes:
* Move tutorials earlier in TOC to encourage their use
* Require the `try-citus` bundle for tutorials (now called `citus-tutorial` bundle)
  * No more flowchart for differing kinds of installations
* Change terminology to "Single-Machine and "Multi-Machine" rather than development and production
* Use native Citus packages for Single-Machine installs rather than `try-citus`

One thing I'm not entirely happy with is the instruction in the single-machine linux installs to change permissions

```bash
sudo chmod a+w /var/run/postgresql
```

It allows a regular (non-postgres) user to start postgres servers and bind to a unix domain socket. This workaround probably weakens postgres security on the system. There are alternative approaches but I didn't like them as much.

1. Init the databases and execute them as the `postgres` user. It complicates the instruction command snippets because you have to indicate which user should run which bash script. It also makes the db logs and files less accessible to the regular unix user.
2. Change the `unix_socket_directories` postgres configuration variable to point at `/tmp`. However the package-built psql is expecting `/var/run/postgres` and requires an extra `-h /tmp` param for every call. This makes our instructions more complicated.
3. Add the current user to the `postgres` group. I'm not sure what side effects this might have for someone so I'm wary.

About branching strategy, I'll wait to create a 5.0 branch until this PR is merged because it's a change that should be in 5.0 as well as 5.1


sumedhpathak:
I ran through the instructions and they look good!
There are some points to discuss tomorrow. I think we should be able to get this merged tomorrow.

1. The /var/run issue. I'm OK with it as-is, but it's a bit clunky like you noted. Let's chat what we the options look like tomorrow.
2. We are calling it "Start Demo Cluster". Is this the best naming?
3. 'Machine' vs 'Node' naming: We've modified the text to 'machine', however the file names still have the 'node' moniker. 

Finally minor but, the OS X tutorial uses Postgres 9.5.0, but the Linux one uses 9.5.1. I'll add this as a note in the rebuilding issue.

begriffs:
I also noticed that the Citus readme quickstart tells people to create a single-machine docker cluster and then to continue to the tutorials. This will have to be addressed once this PR is merged and the new docs deployed.

begriffs:
Points 1 2 and 3 are addressed, so I'll merge this.

PR #59, opened 2016-05-12
Remove \STAGE and copy_from_distributed_table from scaling data ingestion guide

### LINKED ISSUE #52: Update Scaling Data ingest section to reflect new COPY changes
We now have COPY for both hash and append, so we should modify this section to remove copy_to_distributed_table, and any other relevant changes.

marcocitus:
Fixes #52

### DIFF
--- a/performance/scaling_data_ingestion.rst
+++ b/performance/scaling_data_ingestion.rst
@@ -10,7 +10,7 @@ The best method to distribute tables and ingest your data depends on your use ca
 Hash Distributed Tables
 $$$$$$$$$$$$$$$$$$$$$$$
 
-Hash distributed tables support ingestion using standard single row INSERT and UPDATE commands. This sub-section describes how you maximize insert throughput for hash distributed tables.
+Hash distributed tables support ingestion using standard single row INSERT and UPDATE commands, as well as bulk ingestion through COPY.
 
 Real-time Inserts (0-50k/s)
 ---------------------------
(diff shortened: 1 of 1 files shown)


performance/scaling_data_ingestion.rst
```diff
@@ -72,12 +54,34 @@ Note that this query also takes an exclusive lock on the row in PostgreSQL, whic
 
 When the replication factor is 1, it is always safe to enable citus.all_modifications_commutative. Citus does not do this automatically yet.
 
+Bulk Copy (100-200k/s)
+----------------------
+
+Hash distributed tables support `COPY <http://www.postgresql.org/docs/9.5/static/sql-copy.html>`_ from the Citus master for bulk ingestion, which can achieve much higher ingestion rates than regular INSERT statements.
```
sumedhpathak: Minor. Maybe use 'current' link? http://www.postgresql.org/docs/current/static/sql-copy.html

performance/scaling_data_ingestion.rst
```diff
@@ -72,12 +54,34 @@ Note that this query also takes an exclusive lock on the row in PostgreSQL, whic
...
+In psql, the \COPY command can be used to load data from the local machine. The \COPY command actually sends a COPY .. FROM STDIN command to the server before sending the local data, as would an application that loads data directly.
+
+::
+
+    psql -c "\\COPY counters FROM 'counters-20160304.csv' (FORMAT CSV)"
+
+
+A very powerful feature of COPY for hash distributed tables is that it asynchronously copies data to the workers over many parallel connections, one for each shard placement. This means that data can be ingested using multiple workers and multiple cores in parallel. Especially when there are expensive indexes such as a GIN, this can lead to major performance boosts over ingesting into a regular PostgreSQL table.
+
+To avoid opening too many connections to the workers. We recommend only running only one COPY command on a hash distributed table at a time. In practice, running more than two at a time rarely results in performance benefits. An exception is when all the data in the ingested file has a specific partition key value, which goes into a single shard. COPY will only open connections to shards when necessary.
```
sumedhpathak: Should we highlight this as a "Note" if we deem it important? Will this result in shards being marked as inactive?

performance/scaling_data_ingestion.rst
```diff
@@ -259,13 +261,13 @@ To load the data, we can download the data, decompress it, filter out unsupporte
...
     CREATE TEMPORARY TABLE prepare_1 (data jsonb);
     
-    /* Load a file directly from Github archive and filter out rows with unescaped 0-bytes */
+    -- Load a file directly from Github archive and filter out rows with unescaped 0-bytes
     COPY prepare_1 FROM PROGRAM
     'curl -s http://data.githubarchive.org/2016-01-01-15.json.gz | zcat | grep -v "\\u0000"'
     CSV QUOTE e'\x01' DELIMITER e'\x02';
     
-    /* Prepare a staging table */
-    CREATE UNLOGGED TABLE stage_1 AS
```
sumedhpathak: Any reason to change this?

performance/scaling_data_ingestion.rst
```diff
@@ -176,36 +176,38 @@ It may also be useful to create a sequence to generate a unique name for the sta
...
+The example above again uses a select_events_shard function to select the shard to which to append. To ensure parallel data ingestion, this function should balance across many different shards.
 
-An example choose_shard function belows randomly picks one of the 20 smallest shards or creates a new one if there are less than 20 under 1GB. This allows 20 concurrent appends, which allows data ingestion of up to 1 million rows/s (depending on indexes, size, capacity). 
+An example select_events_shard function belows randomly picks one of the 20 smallest shards or creates a new one if there are less than 20 under 1GB. This allows 20 concurrent appends, which allows data ingestion of up to 1 million rows/s (depending on indexes, size, capacity).
 
 ::
 
     /* Choose a shard to which to append */
-    CREATE OR REPLACE FUNCTION choose_shard(table_id regclass)
+    CREATE OR REPLACE FUNCTION select_events_shard()
```
sumedhpathak: Any reason to change this function name? Also, maybe a name like 'choose_underutilized_shard()' might be more descrptive?

performance/scaling_data_ingestion.rst
marcocitus: Mainly that it might not really be a good idea in Citus. What if the node restarts just before master_append_table_to_shard starts? That would not only give an unexpected result, but might actually trigger a master_append_table_to_shard bugs (empty tables resulting in empty string as min/max values).

sumedhpathak:
:shipit:

PR #61, opened 2016-05-14
Add 5.1 configuration settings to the reference

### LINKED ISSUE #51: Add 5.1 user defined functions to reference
* Transaction manager
* Update arguments / default values of any UDFs or config parameters which were changed.

marcocitus:
Closes #51 

I did not document the new master_update_shard_statistics UDF, since it is mainly for internal use and can cause problems if used incorrectly.

### DIFF
--- a/reference/configuration.rst
+++ b/reference/configuration.rst
@@ -33,6 +33,17 @@ Citus tracks worker nodes' locations and their membership in a shared hash table
 Data Loading
 ---------------------------
 
+citus.multi_shard_commit_protocol (enum)
+$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$
+
+Sets the commit protocol to use when performing COPY on a hash distributed table. On each individual shard placement, the COPY is performed in a transaction block to ensure that no data is ingested if an error occurs during the COPY. However, there is a particular failure case in which the COPY succeeds on all placements, but a (hardware) failure occurs before all transactions commit. This parameter can be used to prevent data loss in that case by choosing between the following commit protocols: 
+
+* **1pc:** The transactions in which which COPY is performed on the shard placements is committed in a single round. Data may be lost if a commit fails after COPY succeeds on all placements (rare). This is the default protocol.
+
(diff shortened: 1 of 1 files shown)


reference/configuration.rst
```diff
@@ -33,6 +33,17 @@ Citus tracks worker nodes' locations and their membership in a shared hash table
 Data Loading
 ---------------------------
 
+citus.multi_shard_commit_protocol (enum)
+$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$
+
+Sets the commit protocol to use when performing COPY on a hash distributed table. On each individual shard placement, the COPY is performed in a transaction block to ensure that no data is ingested if an error occurs during the COPY. However, there is a particular failure case in which the COPY succeeds on all placements, but a (hardware) failure occurs before all transactions commit. This parameter can be used to prevent data loss in that case by choosing between the following commit protocols: 
+
+* **1pc:** The transactions in which which COPY is performed on the shard placements is committed in a single round. Data may be lost if a commit fails after COPY succeeds on all placements (rare). This is the default protocol.
```
sumedhpathak: The verbiage and the text '(rare)' seems to imply this is the recommended setting. This is the default setting, but is it also the recommended one for production?

reference/configuration.rst
```diff
@@ -33,6 +33,17 @@ Citus tracks worker nodes' locations and their membership in a shared hash table
 Data Loading
 ---------------------------
 
+citus.multi_shard_commit_protocol (enum)
+$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$
+
+Sets the commit protocol to use when performing COPY on a hash distributed table. On each individual shard placement, the COPY is performed in a transaction block to ensure that no data is ingested if an error occurs during the COPY. However, there is a particular failure case in which the COPY succeeds on all placements, but a (hardware) failure occurs before all transactions commit. This parameter can be used to prevent data loss in that case by choosing between the following commit protocols:
```
sumedhpathak: Is the error primarily data-loss? i.e., will the client report back success even though some data was uncommitted? Or is it around inconsistency, i.e., some data might be copied, and some not?

reference/configuration.rst
marcocitus: It depends, the cost of monitoring prepared transactions might outweigh the tiny failure probability.

reference/configuration.rst
marcocitus: The COPY will report success with warnings. The problem is that you cannot easily recover because it only failed for particular hash ranges.

reference/configuration.rst
sumedhpathak: What is the cost? Is there a performance penalty? Or would one need to write scripts to monitor uncommitted prepare's on workers etc?

sumedhpathak:
:shipit:

PR #64, opened 2016-05-17
Use citus-tutorial bundle built for 5.1

### LINKED ISSUE #49: Rebundle tutorial and test for 5.1
Check `INSERT` especially

begriffs:
I built the new bundles and tested on Mac 10.11.5 and Ubuntu 16.04.

Can someone please independently verify that the tutorial works, preferably on a different Linux distribution? (Testing on something like CentOS 6 would probably be good since it's much older and has rather different internals.)

Fixes #49

### DIFF
--- a/tutorials/tut-cluster.rst
+++ b/tutorials/tut-cluster.rst
@@ -11,9 +11,9 @@ To do the tutorials you'll need a single-machine Citus cluster with a master and
 .. raw:: html 
 
   <ul>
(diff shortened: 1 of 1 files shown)


sumedhpathak:
🚢 We'll need to test these, but the change itself looks good.

begriffs:
Bundle fails in CentOS 6 because it requires a newer glibc, but it works in CentOS 7. This is the same behavior as the previous version so we have no regression and this should be good to merge.

PR #65, opened 2016-05-17
Mention citus cloud in the AWS installation docs

begriffs:
Let people know it's an option and how it differs from CloudFormation.

I dug through our css and found existing styles `.wy-btn-group` and `.wy-text-center` to make my call to action button look the way I wanted. Not sure if it's a class I should rely on existing for the long term...


begriffs:
Discussed with Craig in out-of-band conversation, merging.

PR #73, opened 2016-05-25
Prevent nav menu from clipping on short document bodies

begriffs:
Use floats rather than absolute positioning for menu


begriffs:
Looks good at various widths and I tried it on some browsers and mobile devices on browserstack.

PR #81, opened 2016-05-31
Allow trust for localhost in pg_hba.conf

### LINKED ISSUE #66: Enable local connections to trust in pg_hba.conf - Repartitioning queries failure
Tested Version: v5.1.0-rc.1
Created two tables and tried to execute a repartition join.

The query failed with the following error on master
```
ERROR:  failed to execute job 11
DETAIL:  Too many task tracker failures
```
In the worker log this was the error:
```
2016-05-18 00:48:03 UTC [8246-1] postgres@postgres LOG:  could not receive data from client: Connection reset by peer
2016-05-18 00:48:03 UTC [4127-419] WARNING:  could not connect to node localhost:5432
2016-05-18 00:48:03 UTC [4127-420] DETAIL:  Client error: fe_sendauth: no password supplied
```

This was resolved after I changed pg_hba.conf on the worker
```
# IPv4 local connections:
host    all             all             127.0.0.1/32            trust
# IPv6 local connections:
host    all             all             ::1/128                     trust
```

**Steps to Reproduce**

```SQL
CREATE TABLE large1(id bigint, val bigint);
CREATE TABLE large2(id bigint);
```
citus.large_table_shard_count was by default 4.
Made both the above tables large tables with different partition columns.

Hash distributed the large1 table with val as the partition column.
```SQL
select master_create_distributed_table('large1','val','hash');
select master_create_worker_shards('large1',16,1);
```

Hash distributed the large2 table with id as the partition column.
```SQL
select master_create_distributed_table('large2','id','hash');
select master_create_worker_shards('large2',16,1);
```

Did a few inserts:
```SQL
INSERT INTO large1(id,val) VALUES(1,1)
INSERT INTO large1(id,val) VALUES(1,2)
INSERT INTO large2(id) VALUES(1);
```
Ran the join query which requires repartitioning
```SQL
set citus.task_executor_type to 'task-tracker';
SELECT COUNT(*) FROM large1,large2 where large1.id=large2.id;
```

begriffs:
I reproduced #66 using a multi-machine Ubuntu setup. Workers need to be able to connect to themselves. Master doesn't necessarily need to connect to itself as well but it simplifies the documentation to use the same `pg_hba.conf` settings for all nodes including master.

Fixes #66

### DIFF
--- a/installation/production_deb.rst
+++ b/installation/production_deb.rst
@@ -52,6 +52,10 @@ Before starting the database let's change its access permissions. By default the
   # correspond to 24, 20, and 16-bit blocks in Private IPv4 address spaces.
   host    all             all             10.0.0.0/8              trust
 
+  # Also allow the host unrestricted access to connect to itself
+  host    all             all             127.0.0.1/32            trust
+  host    all             all             ::1/128                 trust
+
 Note: Your DNS settings may differ. Also these settings are too permissive for some environments. The PostgreSQL manual `explains how <http://www.postgresql.org/docs/9.5/static/auth-pg-hba-conf.html>`_ to make them more restrictive.
 
 **4. Start database servers, create Citus extension**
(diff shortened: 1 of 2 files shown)


begriffs:
Closing this to move it against the v5.1 branch since we won't be working directly off master anymore.

PR #86, opened 2016-06-08
Add usage note about COPY

begriffs:


sumedhpathak:
I'm wondering if this belongs in the scaling data ingest section, or in each individual data-loading section for each partition method. (e.g.: https://www.citusdata.com/docs/citus/5.1/dist_tables/append_distribution.html#data-loading)

begriffs:
Good point, the scaling ingest section is a little out of the way for this kind of note. I moved it to the hash distribution insert section and duplicated in the append distribution insert section as well.

PR #87, opened 2016-06-08
Reassure minor upgrade

begriffs:
Some users were understandably a little incredulous that the short instructions for minor Citus upgrades are sufficient. Adding a reassuring note.

### DIFF
--- a/admin_guide/upgrading_citus.rst
+++ b/admin_guide/upgrading_citus.rst
@@ -165,7 +165,7 @@ For users who don’t want to take a cluster down and upgrade all nodes at the s
 Minor Update to Latest 5.x
 ##########################
 
-Upgrading requires first obtaining the new Citus extension and then installing it in each of your database instances, doing the master instance last. The first step varies by operating system.
+Upgrading requires first obtaining the new Citus extension and then installing it in each of your database instances. The first step varies by operating system.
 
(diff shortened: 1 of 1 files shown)


admin_guide/upgrading_citus.rst
```diff
@@ -165,7 +165,7 @@ For users who don’t want to take a cluster down and upgrade all nodes at the s
 Minor Update to Latest 5.x
 ##########################
 
-Upgrading requires first obtaining the new Citus extension and then installing it in each of your database instances. The first step varies by operating system.
+Upgrading requires first obtaining the new Citus extension and then installing it in each of your database instances, doing the master instance last. The first step varies by operating system.
```
sumedhpathak: Any reason to add this line?

sumedhpathak:
Looks good, but not sure about why we need to add a note to install last on the master node.

begriffs:
I thought perhaps if you upgrade the master to a version with new features before upgrading workers then it might ask them to do things they don't yet understand. But if that's not the case (and won't ever be the case for a minor upgrade) then I'll remove that warning.

sumedhpathak:
I don't know if that line will always be the situation. Let's leave it out for now. Other than that, 🚢

PR #88, opened 2016-06-08
No range partition

### LINKED ISSUE #36: Remove range partitioning to avoid confusion
<a href="https://github.com/samay-sharma"></img></a> **Issue by [samay-sharma](https://github.com/samay-sharma)**
_Friday Apr 15, 2016 at 21:22 GMT_
_Originally opened as https://github.com/citusdata/citus_documentation_source/issues/92_

----

The benefits of range over append are subtle and difficult to convey. Also, it'll need a lot of expertise to actually get it to work well.

Even though previously I was of the opinion that it's valuable to keep it just to show that Citus can also do range partitioning, it seems to be just adding confusion and bringing in more questions around range vs hash and append vs hash.

begriffs:
Fixes #36


sumedhpathak:
@begriffs Do you know why there isn't a link to the deployment? It just says "temporarily deployed".

begriffs:
Heroku removes the apps after a certain amount of inactivity. Just now I triggered the preview apps to be redeployed for the open PRs.

begriffs:
I didn't remove the mention of "hash or range" from https://github.com/citusdata/citus_docs/blob/master/dist_tables/querying.rst#repartition-joins because I thought it might refer to a range partition of an append distribution rather a range distribution. Let me know if I'm mistaken about the terminology.

sumedhpathak:
The repartition-joins comment does refer to range-partitioning, so let's remove that too. Apart from that, looks good.

PR #91, opened 2016-06-14
Document master_modify_multiple_shards UDF

begriffs:


begriffs:
Here's how it looks in the preview: https://citus-doc-preview-pr-91.herokuapp.com/reference/user_defined_functions.html#master-modify-multiple-shards

sumedhpathak:
Apart from documenting the UDF, I think we probably want to highlight this as a feature somewhere in the docs. I'm thinking if we should start a section around DML, which would include this feature, Upsert/Update documentation, master_apply_delete_command() for data expiry, and then regular delete/update statement documentation to denote what each of them provide.

index.rst
```diff
@@ -41,6 +41,12 @@ topics.
    dist_tables/postgresql_extensions.rst
 
 .. toctree::
+   :maxdepth: 1
+   :caption: Data Manipulation
```
sumedhpathak: Instead of adding a new section (which we may also want to do at some point), we already cover data deletions in the append and [hash distribution sections of our docs](https://www.citusdata.com/docs/citus/5.1/dist_tables/append_distribution.html#dropping-shards).

Could we add this in those sections?

There are also certain statements there, which need to be updated, e.g. in the append-distribution page:
"Please note that this function only deletes complete shards and not individual rows from shards. If your use case requires deletion of individual rows in real-time, please consider using the hash distribution method."

reference/user_defined_functions.rst
```diff
@@ -362,3 +362,31 @@ This example will attempt to bring the shards of the github_events table to the
 ::
 
 	SELECT replicate_table_shards('github_events', max_shard_copies:=10);
+
+master_modify_multiple_shards
```
sumedhpathak: Can we move this to [Table/Shard DDL section](https://citus-doc-preview-pr-91.herokuapp.com/reference/user_defined_functions.html#master-apply-delete-command)? Currently it shows up at the end, and in the 'cluster management and repair functions' section.

The content in that section is no longer just DDL, its also DML, so we might want to update the section heading to reflect that.

begriffs:
@sumedhpathak I made the changes we discussed.

dist_tables/append_distribution.rst
```diff
@@ -122,15 +122,13 @@ $$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$
...
 
 The function uses shard metadata to decide whether or not a shard needs to be deleted, so it requires the WHERE clause in the DELETE statement to be on the distribution column. If no condition is specified, then all shards are selected for deletion. The UDF then connects to the worker nodes and issues DROP commands for all the shards which need to be deleted. If a drop query for a particular shard replica fails, then that replica is marked as TO DELETE. The shard replicas which are marked as TO DELETE are not considered for future queries and can be cleaned up later.
 
-Please note that this function only deletes complete shards and not individual rows from shards. If your use case requires deletion of individual rows in real-time, please consider using the hash distribution method.
```
sumedhpathak: Let's not remove the entire paragraph as the first note is still valid. We could say something like:
"Please note that this function only deletes complete shards and not individual rows from shards. If your use case requires deletion of individual rows, please see the master_modify_multiple_shards UDF" (or something like that).

dist_tables/append_distribution.rst
```diff
@@ -141,7 +139,19 @@ The example below deletes those shards from the github_events table which have a
...
+---------------
+
+The most flexible way to modify or delete rows throughout a Citus cluster is the master_modify_multiple_shards command. It takes a regular SQL statement as argument and runs it on all workers:
+
+::
+
+  SELECT master_modify_multiple_shards(
+    'DELETE FROM github_events WHERE created_at <= ''2014-01-01 00:00:00''');
+
+This uses a two-phase commit to remove or update data safely everywhere. Unlike master_apply_delete_command, it works at the row- rather than shard-level to modify or delete all rows that match the condition in the where clause. It deletes rows regardless of whether they comprise an entire shard. To learn more about the function, its arguments and its usage, please visit the :ref:`user_defined_functions` section of our documentation.
```
sumedhpathak: 2PC isn't the default. So we could say something like: "The function uses a configurable commit protocol to update or delete data safely across multiple shards."

sumedhpathak:
After addressing the two comments I made, its a 🚢

dist_tables/append_distribution.rst
begriffs: Actually this paragraph was just moved below, not deleted.

dist_tables/append_distribution.rst
```diff
@@ -141,7 +139,19 @@ The example below deletes those shards from the github_events table which have a
                                3
     (1 row)
 
-To learn more about the function, its arguments and its usage, please visit the :ref:`user_defined_functions` section of our documentation.
+To learn more about the function, its arguments and its usage, please visit the :ref:`user_defined_functions` section of our documentation.  Please note that this function only deletes complete shards and not individual rows from shards. If your use case requires deletion of individual rows in real-time, see the section below about deleting data.
```
begriffs: The "please note" paragraph was moved down here.

sumedhpathak: Ah, got it. I missed that. Thanks.

PR #92, opened 2016-06-17
Add "when to use" to the What is Citus page

begriffs:
@samay-sharma can you look this over quickly to see if it looks like it answers the right questions at a high level?

### DIFF
--- a/aboutcitus/what_is_citus.rst
+++ b/aboutcitus/what_is_citus.rst
@@ -1,6 +1,41 @@
 What is Citus?
 ==============
 
-Citus horizontally scales PostgreSQL across commodity servers using sharding and replication. Its query engine parallelizes incoming SQL queries across these servers to enable real-time responses on large datasets.
+Citus horizontally scales PostgreSQL across multiple machines using sharding and replication. It works well when you have a large data set and when you want to get answers from that data in human real-time – typically in less than a second.
 
 Citus extends the underlying database rather than forking it, which gives developers and enterprises the power and familiarity of a traditional relational database. As an extension, Citus supports new PostgreSQL releases, allowing users to benefit from new features while maintaining compatibility with existing PostgreSQL tools.
+
+When to Use Citus
+-----------------
+
+Example use cases:
+
+* Analytic dashboards with subsecond response times
+* Exploratory queries on unfolding events
+* Aggregate reports on large archives
+* Analyzing sessions with funnels, segmentation, and cohorts
+* Extending a data warehouse with real-time capabilities
+
+For concrete examples check out our customer `use cases <https://www.citusdata.com/solutions/case-studies>`_. Typical Citus workloads are operational, with aggregate queries and no long-lived transactions.
+
+Considerations for Use
+----------------------
+
+Although Citus extends PostgreSQL with distributed functionality, it is not a drop-in replacement that scales out all workloads. A performant Citus cluster involves thinking about the data model, tooling, and choice of SQL features used.
+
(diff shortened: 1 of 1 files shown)


sumedhpathak:
@begriffs Is the content signed-off by @ozgune ? Thanks.

begriffs:
@sumedhpathak yeah Ozgun and I revised it together in a google doc. If @samay-sharma can give it a quick look to see if he think's it answers customers' high level questions and approves it then we'll be good to merge.

aboutcitus/what_is_citus.rst
```diff
@@ -1,6 +1,41 @@
 What is Citus?
 ==============
 
-Citus horizontally scales PostgreSQL across commodity servers using sharding and replication. Its query engine parallelizes incoming SQL queries across these servers to enable real-time responses on large datasets.
+Citus horizontally scales PostgreSQL across multiple machines using sharding and replication. It works well when you have a large data set and when you want to get answers from that data in human real-time – typically in less than a second.
```
samay-sharma: Any specific reason why we removed "parallelize". I think that is one of our bigger value propositions and seems like an important thing to mention.

aboutcitus/what_is_citus.rst
```diff
@@ -1,6 +1,41 @@
...
+
+Example use cases:
+
+* Analytic dashboards with subsecond response times
+* Exploratory queries on unfolding events
+* Aggregate reports on large archives
+* Analyzing sessions with funnels, segmentation, and cohorts
+* Extending a data warehouse with real-time capabilities
+
+For concrete examples check out our customer `use cases <https://www.citusdata.com/solutions/case-studies>`_. Typical Citus workloads are operational, with aggregate queries and no long-lived transactions.
```
samay-sharma: do we want to mention joins?

aboutcitus/what_is_citus.rst
```diff
@@ -1,6 +1,41 @@
...
+
+Considerations for Use
```
samay-sharma: At a high level, I think this section makes sense. However, I think we can make it shorter and less repetitive.

eg. We say "Although Citus extends PostgreSQL with distributed functionality, it is not a drop-in replacement that scales out all workloads." and "PostgreSQL provides thousands of features, and Citus doesn’t yet scale them all."

aboutcitus/what_is_citus.rst
```diff
@@ -1,6 +1,41 @@
...
+Data models that have fewer tables (<10) work much better than those that have hundreds of tables. This is a property of distributed systems: the more tables, the more distributed dependencies. Still, compared with NoSQL databases Citus does not require aggressive denormalization.
+
+Citus supports most PostgreSQL tools. Still users may need to take additional steps when using tools that require distributed execution. For example, pg_dump and pg_restore currently don’t take distributed backups, but there are scripted ways to make these tools work on individual PostgreSQL nodes.
+
+PostgreSQL provides thousands of features, and Citus doesn’t yet scale them all. A good way to think about feature coverage is the following: if your workload aligns with use-cases noted in the “When to use Citus” section and you happen to run into an unsupported query, then there’s usually a good workaround.
+
+When Citus is Inappropriate
+---------------------------
+
+Workloads which require a large (non-aggregated) flow of information between nodes generally do not work as well. For instance:
```
samay-sharma: For a user, it may not be clear what workloads would require a large flow of information between nodes. (eg. At a recent customer conversation, they thought that to compute aggregates, Citus would pull data to one node, which isn't true.) So, could we reframe this slightly to better define the workload?

begriffs:
I addressed Samay's concerns so merging.

PR #94, opened 2016-06-22
Document the explain command

### LINKED ISSUE #69: Document EXPLAIN
We're looking to add documentation on distributed EXPLAIN. How does one use it and what are its limitations?

One important point to highlight could be that EXPLAIN ANALYZE runs distributed tasks in serial and therefore significantly increases query runtime.

begriffs:
Fixes #69

### DIFF
--- a/performance/performance_tuning.rst
+++ b/performance/performance_tuning.rst
@@ -25,17 +25,20 @@ For distribution methods, Citus supports both append and hash distribution. Appe
 
 Hash based distribution is more suited to cases where users want to do real-time inserts along with analytics on their data or want to distribute by a non-ordered column (eg. user id). This data model is relevant for real-time analytics use cases; for example, actions in a mobile application, user website events, or social media analytics. This distribution method allows users to perform co-located joins and efficiently run queries involving equality based filters on the distribution column.
 
-Once you choose the right distribution method and column, you can then proceed
-to the next step, which is tuning single node performance.
+Once you choose the right distribution method and column, you can then proceed to the next step, which is tuning worker node performance.
 
-.. _examining_plan:
+.. _postgresql_tuning:
 
-Examining Distributed Query Plan
-##################################
+PostgreSQL tuning
+#################
 
-Citus extends the SQL EXPLAIN command to provide information about simple distributed queries, meaning queries that do not contain re-partition jobs. The EXPLAIN output consists of two parts: the distributed query and the master query. These parts provide information to help tune Citus and the underlying PostgreSQL database.
+The Citus master partitions an incoming query into fragment queries, and sends them to the workers for parallel processing. The workers are just extended PostgreSQL servers and they apply PostgreSQL's standard planning and execution logic for these queries. So, the first step in tuning Citus is tuning the PostgreSQL configuration parameters on the workers for high performance.
 
-Here is an example of explaining the plan for a query in the :ref:`hash partitioning tutorial <tut_hash>`:
+While adjusting parameters to tune the workers we want fast iterations. To begin the tuning process, create a Citus cluster and load a small portion of your data in it. This will keep operations fast. Once the data is loaded, use the EXPLAIN command on the master node to inspect performance. 
+
+Citus extends the SQL EXPLAIN command to provide information about simple distributed queries, meaning queries that do not contain re-partition jobs. The EXPLAIN output shows how each worker is processing the query and also a little about how the master node is combining all the results.
+
+Here is an example of explaining the plan for a particular query in the :ref:`hash partitioning tutorial <tut_hash>`:
 
 ::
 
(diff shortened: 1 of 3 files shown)


begriffs:
Fixes #69

(Just added that comment to auto-close the issue on merge)

sumedhpathak:
@begriffs I think you need to add the 'Fixes' comment in the description (not as a separate comment) for it to auto-close the issue.

Also, could you re-push the deployment of the docs?

begriffs:
Redeployed the docs.

And actually if I'm not mistaken Github will pick up the Fixes comment even within a discussion comment and close the issue when this PR is merged.

begriffs:
@sumedhpathak I made the changes we talked about, care to check it out?

performance/performance_tuning.rst
```diff
@@ -28,6 +28,155 @@ Hash based distribution is more suited to cases where users want to do real-time
 Once you choose the right distribution method and column, you can then proceed
```
sumedhpathak: Can we modify/remove this sentence? The next step now has changed.

performance/performance_tuning.rst
```diff
@@ -28,6 +28,155 @@ Hash based distribution is more suited to cases where users want to do real-time
...
 to the next step, which is tuning single node performance.
 
+.. _examining_plan:
+
+Examining Distributed Query Plan
+##################################
+
+Citus extends the SQL EXPLAIN command to provide information about simple distributed queries, meaning queries that do not contain re-partition jobs. The EXPLAIN output consists of two parts: the distributed query and the master query. These parts provide information to help tune Citus and the underlying PostgreSQL database.
```
sumedhpathak: I think the [first paragraph from the PostgreSQL](https://citus-doc-preview-pr-94.herokuapp.com/performance/performance_tuning.html#postgresql-tuning) tuning section might be a good prelude to diving into the Explain plan. Maybe something to connect it like, "we can use EXPLAIN to visualize, and also tune for performance".

performance/performance_tuning.rst
```diff
@@ -28,6 +28,155 @@ Hash based distribution is more suited to cases where users want to do real-time
...
+  Executor: Real-Time
+  Task Count: 16
+  Tasks Shown: One of 16
+
+Next explain goes into each shard of each worker and finds the plan used internally there by PostgreSQL. By default it shows the output from only one shard. In the case of this example it was running on a worker on localhost, port 9701.
+
+::
+
+  ->  Task
+    Node: host=localhost port=9701 dbname=j
```
sumedhpathak: Is it possible to highlight / bold sections that we are referring to? E.g. the host/port here? 
The example below around one node having a different plan also is a good candidate, if we can bold the fact that the one task uses hash-join instead of nested loop. Will draw the readers' eye to it.

performance/performance_tuning.rst
```diff
@@ -28,6 +28,155 @@ Hash based distribution is more suited to cases where users want to do real-time
...
+
+The above is standard PostgreSQL explain output. You can use it for :ref:`postgresql_tuning` on the workers. After the output of distributed tasks there is a section about Citus' behavior on the master node.
+
+::
+
+  Master Query
+    ->  Limit  (cost=0.00..0.00 rows=0 width=0)
+      ->  Seq Scan on pg_merge_job_0005  (cost=0.00..0.00 rows=0 width=0)
+
+This explains how Citus is combining the results of each task to create the final result. If you see an expensive query plan in this area it means you should apply :ref:`distributed_query_performance_tuning` to improve it.
```
sumedhpathak: The last line isn't entirely accurate, so let's take this out. The query plan alone won't give users information on how to tune their distributed settings.

performance/performance_tuning.rst
```diff
@@ -28,6 +28,155 @@ Hash based distribution is more suited to cases where users want to do real-time
...
+
+Next explain goes into each shard of each worker and finds the plan used internally there by PostgreSQL. By default it shows the output from only one shard. In the case of this example it was running on a worker on localhost, port 9701.
+
+::
+
+  ->  Task
+    Node: host=localhost port=9701 dbname=j
+    ->  Limit  (cost=0.15..6.87 rows=10 width=32)
+      ->  Nested Loop  (cost=0.15..131.12 rows=195 width=32)
+        ->  Seq Scan on wikipedia_changes_102024 c  (cost=0.00..13.90 rows=390 width=64)
```
sumedhpathak: Can we also highlight that its hitting this specific shard? ``wikipedia_changes_102024``?

sumedhpathak:
I think the content itself looks good. I think we need to modify some of the preceding and proceeding sections to tie it all together. It probably requires some discussion so you get context on how we do performance tuning. That's probably best done in-person, so let's look to wrap this up on Tuesday.

performance/performance_tuning.rst
```diff
@@ -49,12 +103,67 @@ Lastly, you can create indexes on your tables to enhance database performance. I
 
 For write performance, you can use general PostgreSQL configuration tuning to increase INSERT rates. We commonly recommend increasing checkpoint_timeout and max_wal_size settings. Also, depending on the reliability requirements of your application, you can choose to change fsync or synchronous_commit values.
 
+Once you have tuned a worker to your satisfaction you will have to manually apply those changes to the other workers as well. To verify that they are all behaving properly, set this configuration variable on the master:
+
+::
+
+  SET citus.explain_all_tasks = 1;
```
ozgune: Minor suggestion.

I was going over https://github.com/citusdata/citus/issues/621 and ended up here.

Is it worth highlighting here that enabling this flag notably hurts ``EXPLAIN ANALYZE`` performance? I'm making this suggestion for two reasons:

1. When the user runs ``EXPLAIN ANALYZE``, they are likely investigating a performance problem. By enabling this flag, they are likely making the performance issue much worse.
2. Citus' ``EXPLAIN ANALYZE`` behavior is inconsistent with that of Postgres'.

lfittl: @ozgune I still find it odd that we recommend people not explaining all tasks - maybe I'm wrong here, but why would the first task thats being run be representative of the other shards? (depending on data distribution with hash, you could even have different plans for different shards)

begriffs: If you've just loaded a small amount of data into the cluster and are working on tuning one of the workers it might be nice to see only that one worker in explain. Once you like the tuning on that worker you could apply it to the others then show all to see if any behave differently due to uneven data distrubtion.

Then again I'm not sure whether the worker chosen by explain stays constant between invocations or changes at random. I should test that...

performance/performance_tuning.rst
```diff
@@ -35,7 +34,62 @@ PostgreSQL tuning
 
 The Citus master partitions an incoming query into fragment queries, and sends them to the workers for parallel processing. The workers are just extended PostgreSQL servers and they apply PostgreSQL's standard planning and execution logic for these queries. So, the first step in tuning Citus is tuning the PostgreSQL configuration parameters on the workers for high performance.
 
-This step involves loading a small portion of the data into regular PostgreSQL tables on the workers. These tables would be representative of shards on which your queries would run once the full data has been distributed. Then, you can run the fragment queries on the individual shards and use standard PostgreSQL tuning to optimize query performance.
+While adjusting parameters to tune the workers we want fast iterations. To begin the tuning process, create a Citus cluster and load a small portion of your data in it. This will keep operations fast. Once the data is loaded, use the EXPLAIN command on the master node to inspect performance.
```
sumedhpathak: This paragraph confuses me a bit. Especially these two sentences:
"While adjusting parameters to tune the workers we want fast iterations"
and
"This will keep operations fast". 

What do we want to communicate here with those statements?

performance/performance_tuning.rst
```diff
@@ -35,7 +34,62 @@ PostgreSQL tuning
...
 
-This step involves loading a small portion of the data into regular PostgreSQL tables on the workers. These tables would be representative of shards on which your queries would run once the full data has been distributed. Then, you can run the fragment queries on the individual shards and use standard PostgreSQL tuning to optimize query performance.
+While adjusting parameters to tune the workers we want fast iterations. To begin the tuning process, create a Citus cluster and load a small portion of your data in it. This will keep operations fast. Once the data is loaded, use the EXPLAIN command on the master node to inspect performance. 
+
+Citus extends the SQL EXPLAIN command to provide information about simple distributed queries, meaning queries that do not contain re-partition jobs. The EXPLAIN output shows how each worker is processing the query and also a little about how the master node is combining all the results.
```
sumedhpathak: I think we can leave out the sentence around "queries that do not contain re-partition jobs". There will need to be context explained here, and it's easier for them to just see the error message.

performance/performance_tuning.rst
```diff
@@ -35,7 +34,62 @@ PostgreSQL tuning
...
+    Executor: Real-Time
+    Task Count: 16
+
+Next it picks one of the workers to and shows you more about how the query behaves there. It indicates the host, port, and database so you can connect to the worker directly if desired:
+
+::
+
+    Tasks Shown: One of 16
+    ->  Task
+      Node: host=localhost port=9701 dbname=j
```
sumedhpathak: Can we use a different dbname other than "j"? Maybe something like "postgres" as a default?

performance/performance_tuning.rst
```diff
@@ -35,7 +34,62 @@ PostgreSQL tuning
...
+::
+
+      ->  Limit  (cost=0.15..6.87 rows=10 width=32)
+        ->  Nested Loop  (cost=0.15..131.12 rows=195 width=32)
+          ->  Seq Scan on wikipedia_changes_102024 c  (cost=0.00..13.90 rows=390 width=64)
+          ->  Index Scan using wikipedia_editors_editor_key_102008 on wikipedia_editors_102008 e  (cost=0.15..0.29 rows=1 width=32)
+            Index Cond: (editor = c.editor)
+            Filter: (bot IS TRUE)
+
+At this stage you can connect to the worker to use standard PostgreSQL tuning to optimize query performance. As you make changes try re-running EXPLAIN from the master.
```
sumedhpathak: Along with "connect to the worker", can we say something like, "You can now connect to the worker at 'localhost', port 'xxx' and tune query performance for the shard: "wikipedia_changes_102024"

performance/performance_tuning.rst
sumedhpathak: @ozgune We already have a note in the last paragraph. @begriffs Could we make that into a proper "Note" form? So it stands out visually?

performance/performance_tuning.rst
```diff
@@ -35,7 +34,62 @@ PostgreSQL tuning
...
 
-This step involves loading a small portion of the data into regular PostgreSQL tables on the workers. These tables would be representative of shards on which your queries would run once the full data has been distributed. Then, you can run the fragment queries on the individual shards and use standard PostgreSQL tuning to optimize query performance.
+While adjusting parameters to tune the workers we want fast iterations. To begin the tuning process, create a Citus cluster and load a small portion of your data in it. This will keep operations fast. Once the data is loaded, use the EXPLAIN command on the master node to inspect performance. 
+
+Citus extends the SQL EXPLAIN command to provide information about simple distributed queries, meaning queries that do not contain re-partition jobs. The EXPLAIN output shows how each worker is processing the query and also a little about how the master node is combining all the results.
+
+Here is an example of explaining the plan for a particular query in the :ref:`hash partitioning tutorial <tut_hash>`:
```
sumedhpathak: Let's evaluate naming in the other open PR. #105 . I don't want to change it here to hash-partitioning, and chose a different name in that PR.
For now, can we just say something like "explain the plan for a query from one of our example tutorials"? What do you think?

performance/performance_tuning.rst
begriffs: Done.

begriffs:
I addressed the latest comments. Good to merge?

sumedhpathak:
@begriffs Could you trigger a deployment for preview?

sumedhpathak:
Looks good. 🚢

PR #103, opened 2016-06-28
Moved one question to the top

umurc:
Moved question about adding primary keys to the top


begriffs:
Looks good 😎 

@umurc want to do the honors of merging? (Just press the green Merge pull request button, and the confirmation button)

umurc:
@begriffs ahh, I would have loved to - seeing it now! :) Will be on the lookout for the next one I submit.

PR #104, opened 2016-06-28
Small updates to "What is Citus" section

ozgune:
Small updates to "What is Citus" section


begriffs:
The changes look good!

For some reason there are two commits in this pull request, but I'll squash it down into a single commit and merge.

PR #105, opened 2016-06-28
Put hash distribution first, call it by its name

### LINKED ISSUE #28: Choose more meaningful tutorial names in the TOC
<a href="https://github.com/begriffs"></img></a> **Issue by [begriffs](https://github.com/begriffs)**
_Friday Mar 18, 2016 at 19:20 GMT_
_Originally opened as https://github.com/citusdata/citus_documentation_source/issues/76_

----

We're looking for more meaningful names than "Real Time Aggregation" and "Updatable User Data"

/cc @ozgune @lithp @samay-sharma

### LINKED ISSUE #95: Restructure docs to emphasize Hash distribution over Append distribution
@begriffs reviewed our current docs, and shared his findings.

One of his insights came from looking at the Tutorials and Distributed Tables sections in our website. In these sections, about 2-4x more people visit Append Distribution related sections than Hash Distribution ones.

We guess that's primarily because we position Append Distribution earlier in our docs than Hash Distribution. We have two reasons for the current order: (a) we had append distribution earlier and have more customers on it, and (b) hash distribution's ingest performance was pretty low up until the COPY changes.

Given that we now have COPY, enable upsert / update / deletes with hash distribution, and we're building MX, we're planning to flip the order in our Tutorials and Distributed Tables sections and give hash distribution more weight.

begriffs:
In addition to reordering the tutorials and swapping their intro paragraphs I changed their names. The new names are pretty literal, maybe too much so. However they are unambiguous.

Fixes #95
Fixes #28

### DIFF
--- a/index.rst
+++ b/index.rst
@@ -19,8 +19,8 @@ topics.
    :caption: Tutorials
 
    tutorials/tut-cluster.rst
-   tutorials/tut-real-time.rst
-   tutorials/tut-user-data.rst
+   tutorials/tut-hash-distribution.rst
+   tutorials/tut-append-distribution.rst
 
 .. toctree::
    :caption: Installation
--- a/index.rst
+++ b/index.rst
@@ -34,8 +34,8 @@ topics.
    :caption: Distributed Tables
 
    dist_tables/working_with_distributed_tables.rst
-   dist_tables/append_distribution.rst
    dist_tables/hash_distribution.rst
+   dist_tables/append_distribution.rst
    dist_tables/querying.rst
    dist_tables/postgresql_extensions.rst
 
--- a/installation/production_deb.rst
+++ b/installation/production_deb.rst
@@ -109,7 +109,7 @@ To verify that the installation has succeeded, we check that the master node has
 
 **Ready to use Citus**
 
-At this step, you have completed the installation process and are ready to use your Citus cluster. To help you get started, we have a :ref:`real-time aggregation tutorial<tut_timeseries>` which has instructions on setting up a Citus cluster with sample data in minutes.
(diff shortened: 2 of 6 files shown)


index.rst
```diff
@@ -19,8 +19,8 @@ topics.
    :caption: Tutorials
 
    tutorials/tut-cluster.rst
-   tutorials/tut-real-time.rst
-   tutorials/tut-user-data.rst
+   tutorials/tut-hash-distribution.rst
+   tutorials/tut-append-distribution.rst
```
begriffs: This diff doesn't make it super clear, but the sections were renamed and then put in the opposite order.

sumedhpathak:
Should we also move the [Hash-distribution](http://docs.citusdata.com/en/latest/dist_tables/hash_distribution.html) section above Append in the documentation proper?

sumedhpathak:
On the naming, I think we should get @ozgune's opinion. Could you make sure you chat with him about it?

begriffs:
@ozgune suggested we find names that are understood more widely. How about this latest change?

ozgune:
I liked the earlier name proposals more (hash distribution and append distribution). The challenge is that both tutorials can accomplish "Sharding data" or "Working with time-series data."

Some differences that I see are the following:

- Append distribution only handles immutable events. Hash distribution supports update/upserts.
- Append distribution supports higher data ingest ratios. Append distribution also scales out better since you can ingest from workers. With recent improvements in Citus, this is less of an issue.
- Append distribution is less likely to mark shards as unavailable.
- Append distribution can't handle co-located joins. Hash distribution can.

PR #118, opened 2016-07-26
[WIP] Reference architecture one

lithp:
do not merge, this is only open so we can collect comments


ref-arch/real-time-analytics.rst
```diff
@@ -0,0 +1,248 @@
...
+as possible and show both historical and live data on a dashboard.
+
+In this reference architecture we'll demonstrate how to build part of the first example
+but this architecture would work equally well for the second and many other business
+use-cases.
+
+Running It Yourself
+-------------------
+
+There's `a github repo <http://github.com>`_ with scripts and usage instructions. If
```
begriffs: Looks like the repo url is incomplete.

ref-arch/real-time-analytics.rst
```diff
@@ -0,0 +1,248 @@
...
+but the github repo has all the details in one place.
+
+Data Model
+----------
+
+The data we're dealing with is an immutable stream of log data. Here we'll insert directly
+into Citus but it's also common for this data to first be routed through something like
+Kafka. Doing so makes the system a little more resilient to failures, lets the data be
+routed to multiple places (such as a warehouse like redshift), and (once data volumes
+become unmanageable high) makes it a little easier to start pre-aggregating the data
```
begriffs: unmanageably

ref-arch/real-time-analytics.rst
```diff
@@ -0,0 +1,248 @@
...
+    response_time_msec INT,
+  )
+  SELECT master_create_distributed_table('http_requests', 'zone_id', 'hash');
+  SELECT master_create_worker_shards('http_requests', 16, 2);
+
+When we call master_create_distributed_table we're telling it to hash-distribute using
+the `zone_id` column. That means dashboard queries will each hit a single shard.
+
+When we call master_create_worker_shards we tell it to create 16 shards, and 2 copies of
+each shard. `We recommend
```
begriffs: It would be more robust to add

```rst
.. _faq_how_to_choose_shard_count:
```

to `faq/faq.rst` before the subheader for that question and then change this link to

```rst
`We recommend <faq_how_to_choose_shard_count>`_
```

ref-arch/real-time-analytics.rst
```diff
@@ -0,0 +1,248 @@
...
+<http://docs.citusdata.com/en/v5.1/faq/faq.html#how-do-i-choose-the-shard-count-when-i-hash-partition-my-data>`_
+using 2-4x as many shards as cores in your cluster, this reduces the overhead of keeping
+track of lots of shards, but also leaves you plenty of shards to spread around when you
+scale up your cluster by adding workers.
+
+Using a replication factor of 2 (or any number > 1, really), means data is written to
+multiple workers, when a node fails the other shard will be used instead.
+
+Note: In Citus Cloud you must use a replication factor of 1, as they have a different HA
```
begriffs: Maybe

```rst
.. note::

  In Citus Cloud you  ...
```

ref-arch/real-time-analytics.rst
```diff
@@ -0,0 +1,248 @@
...
+- Use lock_timeout so you don't get a bunch of these queued up
+
+- There's a local table on each worker keeping track of the high-water mark for each:
+
+.. code-block:: sql
+
+  -- this should be run on each worker
+  CREATE TABLE rollup_thresholds (
+        ingest_time timestamptz,
+        granulatiry text,
```
begriffs: typo

ref-arch/real-time-analytics.rst
```diff
@@ -0,0 +1,434 @@
...
+spread around when you scale up your cluster by adding workers.
+
+Using a replication factor of 2 (or any number > 1, really), means data is written to
+multiple workers. When a node fails the worker will serve queries for a shard using the
+other node so you don't have any downtime.
+
+.. NOTE::
+
+  In Citus Cloud you must use a replication factor of 1 (instead of the 2 used here), as
+  they have a different HA solution.
```
lfittl: As mentioned in the call, I'd word this as:

```
In Citus Cloud you must use a replication factor of 1 (instead of the 2 used here), as Citus Cloud uses PostgreSQL streaming replication for high availability.
```

begriffs:
I know the code is a rough draft, but I experimented with it locally and found some things to change. I'll add some line comments.

ref-arch/real-time-analytics.rst
```diff
@@ -0,0 +1,509 @@
...
+    zone_id INT, -- every customer's site is a different "zone"
+    ingest_time TIMESTAMPTZ DEFAULT now(),
+
+    session_id UUID,
+    url TEXT,
+    request_country TEXT,
+    ip_address CIDR,
+
+    status_code INT,
+    response_time_msec INT,
```
begriffs: Remove final comma

ref-arch/real-time-analytics.rst
```diff
@@ -0,0 +1,509 @@
...
+  availability maintaining shard replicas would be redundant and inneficient.
+
+With this, the system is ready to accept data and serve queries! You can run
+queries such as:
+
+.. code-block:: sql
+
+  INSERT INTO http_request (
+      zone_id, session_id, url, request_country,
+      ip_address, status_code, reponse_time_msec
```
begriffs: Typo: *reponse*_time_msec

ref-arch/real-time-analytics.rst
```diff
@@ -0,0 +1,509 @@
...
+And do some dashboard queries like:
+
+.. code-block:: sql
+
+  SELECT
+    date_trunc('minute', ingest_time) as minute,
+    COUNT(1) AS request_count,
+
+    COUNT(CASE WHEN (status_code between 200 and 299) THEN 1 ELSE 0 END) as success_count,
+    request_count - success_count AS error_count,
```
begriffs: request_count doesn't seem to be accessible here

ref-arch/real-time-analytics.rst
```diff
@@ -0,0 +1,509 @@
...
+  CREATE TABLE http_request_1min (
+        zone_id INT,
+        ingest_time TIMESTAMPTZ, -- which minute this row represents
+
+        error_count INT,
+        success_count INT,
+        request_count INT,
+        average_response_time_msec INT,
+        CHECK (request_count = error_count + success_count)
+  )
```
begriffs: add `;`

ref-arch/real-time-analytics.rst
```diff
@@ -0,0 +1,509 @@
...
+pre-aggregated form. Here, we'll aggregat the raw data into other tables which store
+summaries of 1-minute, 1-hour, and 1-day intervals. These might correspond to zoom-levels
+in the dashboard. When the user wants request times for the last month the dashboard can
+read and chart the values for each of the last 30 days, no math required! For the rest of
+this document we'll only talk about the first granularity, the 1-minute one. The github
+repo has `DDL <http://github.com>`_ for the other resolutions.
+
+.. code-block:: sql
+
+  CREATE TABLE http_request_1min (
```
begriffs: Should be http_requests_1min, plural

ref-arch/real-time-analytics.rst
```diff
@@ -0,0 +1,509 @@
...
+        request_count INT,
+        average_response_time_msec INT,
+        CHECK (request_count = error_count + success_count)
+  )
+  SELECT master_create_distributed_table('http_requests_1min', 'zone_id', 'hash');
+  SELECT master_create_worker_shards('http_requests_1min', 16, 2);
+  
+  -- indexes aren't automatically created by Citus
+  -- this will create the index on all shards
+  CREATE INDEX ON http_requests_1min (zone_id, ingest_time);
```
begriffs: Just needs some arbitrary name
```
ERROR:  creating index without a name on a distributed table is currently unsupported
```

ref-arch/real-time-analytics.rst
```diff
@@ -0,0 +1,509 @@
...
+.. code-block:: plpgsql
+
+  -- this should be run on each worker
+  CREATE FUNCTION rollup_1min(source_shard text, dest_shard text) RETURNS void
+  AS $$
+  DECLARE
+    v_latest_minute_already_aggregated timestamptz;
+    v_new_latest_already_aggregated timestamptz;
+  BEGIN
+    PERFORM SET lock_timeout 100;
```
begriffs: I don't know what the problem is here, but it raised an error

```
ERROR:  syntax error at or near "100"
LINE 7:   PERFORM SET lock_timeout 100;
                                   ^
```

ref-arch/real-time-analytics.rst
```diff
@@ -0,0 +1,509 @@
...
+    PERFORM RESET lock_timeout;
+
+    EXECUTE format('
+      WITH (
+        INSERT INTO %I (
+            zone_id, ingest_time, request_count,
+            error_count, success_count, average_response_time_msec)
+          SELECT
+            zone_id,
+            date_trunc('minute', ingest_time) as minute,
```
begriffs: The single quotes should be quoted: `''minute''` since it's in a string for format().

ref-arch/real-time-analytics.rst
```diff
@@ -0,0 +1,509 @@
...
+Since this function is given some metadata from the master, where does the master get that
+metadata from? Every minute it calls its own function which fires off all the
+aggregations:
+
+.. code-block:: plpgsql
+
+  -- this should be run on the master
+  CREATE FUNCTION run_rollups(source_table text, dest_table text) RETURNS void
+  AS $$
+  DECLARE
```
begriffs: Postgres wants the names declared (and not conflicting with column names).

```sql
  source_shard integer;
  dest_shard integer;
  nname text;
  nport integer;
```

ref-arch/real-time-analytics.rst
```diff
@@ -0,0 +1,509 @@
...
+  BEGIN
+    FOR source_shard, dest_shard, nodename, nodeport IN
+      SELECT
+        a.logicalrelid::regclass||'_'||a.shardid,
+        b.logicalrelid::regclass||'_'||b.shardid,
+        nodename, nodeport
+      FROM pg_dist_shard a
+      JOIN pg_dist_shard b USING (shardminvalue)
+      JOIN pg_dist_shard_placement p ON (a.shardid = p.shardid)
+      WHERE a.logicalrelid = 'first'::regclass AND b.logicalrelid = 'second'::regclass;
```
begriffs: Remove `;`

ref-arch/real-time-analytics.rst
```diff
@@ -0,0 +1,509 @@
...
+-----------------
+
+The rollups make queries faster but we still have a lot of raw data sitting around. How
+long you should keep each granularity of data is a business decision, but once you decide
+it's easy to write a function to expire old data:
+
+.. code-block:: plpgsql
+
+  -- another function for the master
+  CREATE FUNCTION expire_old_request_data RETURNS void
```
begriffs: Add empty list of arguments: `expire_old_request_data()`

ref-arch/real-time-analytics.rst
```diff
@@ -0,0 +1,509 @@
...
+    ingest_time TIMESTAMPTZ DEFAULT now(),
+
+    session_id UUID,
+    url TEXT,
+    request_country TEXT,
+    ip_address CIDR,
+
+    status_code INT,
+    response_time_msec INT,
+  )
```
marcocitus: semicolon.

ref-arch/real-time-analytics.rst
```diff
@@ -0,0 +1,509 @@
...
+
+  -- this gets run on the master
+  CREATE TABLE http_request (
+    zone_id INT, -- every customer's site is a different "zone"
+    ingest_time TIMESTAMPTZ DEFAULT now(),
+
+    session_id UUID,
+    url TEXT,
+    request_country TEXT,
+    ip_address CIDR,
```
marcocitus: Should this be INET? CIDR represents a range of IPs.

ref-arch/real-time-analytics.rst
```diff
@@ -0,0 +1,509 @@
...
+  WHERE zone_id = 1 AND minute = date_trunc('minute', now());
+
+Resources
+---------
+
+That's everything we wanted to cover. This article has been a little more in-depth than
+the rest of our documentation, but it shows a complete system to give you an idea of what
+building a non-trivial application with Citus looks like. We hope it helps you figure out
+how to use Citus for your specific use-case. Have we mentioned there's `a github repo
+<https://github.com>`_ with lots of resources?
```
marcocitus: The URL doesn't point to the Citus repo

ref-arch/real-time-analytics.rst
```diff
@@ -0,0 +1,509 @@
...
+individual country and blowing up your row width.  We have `a blog post
+<https://www.citusdata.com/blog/2016/07/14/choosing-nosql-hstore-json-jsonb/>`_ explaining
+which format to use for your semi-structured data. It says you should usually use jsonb
+but never says how. Let's correct that :)
+
+First, add the new column to our rollup table:
+
+.. code-block:: sql
+
+  ALTER TABLE http_requests_1min ADD COLUMN country_counters (JSONB);
```
marcocitus: I think counters is not a great use of JSONB, since the entire row is re-written whenever you update a counter. Is it possible to add something to the http_request table? e.g. cookies?

ref-arch/real-time-analytics.rst
```diff
@@ -0,0 +1,509 @@
...
+
+  -- careful, doesn't work!
+  SELECT
+    hll_cardinality(hll_union_agg(distinct_sessions))
+  FROM http_request_1day
+  WHERE ingest_time BETWEEN timestamp '06-01-2016' AND '06-08-2016';
+
+Well, it would be easy, except since Citus `can't yet
+<https://github.com/citusdata/citus/issues/120>`_ push down aggregates such as
+hll_union_agg. Instead you have to do a bit of trickery:
```
marcocitus: maybe:
Instead you have to do bit of trickery -> Instead, you can override the sum function to the same effect.

ref-arch/real-time-analytics.rst
```diff
@@ -0,0 +1,509 @@
...
+  WHERE zone_id = 1 AND minute = date_trunc('minute', now());
+
+HLLs aren't just faster, they let you do things you couldn't previously. Say we did our
+rollups, but instead of using HLLs we saved the exact unique counts. This works fine, but
+you can't answer queries such as "how many distinct sessions were there during this
+one-week period in the past we've thrown away the raw data for?". With HLLs, it's easy:
+
+.. code-block:: sql
+
+  -- careful, doesn't work!
```
marcocitus: Aww :)

ref-arch/real-time-analytics.rst
```diff
@@ -0,0 +1,509 @@
...
+
+For example, say you're building an HTTP analytics dashboard. You have enough clients that
+around every millisecond a user hits one of their websites and sends a log record to you.
+You want to ingest all of those records (1000 inserts/sec) and create a dashboard which
+shows your clients things like how many requests to their sites are errors. It's important
+that this data show up with as little latency as possible so your clients can fix problems
+with their sites. It's also useful to show graphs of historical data, however keeping all
+the raw data around forever is prohibitively expensive.
+
+Or maybe you're building an advertising network and want to show clients clickthrough
```
marcocitus: "Or" is generally not a suitable word for starting a paragraph or a sentence.

Maybe:
"Another example is: You're building an advertising network and ..."

ref-arch/real-time-analytics.rst
```diff
@@ -0,0 +1,509 @@
...
+also high, and both historical and live data are important.
+
+In this reference architecture we'll demonstrate how to build part of the first example
+but this architecture would work equally well for the second and many other business
+use-cases.
+
+Running It Yourself
+-------------------
+
+There's `a github repo <https://github.com>`_ with scripts and usage instructions. If
```
marcocitus: Link should be updated

ref-arch/real-time-analytics.rst
```diff
@@ -0,0 +1,509 @@
...
+you've gone through our installation instructions for running on either single or multiple
```
marcocitus: Maybe:
... running a Citus cluster on either a single machine or multiple machines, then you're ready to try it out.

ref-arch/real-time-analytics.rst
```diff
@@ -0,0 +1,509 @@
...
+machines you're ready to try it out. There will be code snippets in this tutorial but they
```
marcocitus: comma before but

ref-arch/real-time-analytics.rst
```diff
@@ -0,0 +1,509 @@
...
+don't quite specify a complete system, the github repo has all the details in one place.
```
marcocitus: , the github repo -> . The github repo

ref-arch/real-time-analytics.rst
```diff
@@ -0,0 +1,509 @@
...
+
+Data Model
+----------
+
+The data we're dealing with is an immutable stream of log data. Here we'll insert directly
```
marcocitus: Maybe:
Here -> In this example,

ref-arch/real-time-analytics.rst
```diff
@@ -0,0 +1,509 @@
...
+into Citus but it's also common for this data to first be routed through something like
```
marcocitus: Comma before but

ref-arch/real-time-analytics.rst
```diff
@@ -0,0 +1,509 @@
...
+Kafka. Doing so makes the system a little more resilient to failures, lets the data be
+routed to multiple places (such as a warehouse like redshift), and (once data volumes
```
marcocitus: Maybe these side notes in parentheses are not so important and could be omitted, or rewritten to something like:

makes it a little easier to start pre-aggregating the data before inserting in case raw data volumes become too high to query at run-time.

ref-arch/real-time-analytics.rst
```diff
@@ -0,0 +1,509 @@
...
+become unmanageably high) makes it a little easier to start pre-aggregating the data
+before inserting.
+
+In this example, the raw data will use the following schema which isn't very realistic as
+far as http analytics go but sufficient for showing off the architecture we have in mind.
+
+.. code-block:: sql
+
+  -- this gets run on the master
```
marcocitus: This comment could be more instructive, e.g.: "Run this on the master"

ref-arch/real-time-analytics.rst
```diff
@@ -0,0 +1,509 @@
...
+  SELECT master_create_distributed_table('http_request', 'zone_id', 'hash');
+  SELECT master_create_worker_shards('http_request', 16, 2);
+
+When we call :ref:`master_create_distributed_table <master_create_distributed_table>`
+we ask Citus to hash-distribute ``http_request`` using the ``zone_id`` column. That means
+that, since the dashboard only looks at a single zone, all dashboard queries will hit a
+single shard.
+
+When we call :ref:`master_create_worker_shards <master_create_worker_shards>` we tell it
+to create 16 shards, and 2 replicas of each shard (for a total of 32 shard replicas).
```
marcocitus: tell to -> tell Citus to

ref-arch/real-time-analytics.rst
```diff
@@ -0,0 +1,509 @@
...
+:ref:`We recommend <faq_choose_shard_count>` using 2-4x as many shards as cores in your
+cluster. If you use 1x as many shards as cores, any queries you run across the entire
```
marcocitus: 1x as many -> at least as many

ref-arch/real-time-analytics.rst
```diff
@@ -0,0 +1,509 @@
...
+dataset will run in parallel and take advantage of all your workers. If you use 2-4x as
+many shards, you make it easy to add workers later, but don't add much additional
+overhead.
+
+Using a replication factor of 2 (or any number > 1, really), means every row is written to
```
marcocitus: Maybe:
(or any number > 1, really) -> or higher

ref-arch/real-time-analytics.rst
```diff
@@ -0,0 +1,509 @@
...
+* The dashboard must aggregate every row in the target date range for every query it
+  answers.
+* Storage costs will grow proportionally with the ingest rate and the length of the
+  queryable history.
+
+Rollups
+-------
+
+In order to fix both problems, we have multiple clients who roll up the raw data into a
+pre-aggregated form. Here, we'll aggregat the raw data into other tables which store
```
marcocitus: aggregat -> aggregate

ref-arch/real-time-analytics.rst
```diff
@@ -0,0 +1,509 @@
...
+``http_request_1min`` shards, and Citus will place matching shards on the same worker.
+This is called colocation; it makes queries such as joins faster and our rollups possible.
+
+.. image:: /images/colocation.png
+  :alt: colocation in citus
+
+In order to populate ``http_request_1min`` we're going to periodically run the equivalent
+of an INSERT INTO SELECT. Citus doesn't yet support INSERT INTO SELECT on distributed
+tables, so instead we'll run a function on all the workers which runs INSERT INTO SELECT
+on every matching pair of shards (this is possible because the shards are colocated)
```
marcocitus: Maybe:
(this is possible because the shards are colocated) -> This is possible because the shards are colocated.

ref-arch/real-time-analytics.rst
marcocitus: Did you mean 100 seconds or milliseconds? 

    SET LOCAL lock_timeout TO '100s';

ref-arch/real-time-analytics.rst
```diff
@@ -0,0 +1,509 @@
...
+    -- runs at a time
+    SELECT ingest_time INTO v_latest_minute_already_aggregated FROM rollup_thresholds
+      WHERE source_shard = source_shard AND dest_shard = dest_shard
+      FOR UPDATE;
+    IF NOT FOUND THEN
+      INSERT INTO rollup_thresholds VALUES (
+        '1970-01-01', source_shard::regclass, dest_shard::regclass);
+      RETURN;
+    END IF;
+    PERFORM RESET lock_timeout;
```
marcocitus: If you use SET LOCAL then you don't need to RESET.

ref-arch/real-time-analytics.rst
```diff
@@ -0,0 +1,509 @@
...
+to read from and the name of the ``http_request_1min`` shard to write to. It can't figure
+it out itself because that kind of metadata is kept on the master, not the workers.
+
+It also uses a local table, to keep track of how much of the raw data has already been
+aggregated:
+
+.. code-block:: sql
+
+  -- every worker should have their own local version of this table
+  CREATE TABLE rollup_thresholds (
```
marcocitus: It would be nice if this was introduced before the function, such that I can run the code snippets sequentially and I know what the rollup_thresholds does before reading the function.

ref-arch/real-time-analytics.rst
```diff
@@ -0,0 +1,509 @@
...
+
+  -- another function for the master
+  CREATE FUNCTION expire_old_request_data RETURNS void
+  AS $$
+    SET citus.all_modification_commutative TO TRUE;
+    SELECT master_modify_multiple_shards(
+      'DELETE FROM http_request WHERE ingest_time < now() - interval ''1 hour'';');
+    SELECT master_modify_multiple_shards(
+      'DELETE FROM http_request_1min WHERE ingest_time < now() - interval ''1 day'';');
+    RESET citus.all_modification_commutative;
```
marcocitus: If you use SET LOCAL then you don't need RESET here.

ref-arch/real-time-analytics.rst
```diff
@@ -0,0 +1,509 @@
...
+minute,
+
+If you've heard of postgres' VACUUM, you know it can be a pain once you have a large
+number of rows. The write pattern we're using turns out to be the ideal VACUUM use-case.
+We never modify rows, we only write to one end of the table while deleting from the other
+end. When VACUUM runs it marks a `visibility map
+<https://www.postgresql.org/docs/9.5/static/storage-vm.html>`_ to keep track of which
+pages have already been vacuumed and can be skipped during the next vacuum. Since we never
+UPDATE, the only pages which have that bit reset are the pages of entirely DELETEd rows.
+All VACUUM needs to do is scan through and reclaim those empty pages, is happens very
```
marcocitus: is -> it

ref-arch/real-time-analytics.rst
```diff
@@ -0,0 +1,509 @@
...
+  consider adding a crontab entry:
+
+  .. code-block:: bash
+  
+    * * * * * psql -c "SELECT expire_old_request_data();"
+
+Review, what have we done?
+--------------------------
+
+That's the entire architecture! The next few sections are just extensions for additional
```
marcocitus: The word extensions is a bit confusing given the PostgreSQL context. Maybe just: The next few sections provide solutions for additional problems that often pop up.

ref-arch/real-time-analytics.rst
```diff
@@ -0,0 +1,509 @@
...
+problems which often pop up. So what makes this better than what it replaced? Well, let's
```
marcocitus: > So what makes this better than what it replaced?

I don't think we've replaced something. At least that hasn't been introduced yet.

ref-arch/real-time-analytics.rst
```diff
@@ -0,0 +1,509 @@
...
+<https://www.postgresql.org/docs/9.5/static/storage-vm.html>`_ to keep track of which
+pages have already been vacuumed and can be skipped during the next vacuum. Since we never
+UPDATE, the only pages which have that bit reset are the pages of entirely DELETEd rows.
+All VACUUM needs to do is scan through and reclaim those empty pages, is happens very
+quickly!
+
+Approximate Distinct Counts
+---------------------------
+
+One kind of query we're particularily proud of is :ref:`approximate distinct counts
```
marcocitus: "proud" seems a bit weird, maybe: "A common question in website analytics systems is: How many unique visitors ..." and later "we can answer the query approximately" -> "the HLL extension can answer the query approximately".

ref-arch/real-time-analytics.rst
```diff
@@ -0,0 +1,509 @@
...
+  -- working version of the above query
+  SELECT
+    hll_cardinality(SUM(distinct_sessions))
+  FROM http_request_1day
+  WHERE ingest_time BETWEEN timestamp '06-01-2016' AND '06-08-2016';
+
+More information on HLLs can be found in `their github repo
+<https://github.com/aggregateknowledge/postgresql-hll>`_.
+
+HLLS are an example from the postgresql community, but there are a couple extensions we've
```
marcocitus: We are part of the postgresql community ;). 

Maybe something like: 
The HLL extension only provides approximate distinct counts, but there are other extensions which do similar approximations for different kinds of queries, this includes ...

ref-arch/real-time-analytics.rst
```diff
@@ -0,0 +1,509 @@
...
+    PERFORM UPDATE rollup_thresholds
+      SET ingest_time = v_new_latest_already_aggregated
+      WHERE source_shard = source_shard AND dest_shard = dest_shard;
+  END;
+  $$ LANGUAGE 'plpgsql';
+
+As discussed above, there's a 1-to-1 correspondence between http_request shards and
+``http_request_1min`` shards. This function accepts the name of the ``http_request`` shard
+to read from and the name of the ``http_request_1min`` shard to write to. It can't figure
+it out itself because that kind of metadata is kept on the master, not the workers.
```
marcocitus: Lots of its. Maybe omit the sentence, or "The worker doesn't know which shard to use, since the Citus metadata is only kept on the master."

ref-arch/real-time-analytics.rst
```diff
@@ -0,0 +1,509 @@
...
+
+Real Time Analytics
+#####################
+
+Over the last few years we've helped many different kinds of clients use Citus and noticed
+a technical problem many businesses have: running real-time analytics over large streams
+of data.
+
+For example, say you're building an HTTP analytics dashboard. You have enough clients that
+around every millisecond a user hits one of their websites and sends a log record to you.
```
marcocitus: The word client has a specific meaning in an HTTP context. Maybe customers?

e.g.

You have enough clients that around every millisecond a user hits one of their websites and sends a log record to you.

->

Overall, requests from your customers generate around 1000 log records per second.

ref-arch/real-time-analytics.rst
```diff
@@ -0,0 +1,509 @@
...
+You want to ingest all of those records (1000 inserts/sec) and create a dashboard which
+shows your clients things like how many requests to their sites are errors. It's important
+that this data show up with as little latency as possible so your clients can fix problems
```
marcocitus: "this data shows up" or "these data show up"

ref-arch/real-time-analytics.rst
```diff
@@ -0,0 +1,509 @@
...
+
+And do some dashboard queries like:
+
+.. code-block:: sql
+
+  SELECT
+    date_trunc('minute', ingest_time) as minute,
+    COUNT(1) AS request_count,
+
+    COUNT(CASE WHEN (status_code between 200 and 299) THEN 1 ELSE 0 END) as success_count,
```
marcocitus: COUNT -> SUM

ref-arch/real-time-analytics.rst
```diff
@@ -0,0 +1,509 @@
...
+
+  -- this should be run on each worker
+  CREATE FUNCTION rollup_1min(source_shard text, dest_shard text) RETURNS void
+  AS $$
+  DECLARE
+    v_latest_minute_already_aggregated timestamptz;
+    v_new_latest_already_aggregated timestamptz;
+  BEGIN
+    PERFORM SET lock_timeout 100;
+    -- since master calls this function every minute, and future invokations will
```
marcocitus: invokations -> invocations

ref-arch/real-time-analytics.rst
```diff
@@ -0,0 +1,509 @@
...
+      WHERE source_shard = source_shard AND dest_shard = dest_shard
+      FOR UPDATE;
+    IF NOT FOUND THEN
+      INSERT INTO rollup_thresholds VALUES (
+        '1970-01-01', source_shard::regclass, dest_shard::regclass);
+      RETURN;
+    END IF;
+    PERFORM RESET lock_timeout;
+
+    EXECUTE format('
```
marcocitus: It's a bit nicer to use $$-notation to specify a multi-line string

ref-arch/real-time-analytics.rst
marcocitus: I think it could be useful to add a bit more explanation of the function here, e.g.

This function computes 1-minute aggregates over the raw data in a shard of the http_request table and writes them to the corresponding shard in the http_request_1min table. To ensure that we aggregate the data only once, we keep track of the time range that we've already rolled up in the rollup_threshold table.

ref-arch/real-time-analytics.rst
```diff
@@ -0,0 +1,509 @@
...
+
+  -- every worker should have their own local version of this table
+  CREATE TABLE rollup_thresholds (
+        ingest_time timestamptz,
+        source_shard regclass,
+        dest_shard regclass,
+        UNIQUE (source_shard, dest_shard)
+  );
+
+Since this function is given some metadata from the master, where does the master get that
```
marcocitus: I think the user might not have this particular question. Maybe:
"The rollup_1min function needs to be called for every shard in the http_request table. Only the master has the metadata, so we need to create a function on the master that goes through all the shards and runs rollup_1min on the workers."

marcocitus:
This is very nice overall. I mostly marked some typos and linguistic things. Feel free to selectively ignore them.

Main points:
- Change the JSONB example to something where the JSONB is in the raw data 
- Introduce the rollup_thresholds table before the function
- Maybe add a "psql ... | xargs ..." type shell command as an alternative for the rollup function, since dblink won't work on Citus Cloud

ref-arch/real-time-analytics.rst
```diff
@@ -0,0 +1,509 @@
...
+  SELECT
+    date_trunc('minute', ingest_time) as minute,
+    COUNT(1) AS request_count,
+
+    COUNT(CASE WHEN (status_code between 200 and 299) THEN 1 ELSE 0 END) as success_count,
+    request_count - success_count AS error_count,
+
+    SUM(response_time_msec) / COUNT(1) AS average_response_time_msec
+  FROM http_request
+  WHERE zone_id = 1 AND minute = date_trunc('minute', now())
```
marcocitus: Instead of "minute =", maybe: ingest_time >= now() - interval '1 hour'

ref-arch/real-time-analytics.rst
```diff
@@ -0,0 +1,509 @@
...
+        a.logicalrelid::regclass||'_'||a.shardid,
+        b.logicalrelid::regclass||'_'||b.shardid,
+        nodename, nodeport
+      FROM pg_dist_shard a
+      JOIN pg_dist_shard b USING (shardminvalue)
+      JOIN pg_dist_shard_placement p ON (a.shardid = p.shardid)
+      WHERE a.logicalrelid = 'first'::regclass AND b.logicalrelid = 'second'::regclass;
+    LOOP
+      SELECT * FROM dblink(
+        format('host=%s port=%d', nodename, nodeport),
```
marcocitus: format isn't quite like printf, %d should be %s

ref-arch/real-time-analytics.rst
```diff
@@ -0,0 +1,509 @@
...
+        format('SELECT rollup_1min(%, %s);', source_shard, dest_shard));
```
marcocitus: rollup_1min(%, %s) -> rollup_1min(%L, %L)

ref-arch/real-time-analytics.rst
```diff
@@ -0,0 +1,509 @@
...
+      SELECT
+        a.logicalrelid::regclass||'_'||a.shardid,
+        b.logicalrelid::regclass||'_'||b.shardid,
+        nodename, nodeport
+      FROM pg_dist_shard a
+      JOIN pg_dist_shard b USING (shardminvalue)
+      JOIN pg_dist_shard_placement p ON (a.shardid = p.shardid)
+      WHERE a.logicalrelid = 'first'::regclass AND b.logicalrelid = 'second'::regclass;
+    LOOP
+      SELECT * FROM dblink(
```
marcocitus: SELECT -> PERFORM

ref-arch/real-time-analytics.rst
```diff
@@ -0,0 +1,509 @@
...
+  );
+
+Since this function is given some metadata from the master, where does the master get that
+metadata from? Every minute it calls its own function which fires off all the
+aggregations:
+
+.. code-block:: plpgsql
+
+  -- this should be run on the master
+  CREATE FUNCTION run_rollups(source_table text, dest_table text) RETURNS void
```
marcocitus: It's usually nice to use regclass for table names, since that ensure the table exists.

ref-arch/real-time-analytics.rst
marcocitus: dblink requires a list of output columns, which works poorly with calling void functions.

You could make rollup_1min return timestamptz and put "AS (out timestamptz)" behind the dblink call.

ref-arch/real-time-analytics.rst
```diff
@@ -0,0 +1,509 @@
...
+          SELECT
+            zone_id,
+            date_trunc('minute', ingest_time) as minute,
+            COUNT(1) as request_count,
+            COUNT(CASE WHEN (status_code between 200 and 299) THEN 1 ELSE 0 END) as success_count,
+            request_count - success_count AS error_count,
+            SUM(response_time_msec) / COUNT(1) AS average_response_time_msec
+            -- later sections will show some more clauses that can go here
+          FROM %I
+          WHERE ingest_time > v_latest_minute_already_aggregated
```
marcocitus: v_latest_minute_already_aggregated -> %L, needs to be in the format arguments since it's not visible to the EXECUTE.

ref-arch/real-time-analytics.rst
```diff
@@ -0,0 +1,509 @@
...
+          GROUP BY zone_id, minute
+        ) as inserted_rows
```
marcocitus: should be at the start, i.e.:  WITH inserted_rows AS ( INSERT

ref-arch/real-time-analytics.rst
```diff
@@ -0,0 +1,509 @@
...
+            zone_id,
+            date_trunc('minute', ingest_time) as minute,
+            COUNT(1) as request_count,
+            COUNT(CASE WHEN (status_code between 200 and 299) THEN 1 ELSE 0 END) as success_count,
+            request_count - success_count AS error_count,
+            SUM(response_time_msec) / COUNT(1) AS average_response_time_msec
+            -- later sections will show some more clauses that can go here
+          FROM %I
+          WHERE ingest_time > v_latest_minute_already_aggregated
+          GROUP BY zone_id, minute
```
marcocitus: needs a "RETURNING ingest_time"

ref-arch/real-time-analytics.rst
```diff
@@ -0,0 +1,509 @@
...
+        ) as inserted_rows
+      SELECT max(minute) INTO v_new_latest_already_aggregated FROM inserted_rows;
```
marcocitus: minute -> ingest_time

ref-arch/real-time-analytics.rst
```diff
@@ -0,0 +1,509 @@
...
+
+In order to populate ``http_request_1min`` we're going to periodically run the equivalent
+of an INSERT INTO SELECT. Citus doesn't yet support INSERT INTO SELECT on distributed
+tables, so instead we'll run a function on all the workers which runs INSERT INTO SELECT
+on every matching pair of shards (this is possible because the shards are colocated)
+
+.. code-block:: plpgsql
+
+  -- this should be run on each worker
+  CREATE FUNCTION rollup_1min(source_shard text, dest_shard text) RETURNS void
```
marcocitus: Bit of annoying PL/pgSQL quirk, but you can't have parameters with the same name as columns that you'll use in queries.

begriffs:
There are two places in the top-level table of contents where Technical Solutions might fit in better. It depends on what we consider the purpose of the section to be.

* If the section is meant as a bridge from the tutorials to bigger, more realistic problems then it might be good to put it between Tutorials and Installation.
* If however the section is intended as a place to walk people through esoteric situations they may encounter after using Citus for a while then maybe it should go down between Administration and Reference.

ozgune:
@lithp -- I'm adding my notes to this pull request instead of #153.

index.rst
```diff
@@ -30,6 +30,11 @@ topics.
    installation/production.rst
 
 .. toctree::
+   :caption: Technical Solutions
+
+   tech_soln/real_time_analytics.rst
```
ozgune: real_time_analytics -> real_time_dashboards

tech_soln/real_time_analytics.rst
```diff
@@ -0,0 +1,425 @@
+.. _introduction:
+
+Real Time Analytics
```
ozgune: Could we use the following for the opening?

Real Time Dashboards
#####################

Citus provides its users real-time responses over large datasets. One workload that
we commonly see at Citus involves powering real-time dashboards over events data.

For example, you could be a cloud services provider helping other businesses serve their
HTTP traffic. In this setup, every time one of your clients receives an HTTP request, you
receive a log record. You want to ingest all of those records and create an HTTP
analytics dashboard which gives your clients insights, such as the number HTTP errors
their site served. It's important that this data show up with as little latency as
possible so your clients can fix problems with their sites. It's also important to show
graphs of historical trends through a dashboard to your clients.

tech_soln/real_time_analytics.rst
```diff
@@ -0,0 +1,425 @@
...
+
+There will be code snippets in this tutorial but they don't specify a complete system.
+There's `a github repo <https://github.com/citusdata/technical-solution-resources>`_ with
+all the details in one place. If you've followed our installation instructions for running
+Citus on either a single or multiple machines you're ready to try it out.
+
+Data Model
+----------
+
+The data we're dealing with is an immutable stream of log data. We'll insert directly into
```
ozgune: What do you think about the following?

The data we're dealing with is an immutable stream of log data. Here we'll insert directly
into Citus. If you're ingesting billions of log records per day into your cluster, you
could also consider micro-batching these records in a distributed message queue, such as
Kafka.

tech_soln/real_time_analytics.rst
```diff
@@ -0,0 +1,425 @@
...
+.. NOTE::
+
+  In Citus Cloud you must use a replication factor of 1 (instead of the 2 used here). As
+  Citus Cloud uses `streaming replication
+  <https://www.postgresql.org/docs/current/static/warm-standby.html>`_ to achieve high
+  availability maintaining shard replicas would be redundant.
+
+With this, the system is ready to accept data and serve queries! We've provided `a data
+ingest script
+<https://github.com/citusdata/technical-solution-resources/blob/master/ingest_example_data.sql>`
```
ozgune: This url and text don't correctly display on my browser.

tech_soln/real_time_analytics.rst
```diff
@@ -0,0 +1,425 @@
...
+  SELECT master_create_distributed_table('http_request', 'site_id', 'hash');
+  SELECT master_create_worker_shards('http_request', 16, 2);
+
+When we call :ref:`master_create_distributed_table <master_create_distributed_table>`
+we ask Citus to hash-distribute ``http_request`` using the ``site_id`` column. That means
+all the data for a particular site will live in the same shard.
+
+When we call :ref:`master_create_worker_shards <master_create_worker_shards>` we tell
+Citus to create 16 shards and 2 replicas of each shard (for a total of 32 shard replicas).
+:ref:`We recommend <faq_choose_shard_count>` using 2-4x as many shards as CPU cores in
```
ozgune: Minor. What do you think about moving the link to `using 2-4x as many shards <faq_choose_shard_count>`?

tech_soln/real_time_analytics.rst
```diff
@@ -0,0 +1,425 @@
...
+  -- this will create the index on all shards
+  CREATE INDEX ON http_requests_1min (zone_id, ingest_time);
+
+This looks a lot like the previous code block. Most importantly: It also shards on
+``site_id`` and it also uses 16 shards with 2 replicas of each. Because all three of
+those match, there's a 1-to-1 correspondence between ``http_request`` shards and
+``http_request_1min`` shards, and Citus will place matching shards on the same worker.
+This is called colocation; it makes queries such as joins faster and our rollups possible.
+
+.. image:: /images/colocation.png
```
ozgune: We have an example image on our blog that we use to demonstrate colocation.

https://www.citusdata.com/blog/2016/08/10/sharding-for-a-multi-tenant-app-with-postgres

What's your take on that image?

tech_soln/real_time_analytics.rst
```diff
@@ -0,0 +1,425 @@
...
+  :alt: colocation in citus
+
+In order to populate ``http_request_1min`` we're going to periodically run the equivalent
+of an INSERT INTO SELECT. Citus `doesn't yet support
+<https://github.com/citusdata/citus/issues/508>`_ INSERT INTO SELECT on distributed
+tables, so instead we'll run a function on all the workers which runs INSERT INTO SELECT
+on every matching pair of shards. This is possible because the tables are colocated.
+
+.. code-block:: plpgsql
+
```
ozgune: Minor. We previously advised running this on the workers or master. Did this change?

tech_soln/real_time_analytics.rst
```diff
@@ -0,0 +1,425 @@
...
+      JOIN pg_dist_shard_placement p ON (a.shardid = p.shardid)
+      WHERE a.logicalrelid = left_table AND b.logicalrelid = right_table;
+    $$ LANGUAGE 'sql';
+
+Using that metadata, every minute it runs a script which calls ``rollup_1min`` once for
+each pair of shards:
+
+.. code-block:: bash
+
+   #!/usr/bin/env bash
```
ozgune: This was simpler in the previous pass. Didn't it work?

tech_soln/real_time_analytics.rst
```diff
@@ -0,0 +1,425 @@
...
+keep raw data for one day and 1-minute aggregations for one month.
+
+.. code-block:: plpgsql
+
+  -- another function for the master
+  CREATE FUNCTION expire_old_request_data() RETURNS void
+  AS $$
+    SET LOCAL citus.all_modification_commutative TO TRUE;
+    SELECT master_modify_multiple_shards(
+      'DELETE FROM http_request WHERE ingest_time < now() - interval ''1 dat'';');
```
ozgune: 1 dat -> 1 day typo

tech_soln/real_time_analytics.rst
```diff
@@ -0,0 +1,425 @@
...
+it to deduplicate. That's both a lot of network traffic and a lot of computation. By using
+HLLs you can greatly improve query speed.
+
+First you must install the hll extension; `the github repo
+<https://github.com/aggregateknowledge/postgresql-hll>`_ has instructions. Next, you have
+to enable it:
+
+.. code-block:: sql
+
+  -- this part must be run on all workers
```
ozgune: Do we need to run this on the master too?

tech_soln/real_time_analytics.rst
```diff
@@ -0,0 +1,425 @@
...
+  CREATE EXTENSION hll;
+
+  -- this part runs on the master
+  ALTER TABLE http_requests_1min ADD COLUMN distinct_ip_addresses (hll);
+
+When doing our rollups, we can now aggregate sessions into an hll column with queries
+like this:
+
+.. code-block:: sql
```
ozgune: Will this work?

tech_soln/real_time_analytics.rst
```diff
@@ -0,0 +1,425 @@
...
+    sfunc = hll_union_agg,
+    stype = internal,
+  );
+
+Now, when you call SUM over a collection of HLLs, PostgreSQL will return the HLL for us.
+You can then compute distinct ip counts over a time period with the following query:
+
+.. code-block:: sql
+
+  -- working version of the above query
```
ozgune: Do we need this line?

tech_soln/real_time_analytics.rst
```diff
@@ -0,0 +1,425 @@
...
+  FROM http_request_1min
+  WHERE ingest_time BETWEEN timestamp '06-01-2016' AND '06-28-2016';
+
+You can find more information on HLLs `in the project's GitHub repository
+<https://github.com/aggregateknowledge/postgresql-hll>`_.
+
+Where the HLL extension provides distinct counts, there are more extensions which do a
+similar thing (improve performance and storage requirements) for other kinds of queries,
+such as `count-min sketch <https://github.com/citusdata/cms_topn>`_ for top-n queries, and
+`HDR <https://github.com/citusdata/HDR>`_, for percentile queries.
```
ozgune: We aren't going to be releasing CMS topN and HDR in this release cycle. Could we remove it?

tech_soln/real_time_analytics.rst
```diff
@@ -0,0 +1,425 @@
...
+volume is also high, and both historical and live data are important.
+
+In this technical solution we'll demonstrate how to build part of the first example,
+but this architecture would work equally well for the second and many other use-cases.
+
+Running It Yourself
+-------------------
+
+There will be code snippets in this tutorial but they don't specify a complete system.
+There's `a github repo <https://github.com/citusdata/technical-solution-resources>`_ with
```
ozgune: Do we want to have this technical solution be the only one in the technical-solution-resources repo? Or do we want to maintain a folder structure for upcoming technical solutions?

tech_soln/real_time_analytics.rst
lithp: I'm not a complete fan of the current image. It focuses on workers and doesn't give any intuition on how shards work, but it's certainly better than the blog's image!

The table names are different; that's a bit jarring. 

It's also difficult to parse, the whole point of colocation is that items with the same key are kept in the same place. With the current image, that's shown visually and is very difficult to misinterpret. With the blog's image, it's in a table of numbers. It's a visual aid but the key point isn't shown visually! I know what it's trying to show and it still took me a minute to figure out how colocation was even involved :)

tech_soln/real_time_analytics.rst
lithp: The one in the previous pass used dblink if I remember correctly.

tech_soln/real_time_analytics.rst
lithp: I suppose, but why do you want to phrase it like this? Kafka is useful for way more than just enabling batching, and the way this is phrased makes it sound like Kafka itself does the batching.

I had originally thought you just wanted me to trim down how much time I spent talking about why Kafka might be useful. I could just get rid of the reference to Kafka entirely.

tech_soln/real_time_analytics.rst
lithp: Ah, I had thought they were already released.

tech_soln/real_time_dashboards.rst
```diff
@@ -0,0 +1,422 @@
...
+volumes become unmanageably high.
+
+We'll use a simple schema for ingesting HTTP event data. This schema serves as an example
+to demonstrate the overall architecture; a real system might use additional columns.
+
+.. code-block:: sql
+
+  -- this is run on the master
+  CREATE TABLE http_request (
+    site_id INT,
```
byucesoy: I think this should be zone_id, because we refer this field as zone_id in rest of the docs and the scripts in https://github.com/citusdata/technical-solution-resources.

tech_soln/real_time_dashboards.rst
```diff
@@ -0,0 +1,422 @@
...
+    ingest_time TIMESTAMPTZ DEFAULT now(),
+
+    url TEXT,
+    request_country TEXT,
+    ip_address INET,
+
+    status_code INT,
+    response_time_msec INT
+  );
+  SELECT master_create_distributed_table('http_request', 'site_id', 'hash');
```
byucesoy: Same with above

tech_soln/real_time_dashboards.rst
```diff
@@ -0,0 +1,422 @@
...
+.. code-block:: sql
+
+  SELECT
+    date_trunc('minute', ingest_time) as minute,
+    COUNT(1) AS request_count,
+    SUM(CASE WHEN (status_code between 200 and 299) THEN 1 ELSE 0 END) as success_count,
+    SUM(CASE WHEN (status_code between 200 and 299) THEN 0 ELSE 1 END) as error_count,
+    SUM(response_time_msec) / COUNT(1) AS average_response_time_msec
+  FROM http_request
+  WHERE zone_id = 1 AND minute > date_trunc('minute', now()) - interval '5 minutes'
```
byucesoy: I think it is not possible to use ALIAS in WHERE clause. It gives me error; 
ERROR:  column "minute" does not exist

Related: https://www.postgresql.org/message-id/18195.1038530280@sss.pgh.pa.us
I am not sure if this is still the case, but probably it is.

tech_soln/real_time_dashboards.rst
```diff
@@ -0,0 +1,422 @@
...
+.. NOTE::
+
+  In Citus Cloud you must use a replication factor of 1 (instead of the 2 used here). As
+  Citus Cloud uses `streaming replication
+  <https://www.postgresql.org/docs/current/static/warm-standby.html>`_ to achieve high
+  availability maintaining shard replicas would be redundant.
+
+With this, the system is ready to accept data and serve queries! We've provided `a data
+ingest script
+<https://github.com/citusdata/realtime-dashboards-resources/blob/master/ingest_example_data.sql>`_
```
byucesoy: Data ingest script tries to ingest data as if table has session_id column, but it is not mentioned anywhere in the reference architecture document. Either add it to here or remove from data ingest script.

tech_soln/real_time_dashboards.rst
```diff
@@ -0,0 +1,422 @@
...
+  CREATE TABLE http_request_1min (
+        site_id INT,
+        ingest_time TIMESTAMPTZ, -- which minute this row represents
+
+        error_count INT,
+        success_count INT,
+        request_count INT,
+        average_response_time_msec INT,
+        CHECK (request_count = error_count + success_count),
+        CHECK (ingest_time = date_trunc('minute', ingest_time)),
```
byucesoy: Typo: comma at the end

tech_soln/real_time_dashboards.rst
```diff
@@ -0,0 +1,422 @@
...
+  );
+  SELECT master_create_distributed_table('http_requests_1min', 'site_id', 'hash');
```
byucesoy: Typo: It should be http_request_1min without 's'. It exists in next two lines also.

tech_soln/real_time_dashboards.rst
```diff
@@ -0,0 +1,422 @@
...
+  SELECT master_create_worker_shards('http_requests_1min', 16, 2);
+  
+  -- indexes aren't automatically created by Citus
+  -- this will create the index on all shards
+  CREATE INDEX ON http_requests_1min (zone_id, ingest_time);
```
byucesoy: We need to give index name, otherwise citus complains: "creating index without a name on a distributed table is currently unsupported"

lithp:
@byucesoy Just pushed a commit which I think handles all of these.

tech_soln/real_time_dashboards.rst
```diff
@@ -0,0 +1,422 @@
...
+        ) SELECT
+          site_id,
+          date_trunc('minute', ingest_time) as minute,
+          COUNT(1) as request_count,
+          SUM(CASE WHEN (status_code between 200 and 299) THEN 1 ELSE 0 END) as success_count,
+          SUM(CASE WHEN (status_code between 200 and 299) THEN 0 ELSE 1 END) as error_count,
+          SUM(response_time_msec) / COUNT(1) AS average_response_time_msec
+        FROM %1$I
+        WHERE
+          date_trunc('minute', ingest_time) > (SELECT max(ingest_time) FROM %2$I)
```
byucesoy: This SELECT returns nothing if table is empty.

tech_soln/real_time_dashboards.rst
```diff
@@ -0,0 +1,422 @@
...
+  -- this will create the index on all shards
+  CREATE INDEX http_request_1min_idx ON http_request_1min (site_id, ingest_time);
+
+This looks a lot like the previous code block. Most importantly: It also shards on
+``site_id`` and it also uses 16 shards with 2 replicas of each. Because all three of
+those match, there's a 1-to-1 correspondence between ``http_request`` shards and
+``http_request_1min`` shards, and Citus will place matching shards on the same worker.
+This is called colocation; it makes queries such as joins faster and our rollups possible.
+
+.. image:: /images/colocation.png
```
byucesoy: Column name in the image refers zone_id

tech_soln/real_time_dashboards.rst
lithp: Just pushed a fix :)

tech_soln/real_time_dashboards.rst
```diff
@@ -0,0 +1,422 @@
...
+  CREATE FUNCTION expire_old_request_data() RETURNS void
+  AS $$
+    SET LOCAL citus.all_modification_commutative TO TRUE;
+    SELECT master_modify_multiple_shards(
+      'DELETE FROM http_request WHERE ingest_time < now() - interval ''1 day'';');
+    SELECT master_modify_multiple_shards(
+      'DELETE FROM http_request_1min WHERE ingest_time < now() - interval ''1 month'';');
+  END;
+  $$ LANGUAGE 'sql';
+
```
byucesoy: Calling this function gives me error;

ERROR:  COMMIT is not allowed in a SQL function
CONTEXT:  SQL function "expire_old_request_data" during startup

I think it is because master_modify_multiple_shards uses COMMIT

tech_soln/real_time_dashboards.rst
lithp: nice catch, fixed

tech_soln/real_time_dashboards.rst
lithp: ... :cry:

lithp: https://github.com/citusdata/citus/issues/704 can't come fast enough

lithp: Interestingly, it works for me... I'll bug you when you come back in tomorrow

ozgune:
@lfittl -- I just checked with @begriffs. Putting this as a section after Cloud and before Administration sounds good.

lfittl:
@ozgune I assume you meant @lithp?

tech_soln/real_time_dashboards.rst
```diff
@@ -0,0 +1,423 @@
...
+     * * * * * /some/path/run_rollups.sh
+
+The dashboard query from earlier is now a lot nicer:
+
+.. code-block:: sql
+
+  SELECT
+    request_count, success_count, error_count, average_response_time_msec
+  FROM http_request_1min
+  WHERE site_id = 1 AND minute > date_trunc('minute', now()) - interval '5 minutes';
```
byucesoy: We should change minute to something like date_trunc('minute', ingest_time)

tech_soln/real_time_dashboards.rst
```diff
@@ -0,0 +1,423 @@
...
+<https://github.com/aggregateknowledge/postgresql-hll>`_ has instructions. Next, you have
+to enable it:
+
+.. code-block:: sql
+
+  -- this part must be run on all nodes
+  CREATE EXTENSION hll;
+
+  -- this part runs on the master
+  ALTER TABLE http_request_1min ADD COLUMN distinct_ip_addresses (hll);
```
byucesoy: I think we should not have parentheses around hll.

tech_soln/real_time_dashboards.rst
```diff
@@ -0,0 +1,423 @@
...
+When doing our rollups, we can now aggregate sessions into an hll column with queries
+like this:
+
+.. code-block:: sql
+
+  SELECT
+    site_id, date_trunc('minute', ingest_time) as minute,
+    hll_add_agg(hll_hash_text(ip_address)) AS distinct_ip_addresses
+  WHERE minute = date_trunc('minute', now())
+  FROM http_request
```
byucesoy: *FROM should come before WHERE condition.
*We should remove minute reference from WHERE statement.
*hll_hash_text function expects text parameter but type of ip_address is inet.

tech_soln/real_time_dashboards.rst
```diff
@@ -0,0 +1,423 @@
...
+  GROUP BY site_id, minute;
+
+Now dashboard queries are a little more complicated, you have to read out the distinct
+number of ip addresses by calling the ``hll_Cardinality`` function:
```
byucesoy: typo: hll_Cardinality -> hll_cardinality

tech_soln/real_time_dashboards.rst
```diff
@@ -0,0 +1,423 @@
...
+  ALTER TABLE http_request_1min ADD COLUMN distinct_ip_addresses (hll);
+
+When doing our rollups, we can now aggregate sessions into an hll column with queries
+like this:
+
+.. code-block:: sql
+
+  SELECT
+    site_id, date_trunc('minute', ingest_time) as minute,
+    hll_add_agg(hll_hash_text(ip_address)) AS distinct_ip_addresses
```
byucesoy: Maybe we should write whole rollup function here instead of just new addition and then we can highlight new line to make it clear that it is mostly same except that hll line. In that way;

-This would be full example and it would be more clear that this is our new rollup function.
-A user who follows this document step by step, would have full version of the new function and can continue to play without needing to do any changes.

tech_soln/real_time_dashboards.rst
```diff
@@ -0,0 +1,423 @@
...
+We have `a blog post
+<https://www.citusdata.com/blog/2016/07/14/choosing-nosql-hstore-json-jsonb/>`_ explaining
+which format to use for your semi-structured data. The post recommends JSONB, here we'll
+demonstrate how to incorporate JSONB columns into your data model.
+
+First, add the new column to our rollup table:
+
+.. code-block:: sql
+
+  ALTER TABLE http_request_1min ADD COLUMN country_counters (JSONB);
```
byucesoy: We should remove parentheses.

tech_soln/real_time_dashboards.rst
```diff
@@ -0,0 +1,423 @@
...
+    site_id, minute,
+    jsonb_object_agg(request_country, country_count)
+  FROM (
+    SELECT
+      site_id, date_trunc('minute', ingest_time) as minute,
+      request_country,
+      count(1) AS country_count
+    FROM http_request
+    GROUP BY site_id, minute, request_country
+  )
```
byucesoy: This FROM clause should have ALIAS or it gives error;
ERROR:  subquery in FROM must have an alias

tech_soln/real_time_dashboards.rst
```diff
@@ -0,0 +1,423 @@
...
+
+  ALTER TABLE http_request_1min ADD COLUMN country_counters (JSONB);
+
+Next, include it in the rollups by adding a clause like this to the rollup function:
+
+.. code-block:: sql
+
+  SELECT
+    site_id, minute,
+    jsonb_object_agg(request_country, country_count)
```
byucesoy: This function call causes an error. I think it is because citus does not support custom aggregate functions(?)
ERROR:  unsupported aggregate function jsonb_object_agg

byucesoy:
Apart from end-to-end tests, I also tried some failure test as @onderkalaci's suggestion.

While we ingest data, I failed one of the workers.

**Before restarting failed node;**
* Remaining worker continued to get new data. :white_check_mark: 
* Rollups are done in remaining worker without any problem. :white_check_mark: 
* SELECT queries are answered from remaining shard :white_check_mark: 
* Placement on failed worker does not marked as invalid :x:

I think this one is problematic because when we restart failed worker, we will get wrong/old data from this worker


**After restarting failed node;**
* There is no sync between workers :question:

I think this would be nice to have, but it may be not in the scope of this reference architecture
* New data are sent to restarted worker :question:  

After restart, user intervention may be required to continue to send data (like manualy syncing workers)
* Rollups also run in restarted worker :x:

We end up incomplete/wrong aggregations

byucesoy:
I'm merging this change based on request of team @ozgune amd @craigkerstiens. We will fix remaining items on iteration basis.

byucesoy:
@craigkerstiens I merged this change but what should I do to generate temporary (and eventually permanent deployment)?

PR #119, opened 2016-07-27
Document common sql workarounds

begriffs:
Subqueries in where, select into, non-distribution column count distinct, join local and distributed.

When talking with @sumedhpathak we determined that hash partitioning on composite keys is actually more of a shard balancing technique than it is a SQL workaround.

### DIFF
--- a/reference/sql_workarounds.rst
+++ b/reference/sql_workarounds.rst
@@ -3,6 +3,8 @@
 SQL Workarounds
 ===============
 
+Before attempting workarounds consider whether Citus is appropriate for your situation. Citus' current version works well for real-time analytics use cases. We are continuously working to increase SQL coverage to better support other use-cases such as data warehousing queries.
+
 Citus supports most, but not all, SQL statements directly. Its SQL support continues to improve. Also many of the unsupported features have workarounds; below are a number of the most useful.
 
 Subqueries in WHERE
@@ -195,3 +197,45 @@ Now Citus will accept a join query between *here* and *there*, and each worker w
 .. note::
 
   Citus Cloud uses PostgreSQL replication, not Citus replication, so this technique does not work there.
+
+.. _data_warehousing_queries:
+
+Data Warehousing Queries
+------------------------
+
+When queries have restrictive filters (i.e. when very few results need to be transferred to the master) there is a general technique to run unsupported queries in two steps. First store the results of the inner queries in regular PostgreSQL tables on the master. Then the next step can be executed on the master like a regular PostgreSQL query.
+
+For example, currently Citus does not have out of the box support for window functions on queries involving distributed tables. Suppose you have a query with a window function on a table of github_events function like the following:
+
+::
+
+    select repo_id, actor->'id', count(*)
+      over (partition by repo_id)
+      from github_events
+     where repo_id = 1 or repo_id = 2;
+
+You can re-write the query like below:
+
+Statement 1:
+
+::
+
+    create temp table results as (
+      select repo_id, actor->'id' as actor_id
+        from github_events
+       where repo_id = 1 or repo_id = 2
+    );
+
+Statement 2:
+
+::
+
+    select repo_id, actor_id, count(*)
+      over (partition by repo_id)
+      from results;
+
+Similar workarounds can be found for other data warehousing queries involving unsupported constructs.
+
+.. Note::
+
+  The above query is a simple example intended at showing how meaningful workarounds exist around the lack of support for a few query types. Over time, we intend to support these commands out of the box within Citus.
--- a/reference/sql_workarounds.rst
+++ b/reference/sql_workarounds.rst
@@ -59,6 +59,31 @@ In this technique we use PL/pgSQL to construct and execute one statement based o
   select * from results_temp;
   drop table results_temp;
 
+Workaround 1½. Build query in SQL client
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+* Pros
+    * Able to distribute the query across shards
+* Cons
+    * Constructed query not suitable when facts subquery returns large number of results
+    * Requires two query roundtrips to client
+
+Like the previous workaround this one creates an explicit list of values for an IN comparison. The client obtains the list of items with one query, and uses it to construct the second query.
+
+.. code-block:: sql
+
+  -- first run this
+  select id from facts where ...conditions...
+
+Interpolate the list of ids into a new query
+
+.. code-block:: sql
+
+  -- notice the explicit list of ids obtained from previous query
+  select dim_a, dim_b, dim_c
+    from dimensions
+   where fact_id in (2,3,5,7,11,13)
+     and ...conditions...;
 
 Workaround 2. Duplicate on Master
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
@@ -67,7 +92,6 @@ Workaround 2. Duplicate on Master
     * Works for subqueries returning any number of results
 * Cons
     * Must transmit full rows from the queries back to the master
-    * Requires extra storage space in the master database
 
 In this workaround the client runs the outer- and sub-query independently, saves their results, and joins them.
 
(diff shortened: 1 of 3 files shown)


reference/sql_workarounds.rst
```diff
@@ -0,0 +1,141 @@
...
+.. code-block:: bash
+
+  psql -c "COPY (query) TO STDOUT" | psql -c "COPY table FROM STDIN"
+
+This does incur network cost. If this workaround is too slow please contact Citus Data support. We can assist you in parallelizing the table insertion across all workers using a more complicated technique.
+
+SELECT DISTINCT (non-distribution column)
+-----------------------------------------
+
+Citus supports SELECT DISTINCT col but only when col is the shard distribution column. If it is not, use GROUP BY for a simple workaround:
```
saicitus: SELECT DISTINCT col is not supported even when col is the distribution column. Could you correct this?

reference/sql_workarounds.rst
```diff
@@ -0,0 +1,141 @@
...
+      array_to_string(facts_array, ','));
+  end;
+  $$;
+
+  -- read results, remove temp table
+  select * from results_temp;
+  drop table results_temp;
+
+
+Workaround 2. Join on Master
```
saicitus: You don't require a join here. You could directly run a subquery in where for the local temp tables which you create. It would work, because the local tables wouldn't be distributed.

reference/sql_workarounds.rst
begriffs: Are you sure?

```sql
create table foo(id bigint, bar integer);

SELECT master_create_distributed_table('foo','id','hash');

SELECT master_create_worker_shards('foo', 16, 1);

insert into foo (id, bar) values (1,2);
insert into foo (id, bar) values (2,2);
insert into foo (id, bar) values (3,2);
insert into foo (id, bar) values (4,1);
insert into foo (id, bar) values (5,3);

select count(distinct id) from foo;
 count
-------
     5
(1 row)

select count(distinct bar) from foo;
ERROR:  cannot compute aggregate (distinct)
DETAIL:  table partitioning is unsuitable for aggregate (distinct)
HINT:  You can load the hll extension from contrib packages and enable distinct approximations.
```

begriffs: Oh right, I was thinking *count* distinct. I'll fix it.

reference/sql_workarounds.rst
```diff
@@ -0,0 +1,141 @@
...
+
+SQL Workarounds
+===============
+
+Citus supports most, but not all, SQL statements directly. Its SQL support continues to improve. Also many of the unsupported features have workarounds; below are a number of the most useful.
+
+Subqueries
+----------
+
+In analytics it is common to model information using a "star schema." This is a collection of dimensional tables (e.g. observations, statistics) referencing a table of facts (e.g. people, pages or events). Furthermore, a common form of query examines dimensions for a subset of the facts. Examples of these kind of queries:
```
saicitus: Im not sure whether it is a good idea to give the fact-dimension table example. In customer use-cases we have not seen this. We have seen this in more complex cases, such as on the same table or between 2 large tables which are not colocated on the column which has a filter on subquery, etc.  In the fact-dimension table example, the most simplest workaround is by a broadcast join because dimension tables are most of the times small.

One idea which comes to my mind to rewrite this description, is by writing the actual queries for each customer use-cases and mention that these are some examples where subqueries in where clause kicks in. 

And in the actual work arounds, take one of the examples and give workarounds for that example.
@samay-sharma @sumedhpathak : Could you give your inputs on this?

reference/sql_workarounds.rst
```diff
@@ -0,0 +1,145 @@
...
+.. code-block:: sql
+
+  select dim_a, dim_b, dim_c
+    from dimensions
+   where fact_id in (select id from facts where ...conditions...)
+     and ...conditions...;
+
+Citus does not allow subqueries in the WHERE clause so we must choose a workaround.
+
+Workaround 1. Generate explicit WHERE-IN expression
```
saicitus: I couldn't see the in-lining workaround in which in the application layer you could inline the intermediary results to the main query. This is most suitable when the intermediary results (results of the subquery) are less in number(ex. subquery having a limit etc). Refer this [doc](https://docs.google.com/document/d/1Rmkb5V8-Cwt2jTPrVbzre58Zh0wNdAgUvxX6vMzQRns/edit?usp=gmail) for that work around. And also there are some customers in production who use this workaround.

reference/sql_workarounds.rst
```diff
@@ -0,0 +1,145 @@
...
+  select * from results_temp;
+  drop table results_temp;
+
+
+Workaround 2. Duplicate on Master
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+* Pros
+    * Works for subqueries returning any number of results
+* Cons
```
saicitus: I think the first con is sufficient, because from it is understood that the master should have the storage to cater the temp tables.
You could also mention that this method is ideal when the temp tables are small(when both the queries have significant filters)

reference/sql_workarounds.rst
begriffs: I thought this version is perhaps superior to the inlining technique because the latter takes more time to stream results to the client whereas this one all happens in the db.

reference/sql_workarounds.rst
```diff
@@ -0,0 +1,145 @@
...
+  -- Run the query on local tables where subqueries are OK
+  select dim_a, dim_b, dim_c
+    from dim_temp
+   where fact_id in (select id from fact_temp);
+
+  -- Remove temp tables
+  drop table dim_temp;
+  drop table fact_temp;
+
+SELECT ... INTO
```
saicitus: I guess the standard query is INSERT INTO SELECT. Refer the [insert documentation](https://www.postgresql.org/docs/9.5/static/sql-insert.html) of postgresql 9.5 which talks about VALUES in insert could be taken from a SQL query.

reference/sql_workarounds.rst
saicitus: If the intermediary results are less, the streaming to application layer wouldn't be an issue and also that the main query wouldn't be too large.
And also, the inlining workaround is far simpler(to understand and implement) than the other ones.

reference/sql_workarounds.rst
begriffs: I see now how fact-dimension is the wrong scenario because it can be accomplished best by broadcasting the small table. @saicitus if you can email me examples of the queries for the scenarios listed in this document then I'll have a better idea of the actual schemas and can write more accurately.

reference/sql_workarounds.rst
begriffs: I added the application layer sql construction workaround.

reference/sql_workarounds.rst
saicitus: @begriffs  I sent the email. We came up with some common 3 patterns(2 table - preventing repartition joins, single users table and segmenting the users based on 2 events which cannot be mapped to 1 row, single table and doing aggregates/lookups over first n[based on another aggregate] rows  )when we see Subqueries in Where clause. Could you edit the documentation based on these ideas?

reference/sql_workarounds.rst
```diff
@@ -0,0 +1,172 @@
...
+      '   and ...conditions...',
+      array_to_string(facts_array, ','));
+  end;
+  $$;
+
+  -- read results, remove temp table
+  select * from results_temp;
+  drop table results_temp;
+
+Workaround 1½. Build query in SQL client
```
saicitus: Could you make this as a separate work around instead of 1.5. 1.5 is too mathematical.

reference/sql_workarounds.rst
```diff
@@ -0,0 +1,172 @@
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+* Pros
+    * Able to distribute the query across shards
+* Cons
```
saicitus: **Requires two query roundtrips to client**
I think this is understood by the first con. So it may be better to remove this con.

begriffs:
@samay-sharma it's ready for another review.

reference/sql_workarounds.rst
begriffs: I called it 1½ because it's closely related to 1. But I can see how that numbering looks odd. I'll change it.

reference/sql_workarounds.rst
begriffs: I added this con to help distinguish the workaround from the previous one, just to make it clear at a glance.

begriffs:
Note, the "insert into select" documentation covers part of #18

begriffs:
@saicitus does this match up with what you had in mind?

begriffs:
@saicitus Craig suggests that we remove the subqueries-in-where case and merge the others while we deliberate on how best to express that one. If you'd like to do that while I'm gone go right ahead, otherwise we can resume this PR on Tuesday when I'm back in the office.

PR #140, opened 2016-08-10
Fix minor typos

metdos:


begriffs:
Thanks, good catch.

Looks like there's a DELETE statement referencing 2014 that could be updated too:
```sql
SELECT master_modify_multiple_shards(
  'DELETE FROM github_events WHERE created_at <= ''2014-01-01 00:00:00''');
```

metdos:
> Looks like there's a DELETE statement referencing 2014 that could be updated too:

Fixed also that one.

PR #152, opened 2016-08-12
Mention shardalias is deprecated, unused, and slated for future removal

robin900:


robin900:
Related to citusdata/citus#739. For release 5.2.

PR #155, opened 2016-08-22
Mark that shardalias is deprecated

lithp:
If you don't want me to bug you with changes this small let me know :)

As of https://github.com/citusdata/citus/issues/739 we no longer respect `shardalias`.


sumedhpathak:
@lithp We already have an open PR created as part of https://github.com/citusdata/citus/pull/739 here: #152. 

Could you review and merge that PR instead and close this one? Thanks!

PR #156, opened 2016-08-23
Remove append-distribution docs

### LINKED ISSUE #135: Remove Append partitioning sections in documentation
We are planning to support hash-partitioned use-cases, so we should remove all references to Append partitioning from Documentation.

### LINKED ISSUE #136: Remove Append-Distributed Data Tutorial
We are planning to focus on hash-partitioned use-cases. We should remove this Tutorial from the Documentation.

begriffs:
Closes #136 
Closes #135 

I chopped references to this distribution method. I did however leave information about it in the reference section. My idea there is that people may want to know the full story about our UDFs and metadata tables and we can remove the reference information when (if?) we remove append distribution support from Citus.

Let me know if the editing is overzealous...


sumedhpathak:
@begriffs I haven't looked at the changes in detail, but I had one high-level question:
We decided to not entirely remove this content, but create a separate section for it. Should we do this in the same PR?

begriffs:
Yeah if we're going to do that then this the PR to do it in. There is a lot of random information being removed, trying to think how to organize it all together in a separate section.

sumedhpathak:
Yep. Let's do that in this PR as we don't want to entirely excise it from the site. Do you have proposals on it? One option is to create a section way at the very end on 'partitioning by time', which has a brief blurb, followed by some reference links to this content. Any other thoughts?

begriffs:
@samay-sharma can you look this over? I moved the append stuff into a single section at the very bottom of the TOC in order to de-emphasize it.

reference/append.rst
```diff
@@ -3,29 +3,60 @@
 Append Distribution
 ###################
 
-Append distributed tables are best suited to append-only event data
-which arrives in a time-ordered series. In the next few sections, we describe how
-users can create append distributed tables, load data into them and also expire
-old data from them.
+While Citus' most common use cases involve hash data distribution,
```
sumedhpathak: Can we add a ``NOTE`` here, with some text around append being a very specialized method (or some such) for further discouragement?

sumedhpathak:
'Working with distributed tables section' still contains this text: "Before we dive into the syntax for these commands, we briefly discuss two important concepts which must be decided during distributed table creation: the distribution column and distribution method." We should amend this to discuss only the distribution column.

dist_tables/working_with_distributed_tables.rst
```diff
@@ -20,14 +20,3 @@ Typically, you should choose that column as the distribution column which is the
   Citus also allows joining on non-distribution key columns by dynamically repartitioning the tables for the query. Still, joins on non-distribution keys require shuffling data across the cluster and therefore aren’t as efficient as joins on distribution keys.
 
 The best option for the distribution column varies depending on the use case and the queries. In general, we find two common patterns: (1) **distributing by time** (timestamp, review date, order date), and (2) **distributing by identifier** (user id, order id, application id). Typically, data arrives in a time-ordered series. So, if your use case works well with batch loading, it is easiest to distribute your largest tables by time, and load it into Citus in intervals of N minutes. In some cases, it might be worth distributing your incoming data by another key (e.g. user id, application id) and Citus will route your rows to the correct shards when they arrive. This can be beneficial when most of your queries involve a filter or joins on user id or order id.
```
sumedhpathak: Can we amend this section as well, and remove the content around distribution by 'time'? I think we can have a short blurb here about 'distributing by tenant or application' and the other being 'distributing by user-id', and very briefly discuss pros/cons here.

sumedhpathak:
Minor: In the Upgrade section (Major upgrades) we have a line: "If you are upgrading the master, then you should stop all data-loading/appending and staging before copying out the metadata". Maybe modify this to say you should stop all 'data inserts/updates/deletes or COPY operations...".

performance/scaling_data_ingestion.rst
```diff
@@ -5,8 +5,6 @@ Scaling Out Data Ingestion
 
 Citus lets you scale out data ingestion to very high rates, but there are several trade-offs to consider in terms of the throughput, durability, consistency and latency. In this section, we discuss several approaches to data ingestion and give examples of how to use them.
 
-The best method to distribute tables and ingest your data depends on your use case requirements. Citus supports two distribution methods: append and hash; and the data ingestion methods differ between them. You can visit the :ref:`working_with_distributed_tables` section to learn about the tradeoffs associated with each distribution method.
-
 Hash Distributed Tables
```
sumedhpathak: Maybe let's remove this top-level heading on hash-distribution? We used to have another on 'Append', but now that its removed, no reason to keep this.

sumedhpathak:
In 'Performance Tuning' section, the first paragraph has this line: "We begin by explaining how choosing the right distribution column and method affect performance.". Amend to remove 'method'.

sumedhpathak:
Unrelated, but in the [General Performance](https://citus-doc-preview-pr-156.herokuapp.com/performance/performance_tuning.html#general) Tuning section, we have a paragraph on ``remote_task_check_interval``. Could we remove that?

sumedhpathak:
Looks good! 🚢

PR #166, opened 2016-09-06
Bump to Citus 5.2.1

jasonmp85:
Just updating the docs to the new 5.2.1 package.


begriffs:
We have the choice of cherry-picking this into the v5.2 branch, or creating a whole new branch v5.2.1 for this version in particular. My instinct is to add patch releases onto v5.2 for simplicity because if we're versioning semantically then there shouldn't be actual interface changes to document in patch releases, right?

jasonmp85:
I think putting this in `v5.2` is fine. Missed that that branch existed, since it was at the very bottom of the branch picker and other PRs were using `master`.

begriffs:
The master branch is fine, we've been merging all new things here and then cherry picking relevant stuff to the various versions.

PR #168, opened 2016-09-08
Document migrating to a multi-tenant app

### LINKED ISSUE #146: Create "Migration" section
Explains how to transition an existing postgres schema.
- For multi-tenant use case, add tenant id to all tables
- For analytics use case, ...

Our meeting notes indicate to put this section below the "Distributed Data Modeling" section.

Also remove "Transitioning From PostgreSQL to Citus" from the "Administration" section.

begriffs:
Fixes #146


begriffs:
Now it reads better. I added the new content in the "Transitioning From PostgreSQL to Citus" because it's a relevant place, although in the referenced issue we talked about removing this section. Trying to remember why.

admin_guide/transitioning_from_postgresql_to_citus.rst
```diff
@@ -16,4 +16,63 @@ directly pipe it to a copy into the distributed table. For example:
 
 One thing to note as you transition from a single node to multiple nodes is that you should create your extensions, operators, user defined functions, and custom data types on all nodes.
 
-If you have any questions or require assistance in scaling out your existing PostgreSQL installation, please get in touch with us at engage@citusdata.com.
+Multi-Tenant Applications
+-------------------------
+
+Citus is well suited to hosting B2B multi-tenant application data. All tenants share a schema and each tenant identifies its own rows in the tables by a pre-existing or artificially added distribution column, isolating each tenant's data in its own shard. This section will explore how to model for the multi-tenant scenario, including necessary adjustments to the schema and queries.
```
sumedhpathak: "isolating each tenant's data in its own shard." This isn't strictly true. Formally, *all* the data for a tenant will be localized to a single shard, however its not guaranteed that the shard will only contain that tenant. One shard will contain all data for potentially multiple tenants.

index.rst
```diff
@@ -38,6 +38,11 @@ topics.
    dist_tables/extensions.rst
 
 .. toctree::
+   :caption: Migration
```
sumedhpathak: I don't see this in the index in the temporary deployment btw. Are there other changes needed in the index in order to make it visible? https://citus-doc-preview-pr-168.herokuapp.com/

migration/transitioning.rst
```diff
@@ -0,0 +1,93 @@
+.. _transitioning_mt:
+
+Citus extends PostgreSQL with distributed functionality, but it is not a drop-in replacement that scales out all workloads. A performant Citus cluster involves thinking about the data model, tooling, and choice of SQL features used. Migrating an existing single-node PostgreSQL database to Citus sometimes requires adjusting the schema and queries for optimal performance.
+
+Migration tactics differ between the two main Citus use cases of multi-tenant applications and real-time analytics. The former is more clear cut so we'll begin there.
```
sumedhpathak: Maybe remove the comment around the migration being more 'clear cut'? We could instead say something like, the multi-tenant model requires fewer data model changes.

migration/transitioning.rst
```diff
@@ -0,0 +1,93 @@
...
+This section will explore how to model for the multi-tenant scenario, including necessary adjustments to the schema and queries.
+
+Schema Migration
+~~~~~~~~~~~~~~~~
+
+Migrating to a multi-tenant architecture requires a few changes to ensure each tenant's queries are routed to a single node. Once queries are routed this way (aka "pushed down") they can be executed without concern for the rest of the cluster. These queries can use the full features of SQL, including joins and transactions, without running into the inherent limitations of a distributed system.
+
+Tables and queries must simply include enough information that Citus knows where to place and route them. The first step to include this information is to examine existing tables and mentally classify each as either "global", "per-tenant," or "reference."
+
+*Global* tables aren't specific to individual tenants. They hold information about the system as a whole (e.g. admin tables for a website). Generally global tables keep to themselves and the application will not contain queries that join them with per-tenant tables nor will foreign key constraints hold between the two.
```
sumedhpathak: What is the recommendation for 'Global' tables?

migration/transitioning.rst
```diff
@@ -0,0 +1,93 @@
...
+
+On the other hand some tables may already have a column suitable to identify which tenant owns which rows. These are the natural *per-tenant* tables. Others don't have such a column, yet are not global because they are combined in queries with per-tenant tables. We call these *reference* tables and our task will be to adjust them for the multi-tenant use case.
```
sumedhpathak: This section probably needs some re-organizing. The recommendations for adjusting reference tables comes in the Example section when we talk about the Distributor table.

sumedhpathak: I had this recommendation for re-organizing the section:
H2. Multi-tenant data model migration:
      -> Overview of model
      -> Brief description of how Citus implements (basically Citus just routes everything to appropriate node).
            H3. Schema Migration:
            Add General Procedure section here as an intro. There are two types of tables (etc + more content)
                  H4. Tenant specific tales: 
                  -> Data which remains local to a tenant.
                  -> Add tenant-id to schema
                  H4. Reference tables
                   -> Three approaches: Denormalize, or keep on master as non-distributed.
           H3. Query migration
           -> Talk again about our executor, and how we route (to set context on why we need to modify queries)
           -> (See Sai's presentation on multi-tenant. He has some examples for joins, CTE's etc)

What do you think about that kind of a structure here? In each section we could provide examples. I like the below 'General Procedure', so we can start with that, and then go into each table type.

sumedhpathak: I also like the idea of providing an Example. What are your thoughts of using an example similar to what's in our blog post? If we can add a reference table to that, and show a before/after schema, that would be nice.

begriffs:
@sumedhpathak I updated the page to match the sharding multi-tenant app blog post, and included a realistic (I hope) reference table. Also removed repartition join as a "solution" to using reference tables. Squashed the commits in this PR as well.

My link to manage the Heroku documentation preview pipeline no longer works. Pinged @craigkerstiens to double check if my permissions have changed.

migration/transitioning.rst
```diff
@@ -0,0 +1,126 @@
...
+       AND purchases.store_id = regions.store_id
+
+This will run an efficient co-located join on the worker holding the shard for this store id.
+
+The second technique is to remove the regions table and denormalizing the database by embedding the sales_tax property directly into purchases. Duplicating just one property isn't too bad, but the situation isn't as nice for wider reference tables. Note that reference columns can also be stored in a single JSONB column in the denormalized table to prevent the latter from becoming awkwardly wide.
+
+Query Migration
+---------------
+
+To execute queries efficiently and isolate them within their tenant Citus needs to route them to a specific shard. Thus every query must identify which single tenant it involves. For non-joins this means that the *where* clause must filter by tenant id. In joins at least one of the tables must be filtered by tenant id. For instance:
```
sumedhpathak: ``In joins at least one of the tables must be filtered by tenant id.``. From speaking with @saicitus this isn't enough. With joins, either one table should be filtered by tenant-id and we add a join-clause between the two tables on tenant-id OR we need to filter both tables by tenant-id.

Sai's [presentation](https://docs.google.com/presentation/d/1HHIQSzEUscfzWcPWBGEmrANEqJEel5GR3IG65sVrvx8/edit#slide=id.g16a1d1e876_0_19) lists both options. Could you amend to include both options (with examples)?

dist_tables/ddl.rst
```diff
@@ -61,6 +61,16 @@ Each created shard is assigned a unique shard id and all its replicas have the s
 
 You are now ready to insert data into the distributed table and run queries on it. You can also learn more about the UDF used in this section in the :ref:`user_defined_functions` of our documentation.
 
+.. note::
```
sumedhpathak: This note doesn't belong here. AFAIK, we now error out if you try to distribute a table which already has some data in it. So we shouldn't have any pre-existing data. Maybe we can add a question in the FAQ to cover how to move data from a PostgreSQL table into a Citus distributed table?

migration/transitioning.rst
```diff
@@ -0,0 +1,126 @@
...
+
+Tenant-Specific Tables
+^^^^^^^^^^^^^^^^^^^^^^
+
+In our example each store is a natural tenant. This is because storefronts benefit from dedicated processing power for their customer data, and stores should not be able to access each other's sales or inventory. The tenant id is in this case the store id. Notice that some tables already have a store id as primary or foreign key. Tables such as these which can join on the tenant id are called *tenant-specific* or *per-tenant* tables. In this example they are stores, products and purchases.
+
+Global and Reference Tables
+^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The tenant-specific tables do not require modification for use in Citus. During migration you would distribute them through the Citus cluster, sharding by tenant id. The other tables — admins and regions — do require modification.
```
begriffs: The definition I was going with was that per-tenant tables are ones that already have the tenant id in the db to be migrated, whereas reference tables do not yet have it but will need it.
> Notice that some tables already have a store id as primary or foreign key. Tables such as these which can join on the tenant id are called *tenant-specific* or *per-tenant* tables.

In the example schema for instance we included the tenant id in stores, products and purchases. However in reality perhaps the schema to be migrated may not even have the store table yet because it was designed for only one store. The store table would need to be created and the store_id added to products and purchases.

Maybe we should remove stores from the initial schema to make this migration example more substantial.

migration/transitioning.rst
```diff
@@ -0,0 +1,92 @@
...
+
+.. image:: ../images/erd/mt-before.png
+
+In our example each store is a natural tenant. This is because storefronts benefit from dedicated processing power for their customer data, and stores do not need to access each other's sales or inventory. The tenant id is in this case the store id. We want to distribute data in the cluster in such a way that rows from the above tables in our schema reside on the same node whenever the rows share a store id.
+
+The first step is preparing the tables for distribution. Citus requires that primary keys contain the distribution column, so we must modify the primary keys of these tables and make them compound including a store id. Making primary keys compound will require modifying the corresponding foreign keys as well.
+
+In our example the products table is already in perfect shape. Stores and orders need slight modification: just updating the primary/foreign keys to include store_id. The line_items table needs the biggest change. Being normalized, it lacks a store id. We must add that column, and include it in the primary key constraint.
```
sumedhpathak: Looks like the `stores` table also has no modifications?

migration/transitioning.rst
```diff
@@ -0,0 +1,92 @@
...
+Another kind of table are those which join with per-tenant tables but which aren't naturally specific to any one tenant. We call them *reference* tables. Two examples are shipping regions and product categories. We advise that you add a tenant id to these tables and duplicate the original rows, once for each tenant. This ensures that reference data is co-located with per-tenant data and quickly accessible to queries.
+
+Data Migration
+--------------
+
+Once the schema is updated and the per-tenant and reference tables are distributed across the cluster it's time to copy data from the original database into Citus. Most per-tenant tables can be copied directly from source tables. However line_items was denormalized with the addition of the  store_id column. We should copy data not from the original line_items table but from the results of a query on the source database:
+
+.. code:: sql
+
+  SELECT orders.store_id AS store_id, line_items.*
```
sumedhpathak: This query is just a select query, and not a backfill query. @saicitus Had noted in Slack the query he had in mind:

```SQL
UPDATE line_items as l SET store_id=o.store_id from orders as o where l.order_id=o.order_id
```

Was there something else you had in mind with just noting a SELECT query here?

migration/transitioning.rst
```diff
@@ -0,0 +1,92 @@
...
+
+.. image:: ../images/erd/mt-after.png
+
+We call the tables considered so far *per-tenant* because querying them for our use case requires information for only one tenant per query. Their rows are distributed across the cluster according to the hashed values of their tenant ids.
+
+There are other types of tables to consider during a transition to Citus. Some are system-wide tables such as information about site administrators. We call them *global* tables and they do not participate in join queries with the per-tenant tables and may remain on the Citus coordinator node unmodified.
+
+Another kind of table are those which join with per-tenant tables but which aren't naturally specific to any one tenant. We call them *reference* tables. Two examples are shipping regions and product categories. We advise that you add a tenant id to these tables and duplicate the original rows, once for each tenant. This ensures that reference data is co-located with per-tenant data and quickly accessible to queries.
+
+Data Migration
```
sumedhpathak: Maybe let's just rename this to "Backfilling the tenant_id", since this isn't about data migration as much as it is about just backfill of the tenant_id.

Also, there is an extra space here: `denormalized with the addition of the  store_id column.` (between "the" and "store_id".)

migration/transitioning.rst
```diff
@@ -0,0 +1,92 @@
...
+  -- after
+  SELECT * FROM orders WHERE order_id = 123 AND store_id = 42;
+
+Likewise insert statements must always include a value for the tenant id column. Citus inspects that value for routing the insert command.
+
+When joining tables make sure they are all filtered by tenant id. For instance here is how to inspect how many awesome wool pants a given store has sold:
+
+.. code:: sql
+
+  SELECT sum(l.quantity)
```
sumedhpathak: We discussed two options for modifying join queries in our discussion with Sai. The first option is the one you've documented, but the other option was:

- Filter one of the tables on store_id *AND* Join both tables on store_id as well.

Could we document this as well, with the example query?

migration/transitioning.rst
begriffs: I would use the update statement if I were transforming a table in place on the existing database prior to copying it into the new database. The idea behind the select statement above is that it can be run on the old database and its result can be fed into the new database to populate the new table. Note that this select statement outputs line items along with their corresponding store ids.

migration/transitioning.rst
```diff
@@ -0,0 +1,92 @@
...
+3. Copy the data from the original database to Citus.
+
+Schema Migration
+----------------
+
+Transitioning from a standalone database instance to a sharded multi-tenant system requires identifying and modifying three types of tables which we may term *per-tenant*, *reference*, and *global*. The distinction hinges on whether the tables have (or reference) a column serving as tenant id. The concept of tenant id depends on the application and who exactly are considered its tenants.
+
+Consider an example multi-tenant application similar to Etsy or Shopify where each tenant is a store. Here's a portion of a simplified schema:
+
+.. image:: ../images/erd/mt-before.png
```
begriffs: The tool I'm using ([erd](https://github.com/BurntSushi/erd)) manages the layout. Internally it's calling graphviz which does have the ability to honor graph layout preferences. In the future we could open an issue in erd to allow layout specification or I could use graphviz's `dot` tool directly. Or another relation diagram tool if you know a good one.

Also I tried appending " (pk)" and " (fk)" to lines but the program couldn't parse those characters. I'll add a legend to the schema diagrams that explains the font conventions.

begriffs:
A note about our diagram tool, layout, and font conventions. The tool I'm using ([erd](https://github.com/BurntSushi/erd)) manages the layout. Internally it's calling graphviz which does have the ability to honor graph layout preferences. In the future we could open an issue in erd to accept layout specification and pass them along, or I could use graphviz's `dot` tool directly. Or another relation diagram tool if you know a good one.

Also I tried appending " (pk)" and " (fk)" to lines but the program couldn't parse those characters. I'll add a legend to the schema diagrams that explains the font conventions.

migration/transitioning.rst
begriffs: IIRC we discussed both approaches but decided to document just one of them. We mentioned that the other approach of joining tables and filtering one of them is more elegant but there are times when it breaks down such as when applied to a CTE.

migration/transitioning.rst
```diff
@@ -0,0 +1,92 @@
+.. _transitioning_mt:
+
+Migrating an existing single-node PostgreSQL database to Citus sometimes requires adjusting the schema and queries for optimal performance. Citus extends PostgreSQL with distributed functionality, but it is not a drop-in replacement that scales out all workloads. A performant Citus cluster involves thinking about the data model, tooling, and choice of SQL features used.
```
sumedhpathak: As noted, we aren't making a specific recommendation here around migration. So we can just say something like "Migrating an existing solution to Citus" or "Migrating an existing relational store to Citus".

migration/transitioning.rst
```diff
@@ -0,0 +1,92 @@
...
+
+Citus is well suited to hosting B2B multi-tenant application data. In this model application tenants share a Citus cluster and a schema. Each tenant's table data is stored in a shard determined by a configurable tenant id column. Citus pushes queries down to run directly on the relevant tenant shard in the cluster, spreading out the computation. Once queries are routed this way they can be executed without concern for the rest of the cluster. These queries can use the full features of SQL, including joins and transactions, without running into the inherent limitations of a distributed system.
+
+This section will explore how to model for the multi-tenant scenario, including necessary adjustments to the schema and queries.
+
+The general process is
+
+1. Recreate the schema of the original database on the Citus coordinator node.
+2. Modify the schema for multi-tenant distribution.
+3. Copy the data from the original database to Citus.
```
sumedhpathak: As discussed, I think we can remove this section. This PR isn't a step-by-step guide yet.

migration/transitioning.rst
sumedhpathak: As discussed, we can leave it as a SELECT, but maybe note that we are joining with `orders` to get the store_id, and then the output of this can be copied to a file, followed by ingest into Citus.

migration/transitioning.rst
sumedhpathak: As discussed, let's capture the second option as well.

migration/transitioning.rst
```diff
@@ -0,0 +1,98 @@
...
+   :alt: Schema before migration
+
+   (Underlined items are primary keys, italicized items are foreign keys.)
+
+In our example each store is a natural tenant. This is because storefronts benefit from dedicated processing power for their customer data, and stores do not need to access each other's sales or inventory. The tenant id is in this case the store id. We want to distribute data in the cluster in such a way that rows from the above tables in our schema reside on the same node whenever the rows share a store id.
+
+The first step is preparing the tables for distribution. Citus requires that primary keys contain the distribution column, so we must modify the primary keys of these tables and make them compound including a store id. Making primary keys compound will require modifying the corresponding foreign keys as well.
+
+In our example the stores and products tables are already in perfect shape. The orders table needs slight modification: updating the primary and foreign keys to include store_id. The line_items table needs the biggest change. Being normalized, it lacks a store id. We must add that column, and include it in the primary key constraint.
```
sumedhpathak: The last discussed optional item here was to add the explicit `ALTER TABLE` statements to add the store_id and modify the PK and FK's. This will make the before/after pics more readable and understandable.

Issue #107, opened 2016-06-28
Create introduction to Distributed Tables section


begriffs:
It currently begins by talking about advanced features and could use some background info. It should talk about the types tables in Citus, from reference tables to distributed tables, and contrasting hash and append distribution.

begriffs:
Done.

PR #173, opened 2016-09-23
Add section about sharding strategy

### LINKED ISSUE #145: Add "Distributed Data Modeling" section after Installation section
- Choosing a sharding strategy
- Use cases
- Tradeoffs
- Any info about Append distribution should link to special section

begriffs:
This moves content that used to be in "Distributed Tables" into its own section and adds a discussion of Citus use cases and tradeoffs.

Fixes #145 

Note that once we merge the migration PR we can link to it from the multi-tenant tradeoffs section in this PR.


sharding/data_modeling.rst
```diff
@@ -1,7 +1,7 @@
-.. _distributed_tables:
+.. _distributed_data_modeling:
 
 Picking a Distribution Column
-#############################
+-----------------------------
 
 Every distributed table in Citus has exactly one column which is chosen as the distribution column. This informs the database to maintain statistics about the distribution column in each shard. Citus’s distributed query optimizer then leverages these statistics to determine how best a query should be executed.
```
sumedhpathak: This paragraph is OK to remain, but the below two seem like they should be maybe after the use-case discussion?

sharding/data_modeling.rst
```diff
@@ -10,6 +10,31 @@ Typically, you should choose that column as the distribution column which is the
 .. note::
   Citus also allows joining on non-distribution key columns by dynamically repartitioning the tables for the query. Still, joins on non-distribution keys require shuffling data across the cluster and therefore aren’t as efficient as joins on distribution keys.
 
-The best choice of distribution column varies depending on the use case and queries. Two common scenarios are the multi-tenant B2B application and the realtime analytics dashboard. In the former all tables include a tenant id and are distributed by it. When SQL queries are restricted to accessing data about a single tenant then Citus can execute them within a single shard. Having all data colocated in a shard is efficient and supports all SQL features. However running queries on a single shard limits the ability to parallelize execution.
+Use Cases and Tradeoffs
```
sumedhpathak: Maybe remove this top-level section, and just dive into a discussion of the two separate data models?

sharding/data_modeling.rst
```diff
@@ -10,6 +10,31 @@ Typically, you should choose that column as the distribution column which is the
...
+
+Multi-Tenant
+~~~~~~~~~~~~
+
+* ✔ SQL Coverage
+* ✔ Ease of Migration
+* ✘ Less Parallelism
+* ✘ Less effective for large (or few) tenants
+
+In the multi-tenant use case all tables include a tenant id and are distributed by it. When SQL queries are restricted to accessing data about a single tenant then Citus can execute them within a single shard. Having all data colocated in a shard is efficient and supports all SQL features. However running queries on a single shard limits the ability to parallelize execution.
```
sumedhpathak: So I think this content is good but should probably be expanded a bit. We want to establish what multi-tenancy is, and how it would work in Citus. So basically, what does 'tenant id' mean? That may not be immediately obvious. We should also discuss the concept of co-location, and thus its necessary that *every* table is partitioned by the same tenant-id column. The blog post has some good content we can use: https://www.citusdata.com/blog/2016/08/10/sharding-for-a-multi-tenant-app-with-postgres/. Providing example use-cases like the blog-post does would help in allowing a user to identify.
It probably also makes sense to link to that post.

sharding/data_modeling.rst
```diff
@@ -10,6 +10,31 @@ Typically, you should choose that column as the distribution column which is the
...
+Use Cases and Tradeoffs
+-----------------------
 
-The other common option, realtime analytics, distributes by another column (such as user id). The queries in this scenario typically request aggregates information from multiple shards. This permits query parallelism but restricts some of the SQL features available, due to the constraints of being a distributed system.
+The best choice of distribution column varies depending on the use case and queries. Two common scenarios are the multi-tenant B2B application and the realtime analytics dashboard. Both cases involve tradeoffs.
+
+Multi-Tenant
+~~~~~~~~~~~~
+
+* ✔ SQL Coverage
```
sumedhpathak: I like the simplicity of these, but it may not be immediately obvious what these points are (e.g. why SQL coverage is higher). I'd be OK to move these to a more text/paragraph format, and discuss these *after* the below discussion on multi-tenancy.

sharding/data_modeling.rst
```diff
@@ -10,6 +10,31 @@ Typically, you should choose that column as the distribution column which is the
...
+* ✘ Less effective for large (or few) tenants
+
+In the multi-tenant use case all tables include a tenant id and are distributed by it. When SQL queries are restricted to accessing data about a single tenant then Citus can execute them within a single shard. Having all data colocated in a shard is efficient and supports all SQL features. However running queries on a single shard limits the ability to parallelize execution.
+
+Existing schemas and queries typically require little adjustment during migration to the multi-tenant architecture. Additionally multi-tenant apps with few (or very large) tenants will not see a significant performance improvement.
+
+Real-Time Analytics
+~~~~~~~~~~~~~~~~~~~
+
+* ✔ Parallelism
```
sumedhpathak: Same feedback here as above. Let's make these in a more paragraph format with a bit more discussion, and move them after we discuss what we mean by real-time analytics.

sharding/data_modeling.rst
```diff
@@ -10,6 +10,31 @@ Typically, you should choose that column as the distribution column which is the
...
+* ✔ Linear Scalability
+* ✘ Reduced SQL Coverage
+* ✘ Upfront Data Modeling
+
+The other common option, realtime analytics, distributes by another column (such as user id). The queries in this scenario typically request aggregate information from multiple shards. This permits query parallelism but restricts some of the SQL features available, due to the constraints of being a distributed system.
```
sumedhpathak: Let's expand this content as well. Let's call out a use-case (e.g. real-time dash-boards), which then lays context for why we need parallelism and performance. We can point to our technical solution as an example. We can also add our existing paragraphs here. (The ones about statistics being gathered, join-performance and note on data shuffling).

sumedhpathak: Instead of 'another column', can we note something like shard by a 'lower-level entity than a tenant, for example a user-id'. 

I got it from our [blog-post](https://www.citusdata.com/blog/2016/08/10/sharding-for-a-multi-tenant-app-with-postgres/):
"It’s important to note as we did in our previous post that different distribution models can have different benefits and trade-offs. In some cases modeling on a lower level entity id such as products or purchases can be the right choice–you gain more parallelism for analytics and trade off simplicity in querying a single store. Either choice of picking a multi-tenant data model or adopt a more distributed document model can be made to scale, but each comes with its own trade-offs. If you have the need today to scale out your multi-tenant app then give Citus Cloud a try or if you have any questions on which might work best for your situation please don’t hesitate to reach out to us. We can help."

begriffs:
OK, this section now provides a higher level overview of our use cases and their implications for data modeling. Let me know if you think we need concrete examples. Also there's a broken link to the data migration section will I can resolve once the PR for that section is merged.

begriffs:
Fixes #107 as well

ozgune:
hey @begriffs -- I read over the changes today. I had a few notes/questions that I wanted to share.

1. We previously emphasized real-time analytics and data modeling for that use case. Now that we have two separate use-cases, one geared towards multi-tenancy and another towards real-time analytics, writing a good ``Choosing a distribution column`` section becomes harder. What if we started with an initial heuristic to pick a use-case and then dived deeper into data modeling based on that? -- don't know
    * If you have tenants in your data and you're looking to migrate into Citus, choose distribution by tenant_id. Else if you have real-time analytics or better high throughput writes/look-ups, choose entity_id.

2. Since this is the data modeling section, what do you think of emphasizing the distribution column in the subsections? For example, instead of "Multi-tenant Applications", what if we named these sections "Shard by tenant_id" / "Distribute by Tenant_Id" or "Distribute by Entity_Id"?
    * Do we also want to give specific examples in these data modeling sections?

3. Could we describes the benefits to "Multi-tenant Applications" more? When I read this section, it feels a bit underwhelming.

4. (small) Does this pull request close #107? We talk about hash distribution in this section. We don't however talk about a few related concepts: (a) reference tables, (b) coordinator tables (the tables you create on the coordinate node and can't join with other distributed tables), and (c) append distribution.

begriffs:
Open question: what should we call the RT examples? They are currently named inconsistently. Some have domain-specific names and others are the names of generic schema constructions.

* Raw Events Table
* Events and Summaries
* Updatable Large Table
* Behavioral Analytics
* Star Schema

sharding/data_modeling.rst
```diff
@@ -0,0 +1,222 @@
+.. _distributed_data_modeling:
+
+Distributed modeling refers to choosing how to distribute information across nodes in a multi-machine database cluster and query it efficiently. There are common use cases for a distributed database with well understood design tradeoffs. It will be helpful for you to identify whether your application falls into one of these categories in order to know what features and performance to expect.
+
+Citus uses a column in each table to determine how to allocate its rows among the available shards. In particular, as data is loaded into the table, Citus uses this so-called *distribution column* as a hash key to allocate each incoming row to a shard.
+
+The database administrator, not Citus, picks the distribution column of each table. Thus the main task in distributed modeling is choosing the best division of tables and their distribution columns to fit the queries required by an application.
```
begriffs: I added this flourish because of a question I heard in a tech-fit. The person wondered if they could pick the column or if it was out of their control. Samay assured me this is a common question. However "The database administrator picks..." does answer the question more succinctly.

sharding/data_modeling.rst
```diff
@@ -0,0 +1,190 @@
...
+  ORDER BY minute ASC;
+
+Events and Summaries
+--------------------
+
+The previous example calculates statistics at runtime, doing possible recalculation between queries. Another approach is precalculating aggregates. This avoids recalculating raw event data and results in even faster queries. For example, a web analytics dashboard might want a count of views per page per day. The raw events data table looks like this:
+
+.. code-block:: postgres
+
+  CREATE TABLE page_views (
```
begriffs: Yeah I got it from https://www.citusdata.com/blog/2016/11/29/event-aggregation-at-scale-with-postgresql/ - which distributes by tenant_id, but it's simpler to use page_id here so I'll make that change. Looks like I used page_id when I adapted the blog post content in the issue for this PR (#145) but tenant_id slipped through the second time when I copied directly from the blog post.

sharding/data_modeling.rst
```diff
@@ -0,0 +1,222 @@
...
+
+The first step is identifying what constitutes a tenant in your app. Common instances include company, account, organization, or customer. The column name will thus be something like :code:`company_id` or :code:`customer_id:`. Examine each of your queries and ask yourself: would it work if it had additional WHERE clauses to restrict all tables involved to rows with the same tenant id? You can visualize these clauses as executing queries *within* a context, such restricting queries on sales or inventory to be within a certain store.
+
+If you're migrating an existing database to the Citus multi-tenant architecture then some of your tables may lack a column for the application-specific tenant id. You will need to add one and fill it with the correct values. This will denormalize your tables slightly, but it's because the multi-tenant model mixes the characteristics of hierarchical and relational data models. For more details and a concrete example of backfilling the tenant id, see our guide to `Transitioning to Citus`_.
+
```
begriffs: The Transitioning section has before-and-after entity relation diagrams. What do you have in mind for this section? Something like a line_items table with its original columns and a new `store_id` column outlined with a dashed border? The image would basically show, "you have to add this new column." Or are you thinking having boxes for tenant 1, 2, 3 and each tenant box would contain the same tables? That would be to show that tenants share a schema.

sharding/data_modeling.rst
```diff
@@ -0,0 +1,222 @@
...
+  GROUP BY bot;
+
+Star Schema
+-----------
+
+We've already seen how every row in a distributed table is stored on a shard. However for small tables there is a trick to achieve a kind of universal colocation. We can choose to place all its rows into a single shard but replicate that shard to every worker node. It introduces storage and update costs of course, but this can be more than counterbalanced by the performance gains of read queries.
+
+We call tables replicated to all nodes *reference tables.* They usually provide metadata about items in a larger table and are reminiscent of what data warehousing calls dimension tables. For example, suppose we have a large table of phone calls:
+
+.. code-block:: postgres
```
samay-sharma: @ozgune : I agree. Once we have complete reference table implementation in v6.1 (and get some input on its applicability), we can re-introduce this example.

I would however, still prefer to keep the text sections just to let people know that they can do something like this with Citus.

sharding/data_modeling.rst
```diff
@@ -0,0 +1,234 @@
...
+
+If you're migrating an existing database to the Citus multi-tenant architecture then some of your tables may lack a column for the application-specific tenant id. You will need to add one and fill it with the correct values. This will denormalize your tables slightly. For more details and a concrete example of backfilling the tenant id, see our guide to `Transitioning to Citus`_.
+
+Distributing by Entity ID
+=========================
+
+While the multi-tenant architecture introduces a hierarchical structure and uses data co-location to parallelize queries between tenants, real-time architectures depend on specific distribution properties of their data to achieve highly parallel processing. We use "entity id" as a term for distribution columns in the real-time model, as opposed to tenant ids in the multi-tenant model. Typical entites are users, devices, or events.
```
ozgune: Minor question. users, devices, or events -> users, hosts, or devices?

We don't have any production customers who distribute at the granularity of an event. That's why I wanted to ask.

sharding/data_modeling.rst
```diff
@@ -0,0 +1,234 @@
...
+
+Real-time queries typically ask for numeric aggregates grouped by date or category. Citus sends these queries to each shard for partial results and assembles the final answer on the coordinator node. Hence queries run fastest when as many nodes contribute as possible, and when no individual node bottlenecks.
+
+Hence the more evenly a choice of entity id distributes data to shards the better. At the least the column should have a high cardinality. For comparison, a binary gender field is a poor choice because it assumes at most two values. These values will not be able to take advantage of a cluster with many shards. The row placement will skew into only two shards:
```
ozgune: Minor. We had two "hences" after each other. What if we made a minor revision?

sharding/data_modeling.rst
```diff
@@ -0,0 +1,234 @@
...
+  FROM page_views
+  WHERE view_time >= $1
+  GROUP BY view_time::date, page_id
+  ON CONFLICT (day, page_id) DO UPDATE SET
+    view_count = daily_page_views.view_count + EXCLUDED.view_count;
+
+Updatable Large Table
+---------------------
+
+(Device table that has characteristics that get updated. Sharded by device id.)
```
ozgune: Do we want to fill in this section?

sharding/data_modeling.rst
```diff
@@ -0,0 +1,234 @@
...
+
+  CREATE TABLE daily_page_views (
+    day date,
+    page_id int,
+    view_count bigint,
+    PRIMARY KEY (day, page_id)
+  );
+
+  SELECT create_distributed_table('daily_page_views', 'page_id',
+                                  colocate_with => 'page_views');
```
ozgune: For the examples in this page, we don't need to show the ``colocate_with`` option.

For simplicity, could we remove ``colocate_with`` from this section and the next?

ozgune:
@begriffs -- I added minor comments to this pull request. I'm very excited about these changes.

We're missing a few sections (Updateable Large Tables and Modeling Concepts). I also have a few higher level questions. Since this section is a major improvement over what we had, I'm inclined to address the current feedback and check this PR in.

We'd then open a second pull request that could have any of the following (to be scoped together):

* We now have users coming to Citus for the NoSQL use case. Do we want to mention that in our data modeling section?
* *Do we want to add an example table schema and/or a diagram to the data modeling section?* We could also demonstrate foreign keys here.
* Do we want to have two examples for the multi-tenant data model? One for OLTP ([example blog](https://www.citusdata.com/blog/2016/10/03/designing-your-saas-database-for-high-scalability/)) and another for OLAP ([example blog](https://www.citusdata.com/blog/2016/12/22/scaling_out_sql_with-colocation/))?
* We haven't finalized section names (Raw Events Table, etc.). What do we name these tables?
* Both in the multi-tenant and real-time sections, we're in fact using the hierarchical data model.
* We're missing two sections from our outline -- Updateable Large Table and Modeling Concepts. Should we fill these sections in?

begriffs:
Final changes and then shipit:

[ ]  Remove example code from the events with reference tables section. (but save it for later when our sql coverage has improved)

[ ] switch to User ID as entity in "Events with Roll-Ups"

[ ] find a more literal way to say "unknown unknowns" and without the quote marks

[ ] change name from "Distributed modeling" to data modeling, or even distributed data modeling if it fits

[ ] Remove unused colocation image

PR #175, opened 2016-10-11
Fix outbound links in presence of adblocker

### LINKED ISSUE #63: All links fail without Google Analytics
This technique aborts link clicks with the `return false` when a browser privacy extension blocks access to GA. This may not affect very many people, but wanted to point it out.

```html
<a
  href="https://packages.citusdata.com/tutorials/citus-tutorial-osx-1.1.0.tar.gz" 
  onclick="trackOutboundLink('https://packages.citusdata.com/tutorials/citus-tutorial-osx-1.1.0.tar.gz'); return false;"
>Download</a>
```

/cc @craigkerstiens

begriffs:
If Google Analytics callback hasn't occurred one second after clicking a link, go ahead and run the callback ourselves because an adblocker has thwarted it. This techniques comes from the Google Analytics section about the [hitCallback](https://developers.google.com/analytics/devguides/collection/analyticsjs/sending-hits#handling_timeouts).

The particular timeout could perhaps be shortened because it feels like a long time for the link to activate. The danger with shortening it is interrupting analytics if GA is laggy.

Fixes #63

### DIFF
--- a/_themes/citus/layout.html
+++ b/_themes/citus/layout.html
@@ -220,7 +220,7 @@ var createFunctionWithTimeout = function(callback, opt_timeout) {
 var trackOutboundLink = function(url) {
   ga('send', 'event', 'outbound', 'click', url, {
     'transport': 'beacon',
-    'hitCallback': createFunctionWithTimeout(function(){document.location = url;})
+    'hitCallback': createFunctionWithTimeout(function(){document.location = url;}, 500)
   });
 }
 </script>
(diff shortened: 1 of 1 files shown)


lfittl:
@begriffs This looks correct from a Javascript side of things.

begriffs:
Think I should cut the timeout down to half a second?

lfittl:
@begriffs Good question - can you test this locally with adblocker enabled and see what makes sense?

Given that GA uses 1 second in their example, I'd be inclined to keep that, unless we see it being noticeably slow.

PR #207, opened 2016-11-14
Remove documentation where we say Insert/Select isnt supported

### LINKED ISSUE #196: Remove or change INSERT / SELECT workaround in docs
Now that we have INSERT / SELECT for colocated tables in v6.0, we should probably remove that section from our faq and sql workarounds. If we want to still have a section, then we should clarify that it now applies only to non-colocated tables.

sumedhpathak:
Remove notes where we say Insert-Into-Select isn't supported, as we support it 6.0 onwards. 

Fixes #196

### DIFF
--- a/faq/faq.rst
+++ b/faq/faq.rst
@@ -34,14 +34,12 @@ As the Citus master node is similar to a standard PostgreSQL server, regular Pos
 How do I ingest the results of a query into a distributed table?
 ----------------------------------------------------------------
 
-Citus currently doesn't support the INSERT / SELECT syntax for copying the results of a query into a distributed table. As a simple workaround, you could use
+Citus now supports the INSERT / SELECT syntax for copying the results of a query
+on a distributed table into a distributed table, when the tables are colocated.
 
-::
-
-  psql -c "COPY (query) TO STDOUT" | \
-  psql -c "COPY table FROM STDIN"
-
-In addition to the above, there are other workarounds which are more complex to implement but are more efficient. Please contact us if your use-case demands such ingest workflows.
+If your tables are not colocated, or you are using append distribution, there
+are workarounds which are more complex to implement. Please contact us if your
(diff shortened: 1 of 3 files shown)


faq/faq.rst
```diff
@@ -34,14 +34,12 @@ As the Citus master node is similar to a standard PostgreSQL server, regular Pos
 How do I ingest the results of a query into a distributed table?
 ----------------------------------------------------------------
 
-Citus currently doesn't support the INSERT / SELECT syntax for copying the results of a query into a distributed table. As a simple workaround, you could use
+Citus now supports the INSERT / SELECT syntax for copying the results of a query
```
samay-sharma: Maybe remove 'now'?

faq/faq.rst
```diff
@@ -34,14 +34,12 @@ As the Citus master node is similar to a standard PostgreSQL server, regular Pos
...
+on a distributed table into a distributed table, when the tables are colocated.
 
-::
-
-  psql -c "COPY (query) TO STDOUT" | \
-  psql -c "COPY table FROM STDIN"
-
-In addition to the above, there are other workarounds which are more complex to implement but are more efficient. Please contact us if your use-case demands such ingest workflows.
+If your tables are not colocated, or you are using append distribution, there
+are workarounds which are more complex to implement. Please contact us if your
```
samay-sharma: Maybe say `there are other workarounds which you can use (for eg. using COPY to copy data out and then back into the destination table)` rather than complex to implement.

faq/faq.rst
samay-sharma: Link to https://www.postgresql.org/docs/9.6/static/sql-insert.html on INSERT / SELECT

sumedhpathak: Done

sumedhpathak: Done

PR #228, opened 2016-12-05
Add Multi-tenant info in SQL Workarounds

### LINKED ISSUE #197: Incorporate SQL support for multi-tenant in the introduction to workarounds section
Currently, the introduction in the [workarounds page](https://github.com/citusdata/citus_docs/blob/master/reference/sql_workarounds.rst) says,

"Before attempting workarounds consider whether Citus is appropriate for your situation. Citus' current version works well for real-time analytics use cases. We are continuously working to increase SQL coverage to better support other use-cases such as data warehousing queries.

Citus supports most, but not all, SQL statements directly. Its SQL support continues to improve. Also many of the unsupported features have workarounds; below are a number of the most useful."

We should probably put some information about multi-tenant in this introduction and also make some changes like "Citus supports all SQL statements directly if you are sharding by tenant id. For queries which need to span across nodes, Citus may not support all SQL statements directly".

sumedhpathak:
SQL Workarounds now mentions full sql support in multi-tenant use-cases.

Fixes #197

### DIFF
--- a/reference/sql_workarounds.rst
+++ b/reference/sql_workarounds.rst
@@ -3,9 +3,15 @@
 SQL Workarounds
 ===============
 
-Before attempting workarounds consider whether Citus is appropriate for your situation. Citus' current version works well for real-time analytics use cases. We are continuously working to increase SQL coverage to better support other use-cases such as data warehousing queries.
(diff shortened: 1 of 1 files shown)


begriffs:
👍

PR #233, opened 2016-12-07
Master_add_node also accepts IP addresses

### LINKED ISSUE #214: Document master_add_node also accepts IP addresses apart from DNS names
This question has come up several times now in customer conversations. We used to only denote DNS names, but also accept IP addresses.

begriffs:
Fixes #214

### DIFF
--- a/admin_guide/cluster_management.rst
+++ b/admin_guide/cluster_management.rst
@@ -22,10 +22,7 @@ Adding a worker
 
 Citus stores all the data for distributed tables on the worker nodes. Hence, if you want to scale out your cluster by adding more computing power, you can do so by adding a worker.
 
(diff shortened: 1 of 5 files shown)


sumedhpathak:
Some other potential sections to changes:

https://docs.citusdata.com/en/v6.0/faq/faq.html?highlight=master_add_node#how-do-i-add-nodes-to-an-existing-citus-cluster

https://docs.citusdata.com/en/v6.0/installation/production_deb.html?highlight=master_add_node#steps-to-be-executed-on-the-master-node (we note this line: "Add the workers’ DNS names and server ports to the table.")

https://docs.citusdata.com/en/v6.0/installation/production_rhel.html?highlight=master_add_node

PR #237, opened 2016-12-14
Document insert into select

### LINKED ISSUE #211: Document insert into select
In addition to using in the tutorials we should have some high level reference on how it works, it sounds like based on changelog we require it to be co-located, but provide no even simple examples.

_______

### Example sql

See blog post

### Corner cases, gotchas

* Cannot mix stable and volatile functions for values to be inserted. ("Consider what happens with replication-factor > 1. If you run NOW() on the workers, the two replicas now differ.")
   ```
   DETAIL:  Stable and volatile functions are not allowed in INSERT ... SELECT queries
   LOCATION:  ErrorIfInsertSelectQueryNotSupported, multi_router_planner.c:643
   ```
  * One instance of this is `timestamp` being used in immutable functions and `timestamptz` in stable functions
* Cannot EXPLAIN these queries (but no need to document it really)
   ```
   DETAIL:  Select query cannot be pushed down to the worker.
   LOCATION:  RouterModifyTaskForShardInterval, multi_router_planner.c:393
   ```
* Citus takes locks to ensure consistency when replication-factor>1. If you  know your inserts are not going to affect an ongoing rollup, you can set `SET LOCAL citus.all_modifications_commutative TO on;` to avoid stricter locking (get better throughput)

* https://www.citusdata.com/blog/2016/11/29/event-aggregation-at-scale-with-postgresql/
* Also known as the "precomputed view" pattern http://igoro.com/archive/precomputed-view-a-cool-and-useful-sql-pattern/
* https://github.com/citusdata/citus/issues/508

### Link to regression tests and diffs if applicable

* informative commit: https://github.com/citusdata/citus/commit/1673ea937c08e13ebf4fc192da037b615b5e519d
* regression test: https://github.com/citusdata/citus/blob/master/src/test/regress/sql/multi_insert_select.sql

To support scaling data pipelines. In particular pre-aggregating data avoids the cost of processing raw data at runtime.

* ETL (but not our intended use)
* Reducing space (over time replacing raw data by the aggregates)
  * The very old data can be aggregated, where newer data can be retained in its full details
* For realtime use case
  * Predefined aggregates that will be rolled up
  * Pre-computing aggregate functions on time intervals, such as daily page view counts in a web analytics dashboard.
* Multi-tenant use case
  * Rollups can improve queries constrained to a single core

## Communication goals (e.g. detailed howto vs advertising capability)

* Explain what roll ups are and how they would help
* Talk about the syntax
* Examples
* How it works, running on each shard, how colocation is important
* Gotchas

Provide example of pre-computing aggregates, both creating initial value and using upsert to update them periodically (aka incremental refresh). If we do not identify another use case then the new docs section might better be called "Pre-Computing Aggregates" rather than "insert into ... select".

In what way should this differ from the blog post? In other words, can we copy the post entirely as a docs article, adjusting statements that mention particular performance numbers (such as, "10 million events per second and generating 1.7 million aggregates")?

* As a new section inside "Ingesting, Modifying Data (DML)"
  * Title it in terms of insert into select rather than a use case so that it blends better with the other points in the section

begriffs:
Fixes #211

Ready for review. I have tested the code snippets on Cloud.


dist_tables/dml.rst
```diff
@@ -38,6 +38,70 @@ For example:
...
+
+It is important to be able to update rollups *incrementally*, that is, without re-scanning all data. Both the rolling up and updating can be accomplished with :code:`INSERT INTO ... SELECT`, whereby the results of a select query are inserted into a table. A full reference for this feature is available in the PostgreSQL `docs <https://www.postgresql.org/docs/current/static/sql-insert.html>`_. It works well on Citus because it is parallelizable under certain conditions. We'll see more about that later.
+
+First some basics. Insert into select uses the results of a select query to populate a table. For instance consider a simple select statement which counts the number of web page views per page per day.
```
ozgune: Small. We use "insert into select" several times in this section.

Is it worth standardizing how we write it down? Insert Into Select, ``INSERT INTO SELECT``, etc.?

dist_tables/dml.rst
```diff
@@ -38,6 +38,70 @@ For example:
...
+    primary key (day, page_id)
+  );
+
+  INSERT INTO daily_page_views (day, page_id, view_count)
+    -- this is our original select query
+    SELECT view_time::date AS day, page_id, count(*) AS view_count
+    FROM page_views
+    GROUP BY view_time::date, page_id;
+
+We can use PostgreSQL's upsert feature to incrementally update certain parts of the rollup table. For instance suppose we have already rolled up visits happening before a certain timestamp (we'll call it :code:`$1`) and want to update the rollups to include more recent views. Just add a WHERE clause and specify "ON CONFLICT." The latter is PostgreSQL's technique for doing `upserts <https://www.postgresql.org/docs/9.5/static/sql-insert.html#SQL-ON-CONFLICT>`_.
```
ozgune: Small style questions.

* In our documentation, when we're referring to the actions we expect our customers to take, do we say "we" or "you"? For example, "You can use PostgreSQL's upsert feature ..."
* How formal / informal do we want to be in our documentation? For example, "Just add a WHERE clause and specify ON CONFLICT" feels a bit informal to me -- don't know, just asking

dist_tables/dml.rst
```diff
@@ -38,6 +38,70 @@ For example:
...
+- GROUP BY includes the distribution column
+
+The example query shown above will work well if :code:`page_views` and :code:`daily_page_views` are distributed by :code:`page_id` column. Distributing our data by page allows rollup to scale linearly with the number of cores.
+
+.. code-block:: postgres
+
+  SELECT create_distributed_table('page_views', 'page_id');
+  SELECT create_distributed_table('daily_page_views', 'page_id');
+
+There are a few things to be aware of when running distributed insert into selects. First, Citus takes locks to ensure consistency when the :ref:`replication_factor` is greater than one. In this situation if you know your inserts are not going to affect an ongoing rollup, you can set :code:`SET LOCAL citus.all_modifications_commutative TO on;`. This avoids strict locking and provides better throughput.
```
ozgune: Minor question. This isn't an issue on Cloud deployments. We're also planning to reduce our default replication factor to 1.

If we add this section, does it confuse the reader more?

dist_tables/dml.rst
```diff
@@ -38,6 +38,70 @@ For example:
...
+
+Next bear in mind that any functions in the select statement for insertion must be immutable. Stable functions are not allowed, and volatile functions are certainly not allowed. The PostgreSQL docs explain the difference between these `volatility categories <https://www.postgresql.org/docs/current/static/xfunc-volatility.html>`_. Consider what would happen with replication factor greater than one. If you run :code:`now()` on the workers, the two replicas would differ. One surprising way that stable functions can sneak in is using timestamps with timezones (:code:`timestamptz`). Functions dealing with timezones rely implicitly on the timezone configuration parameter and hence are marked stable rather than immutable.
```
ozgune: Minor note -- this could be non-actionable.

When I read this paragraph, I'm a bit confused by the first paragraph. In particular, "any functions ... immutable" requires a certain knowledge of PostgreSQL functions and their properties. If we introduced function volatility, we could ease the user into "PG functions' properties."

We may not need to think about this though. We added support for stable functions in ``INSERT INTO SELECT`` in 6.1. Once we release 6.1, one option could be to remove this paragraph (#246)?

dist_tables/dml.rst
begriffs: * My own heuristic is to use "we" when the documentation is guiding the reader through an example, because we (the author and reader) are considering it together. I use "you" less frequently, in places where older English writing would traditionally use "[one](https://en.wikipedia.org/wiki/One_(pronoun))." I could be more consistent in this section though, I'll fix it.
* "Just add a WHERE..." was just sloppy, I'll rewrite it.

dist_tables/dml.rst
begriffs: If the replication factor defaults to more than one in Citus 6.0 then we should probably include this paragraph but remove it in a later version where that is no longer the case.

dist_tables/dml.rst
begriffs: Yeah let's remove the paragraph in the 6.1 branch of the docs when it gets released.

dist_tables/dml.rst
```diff
@@ -38,6 +38,70 @@ For example:
 
     If COPY fails to open a connection for a shard placement then it behaves in the same way as INSERT, namely to mark the placement(s) as inactive unless there are no more active placements. If any other failure occurs after connecting, the transaction is rolled back and thus no metadata changes are made.
 
+INSERT INTO ... SELECT
```
begriffs: True, the "Real-time event aggregation at scale" blog post also distributes on tenant_id. I think I was just trying to simplify the example by omitting the tenant stuff and going with an entity id approach.

I can sketch an idea for a diagram in our docs sync-up and see what we think.

begriffs: (diagram added)

ozgune: @begriffs and I chatted about this and I'm summarizing our chat below.

(1) We're communicating three concepts in this section: (a) what are roll ups, (b) how do they apply in a distributed setting with table colocation, and (c) how can you do them incrementally? Could we talk about them in this order? 

I usually find that (c) is usually more advanced than other concepts. That's one of the reasons we didn't mention them in our [reference architecture](https://docs.citusdata.com/en/v6.0/tech_soln/real_time_dashboards.html)

(2) Our current ``INSERT INTO ... SELECT`` implementation assumes that the source and destination tables are colocated. Could we mention this concept earlier on?

(3) Could we standardize on a diagram that demonstrates the concept of colocation? (We could optionally add a diagram for incremental upserts after this step.)

(4) When I think about this section, I'm slightly inclined to roll-up on ``tenant_id``. This is because our data modeling section talks about ``tenant_id`` vs ``entity_id``. When we mention ``page_id``, it feels a bit low in the hierarchical data model to me.

ozgune: @begriffs and I chatted about this and I'm summarizing our chat below.

We decided to refer to another section for the colocation image. We also decided to remove the UPSERT image.

We also came up with the following outline.

1. Talk in general about data roll-up and aggregation -- This is similar to our first and second paragraphs without mention of Citus
2. Talk about data roll-up in distributed systems and segue way into Citus
3. Create raw events table ``page_views`` and distribute that table by site_id
4. Show example SELECT query and mention that this query will need to go over the entire results each time
5. Create table ``daily_page_views`` and distribute by site_id 
6. INSERT INTO SELECT distributed table and corresponding paragraphs
7. Run example SELECT query on the aggregated table
8. Other paragraphs
9. Cut last two paragraphs because we're already addressing them in 6.1

begriffs:
I reorganized the section again based on the latest conversation with Ozgun. The document seemed to flow a little later by discussing rollup more fully in a single-node scenario and moving list items 2 and 3 a little later. I'll do one last test of the commands in Citus and investigate whether the GROUP BY does actually need to contain the distribution column.

dist_tables/dml.rst
```diff
@@ -38,6 +38,105 @@ For example:
...
+.. code-block:: postgresql
+
+  CREATE TABLE daily_page_views (
+    site_id int,
+    day date,
+    url text,
+    view_count bigint,
+    PRIMARY KEY (site_id, day, url)
+  );
+
```
ozgune: <pre>
SELECT create_distributed_table('daily_page_views', 'site_id');
</pre>

ozgune: (maybe then take a break from this code snippet and explain colocation a bit)

In this example, we distributed both :code:`page_views` and :code:`daily_page_views` on the `site_id` column. This ensures that data corresponding to a particular site will get co-located (link to colocation_groups) to the same node. Keeping the two tables' rows together on each node minimizes network traffic between nodes and enables highly parallel execution.

Once we created this new distributed table, we can then run INSERT INTO SELECT to roll up raw page views into the aggregated table. In the following, we aggregate page views each day. We find that users also wait for a certain time period after the end of day, to accommodate for late arriving data.

dist_tables/dml.rst
```diff
@@ -38,6 +38,105 @@ For example:
...
+  INSERT INTO daily_page_views (day, site_id, url, view_count)
+    -- this is our original select query
+    SELECT view_time::date AS day, site_id, url,
+           count(*) AS view_count
+    FROM page_views
+    GROUP BY view_time::date, site_id, url;
```
ozgune: <pre>
INSERT INTO daily_page_views (day, site_id, url, view_count)
    -- this is our original select query, where we roll up yesterday's data
    SELECT view_time::date AS day, site_id, url,
           count(*) AS view_count
    FROM page_views
    WHERE view_time >= date '2017-01-01' AND view_time < date '2017-01-02'
    GROUP BY view_time::date, site_id, url;
</pre>

dist_tables/dml.rst
```diff
@@ -38,6 +38,105 @@ For example:
...
+
+  -- now the results are available right out of the table
+  SELECT day, site_id, url, view_count
+  FROM daily_page_views;
+
+
+Unfortunately as the numbers of tracked web sites and their pages grow, the aggregate query for populating (or as we'll see later, updating) the rollup table slows down. To compensate we can combine rollup tables and distributed computing to scale the application. Since we GROUP BY each site separately we can parallelize the computation across nodes in a distributed database. Each node in a Citus cluster can hold the data for different web sites and the nodes can each compute rollups locally and write a corresponding part of :code:`daily_page_views`. We distribute both :code:`page_views` and :code:`daily_page_views` so that their rows will stay on the same machine when their :code:`site_id` values match.
```
ozgune: Could we remove this paragraph until "In summary, ..."?

dist_tables/dml.rst
```diff
@@ -38,6 +38,105 @@ For example:
...
+  -- Then run the ordinary INSERT INTO SELECT as before
+
+  INSERT INTO daily_page_views (day, site_id, url, view_count)
+    SELECT view_time::date AS day, site_id, url,
+           count(*) AS view_count
+    FROM page_views
+    GROUP BY view_time::date, site_id, url;
+
+
+In summary, INSERT INTO SELECT on Citus requires that:
```
ozgune: (moving from two paragraphs above to here)

It's worth noting that for INSERT INTO SELECT to work on distributed tables, Citus requires the source and destination table to be colocated. In summary: ...

dist_tables/dml.rst
```diff
@@ -38,6 +38,105 @@ For example:
...
+
+- The tables queried and inserted are distributed by analogous columns
+- The select query includes the distribution column
+- The insert statement includes the distribution column
+
+Rollups keep statistics queries fast but do require upkeep. New items must be periodically added or existing entries updated. In order that this periodic update be fast we need to do it *incrementally*, meaning without having to re-scan the entire underlying dataset (as, for instance, a materialized view would require). PostgreSQL's upsert feature is what we need.
```
ozgune: @begriffs -- I edited the example query above to roll-up each day (incrementally).

Could we revise these two paragraphs to say that we could also use UPSERT to handle late arriving data?

dist_tables/dml.rst
```diff
@@ -38,10 +38,98 @@ For example:
...
+    site_id int,
+    day date,
+    url text,
+    view_count bigint,
+    PRIMARY KEY (site_id, day, url)
+  );
+
+  SELECT create_distributed_table('daily_page_views', 'site_id');
+
+In this example, we distributed both :code:`page_views` and :code:`daily_page_views` on the :code:`site_id` column. This ensures that data corresponding to a particular site will be :ref:`co-located <colocation>` on the same node. Keeping the two tables' rows together on each node minimizes network traffic between nodes and enables highly parallel execution.
```
begriffs: This link to the co-location section will activate once PR 243 is merged.

begriffs:
There is still a split in usage for "you/we" but it is consistent. The "you" sentences talk about hypothetical situations, and the "we" ones present and discuss examples.

begriffs:
@ozgune the the changes are made, ready for your OK.

dist_tables/dml.rst
```diff
@@ -38,10 +38,98 @@ For example:
 
     If COPY fails to open a connection for a shard placement then it behaves in the same way as INSERT, namely to mark the placement(s) as inactive unless there are no more active placements. If any other failure occurs after connecting, the transaction is rolled back and thus no metadata changes are made.
 
+Distributed Aggregations
+$$$$$$$$$$$$$$$$$$$$$$$$
+
+Applications like event data pipelines and real-time dashboards require fast queries across predefined aggregations of incoming data. One way to make these queries fast is by calculating and saving aggregates ahead of time. This is called "rolling up" the data and it avoids the cost of processing raw data at run-time. As an extra benefit, rolling up timeseries data into hourly or daily statistics can also save space. Old data may be deleted when its full details are no longer needed and aggregates suffice.
```
ozgune: Minor edit.

... require fast queries across predefined aggregations of incoming data. -> ... require sub-second queries on large volumes of data.

dist_tables/dml.rst
```diff
@@ -38,10 +38,98 @@ For example:
...
+    url text,
+    view_count bigint,
+    PRIMARY KEY (site_id, day, url)
+  );
+
+  SELECT create_distributed_table('daily_page_views', 'site_id');
+
+In this example, we distributed both :code:`page_views` and :code:`daily_page_views` on the :code:`site_id` column. This ensures that data corresponding to a particular site will be :ref:`co-located <colocation>` on the same node. Keeping the two tables' rows together on each node minimizes network traffic between nodes and enables highly parallel execution.
+
+Once we created this new distributed table, we can then run :code:`INSERT INTO ... SELECT` to roll up raw page views into the aggregated table. In the following, we aggregate page views each day. Citus users often wait for a certain time period after the end of day to run a query like this, to accommodate late arriving data.
```
ozgune: Minor. .. we created this new distributed -> .. we create this new distributed

dist_tables/dml.rst
```diff
@@ -38,10 +38,98 @@ For example:
...
+    WHERE view_time >= date '2017-01-01' AND view_time < date '2017-01-02'
+    GROUP BY view_time::date, site_id, url;
+
+  -- now the results are available right out of the table
+  SELECT day, site_id, url, view_count
+    FROM daily_page_views
+    WHERE site_id = 5 AND
+      day >= date '2016-01-01' AND day < date '2017-01-01';
+
+It's worth noting that for :code:`INSERT INTO ... SELECT` to work on distributed tables, Citus requires the source and destination table to be co-located. In summary:
```
ozgune: Minor question. Does the following flow better?

In summary: -> In other words:

begriffs: "In other words" in this position feels it is listing the criteria for the source and destination tables to be co-located. "In summary" feels better, more like it ties together more of the prior observations.

dist_tables/dml.rst
```diff
@@ -38,6 +38,105 @@ For example:
...
+
+  CREATE TABLE page_views (
+    site_id int,
+    url text,
+    host_ip inet,
+    view_time timestamp default now(),
+
+    PRIMARY KEY (site_id, url)
+  );
+
```
begriffs: (This has been fixed but the comment isn't disappearing because it is attached to a line in the code that hasn't changed.)

dist_tables/dml.rst
begriffs: (also fixed)

begriffs:
I verified that the code examples in this PR work with nonempty tables.

PR #253, opened 2017-01-19
Update discussions of rep factor default value

### LINKED ISSUE #219: Update docs regarding reducing default value for citus.shard_replication_factor to 1
Refer to issue in [Citus docs](https://github.com/citusdata/citus/issues/998) for more details

begriffs:
Fixes #219

Note that the Cloud > Features > Replication Model section claims rep factor 2 is most popular. I left this unchanged, believing it has to do with streaming- rather than coordinator-replication.

### DIFF
--- a/admin_guide/cluster_management.rst
+++ b/admin_guide/cluster_management.rst
@@ -54,7 +54,14 @@ In this sub-section, we discuss how you can deal with node failures without incu
 Worker Node Failures
 --------------------
 
-Citus can easily tolerate worker node failures because of its logical sharding-based architecture. While loading data, Citus allows you to specify the replication factor to provide desired availability for your data. In face of worker node failures, Citus automatically switches to these replicas to serve your queries. It also issues warnings like below on the master so that users can take note of node failures and take actions accordingly.
+Citus can easily tolerate worker node failures because of its logical sharding-based architecture. While loading data, Citus allows you to specify the replication factor to provide desired availability for your data. The replication factor has a value of one by default, but can be set to two or higher for fault tolerance.
+
+.. code-block:: postgresql
+
+  SET citus.shard_replication_factor = 2;
+
+In face of worker node failures, Citus automatically switches to these replicas to serve your queries.
+It also issues warnings like below on the master so that users can take note of node failures and take actions accordingly.
 
 ::
 
@@ -74,6 +81,7 @@ make this simpler, Citus enterprise provides a replicate_table_shards UDF which
 can be called after. This function copies the shards of a table across the
 healthy nodes so they all reach the configured replication factor.
 
+
 To remove a permanently failed node from the list of workers, you should first
 mark all shard placements on that node as invalid (if they are not already so)
 using the following query:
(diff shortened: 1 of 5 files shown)


sumedhpathak:
In Cluster Management: https://docs.citusdata.com/en/v6.0/admin_guide/cluster_management.html#worker-node-failures, Could we also add a *NOTE* or something like that to denote that our current default replication factor is 1? We should also add that if users want fault tolerance, they should set this value to 2 or higher?

admin_guide/cluster_management.rst
```diff
@@ -74,6 +81,7 @@ make this simpler, Citus enterprise provides a replicate_table_shards UDF which
 can be called after. This function copies the shards of a table across the
 healthy nodes so they all reach the configured replication factor.
 
+
```
sumedhpathak: Why the newline change?

ref_arch/real_time_dashboards.rst
```diff
@@ -30,8 +30,7 @@ all the details in one place. If you've followed our installation instructions f
 Citus on either a single or multiple machines you're ready to try it out.
 
 Data Model
-----------
-
```
sumedhpathak: What is the change here? Visually I see no difference between this change and current docs?
https://docs.citusdata.com/en/v6.0/tech_soln/real_time_dashboards.html#data-model
https://citus-doc-preview-pr-253.herokuapp.com/ref_arch/real_time_dashboards.html#data-model

reference/append.rst
```diff
@@ -157,7 +157,7 @@ Citus assigns a unique shard id to each new shard and all its replicas have the
 By default, the \\copy command depends on two configuration parameters for its behavior. These are called citus.shard_max_size and citus.shard_replication_factor.
 
 (1) **citus.shard_max_size :-** This parameter determines the maximum size of a shard created using \\copy, and defaults to 1 GB. If the file is larger than this parameter, \\copy will break it up into multiple shards.
-(2) **citus.shard_replication_factor :-** This parameter determines the number of nodes each shard gets replicated to, and defaults to two. The ideal value for this parameter depends on the size of the cluster and rate of node failure. For example, you may want to increase the replication factor if you run large clusters and observe node failures on a more frequent basis.
+(2) **citus.shard_replication_factor :-** This parameter determines the number of nodes each shard gets replicated to, and defaults to one. The ideal value for this parameter depends on the size of the cluster and rate of node failure. For example, you may want to increase the replication factor if you run large clusters and observe node failures on a more frequent basis.
```
sumedhpathak: >For example, you may want to increase the replication factor if you run large clusters and observe node failures on a more frequent basis.

Should we modify this statement to say something like "Set it to 2 if you want Citus to replicate data automatically and provide fault tolerance. You may want to increase the factor even higher if you run large clusters ... "?

begriffs:
I tested the tutorial and real-time dashboards sections with rep factor 1 and they work. (Tested the dashboards section up to HLL actually, and had to modify a script that calls rollup_1min for pairs of shards because my psql outputs fancy ansi decorations that messes up the xargs)

sumedhpathak:
Looks good! 🚢

PR #258, opened 2017-01-30
Fix a small typo in the Data Modeling section

aamederen:


begriffs:
Good catch, that was a weird one.

PR #270, opened 2017-01-31
Tenant isolation

begriffs:
One question. When I talk about the benefits for noisy neighbors etc, those benefits will be realized only after calling `master_move_shard_placement` to get the isolated shard onto its own node. Should I add details about that other UDF here as well?

### DIFF
--- a/admin_guide/cluster_management.rst
+++ b/admin_guide/cluster_management.rst
@@ -87,7 +87,7 @@ Tenant Isolation
 
 .. note::
 
-  Tenant isolation is a feature of **Citus Enterprise Edition** and :ref:`Citus Cloud <cloud>` only.
+  Tenant isolation is a feature of **Citus Enterprise Edition** and :ref:`Citus Cloud <cloud_overview>` only.
 
 Citus places table rows into worker shards based on the hashed value of the rows' distribution column. Multiple distribution column values often fall into the same shard. In the Citus multi-tenant use case this means that tenants often share shards.
 
--- a/admin_guide/cluster_management.rst
+++ b/admin_guide/cluster_management.rst
@@ -87,7 +87,7 @@ Tenant Isolation
 
 .. note::
 
-  Tenant isolation is a feature of **Citus Enterprise Edition** only.
+  Tenant isolation is a feature of **Citus Enterprise Edition** and :ref:`Citus Cloud <cloud>` only.
 
 Citus places table rows into worker shards based on the hashed value of the rows' distribution column. Multiple distribution column values often fall into the same shard. In the Citus multi-tenant use case this means that tenants often share shards.
 
--- a/admin_guide/cluster_management.rst
+++ b/admin_guide/cluster_management.rst
@@ -91,7 +91,9 @@ Tenant Isolation
 
 Citus places table rows into worker shards based on the hashed value of the rows' distribution column. Multiple distribution column values often fall into the same shard. In the Citus multi-tenant use case this means that tenants often share shards.
 
-Many multi-tenant SaaS providers face the problem that some tenants grow much bigger than the others, or require special resource guarantees. In Citus Enterprise Edition large tenants can be *isolated* to their own dedicated shards to better control the resource allocation between small and large tenants.
+However sharing shards can cause resource contention when tenants differ drastically in size. This is a common situation for systems with a large number of tenants -- we have observed that the size of tenant data tend to follow a Zipfian distribution as the number of tenants increases. This means there are a few very large tenants, and many smaller ones. To improve resource allocation and make guarantees of tenant QoS it is worthwhile to move large tenants to dedicated nodes.
+
+Citus Enterprise Edition and :ref:`Citus Cloud <cloud_overview>` provide the tools to isolate a tenant on a specific node. This happens in two phases: 1) isolating the tenant's data to a new dedicated shard, then 2) moving the shard to the desired node. To understand the process it helps to know precisely how rows of data are assigned to shards.
 
 Every shard is marked in Citus metadata with the range of hashed values it contains (more info in the reference for :ref:`pg_dist_shard <pg_dist_shard>`). The Citus UDF :code:`isolate_tenant_to_new_shard(table_name, tenant_id)` moves a tenant into a dedicated shard in three steps:
 
--- a/admin_guide/cluster_management.rst
+++ b/admin_guide/cluster_management.rst
@@ -42,31 +42,6 @@ The Citus master only stores metadata about the table shards and does not store 
 
 However, in some write heavy use cases where the master becomes a performance bottleneck, users can add another master. As the metadata tables are small (typically a few MBs in size), it is possible to copy over the metadata onto another node and sync it regularly. Once this is done, users can send their queries to any master and scale out performance. If your setup requires you to use multiple masters, please `contact us <https://www.citusdata.com/about/contact_us>`_.
 
-Multi-tenancy
-#############
-
-Citus provides ways to customize table distribution for the multi-tenant use case. By default Citus assigns table rows to shards using the table's distribution column and method. Many multi-tenant SaaS providers need to override this logic for particular tenants for a few reasons:
-
-* Preferential treatment. Providers may have a large (or important) customer whose data warrants dedicated resources. For instance, they want to ensure that customer_id = 10 gets its premium share.
-* Protection against noisy neighbors. Providers may have a noisy customer whose requests are hurting the performance of other customers assigned to the same shard. Providers would like to ensure the noisy customer receives fewer resources by isolating it on smaller instances.
-* Supporting pre-production testing or internal development. Providers are free to experiment with new queries in an isolated test node in the production cluster. Problems in the test node won't affect customers. Once everything looks good then the changes can be applied throughout the cluster.
-
-Citus Enterprise provides a UDF to override the shard placement for a tenant.
-
-.. code-block:: postgresql
-
-  -- This query creates an isolated shard for the given tenant_id and returns the new shard id.
-  SELECT isolate_tenant_to_new_shard('table_name', tenant_id);
-
-  -- For example
-  SELECT isolate_tenant_to_new_shard('lineitem', 135);
-
-  -- If the given table has co-located tables, the query above errors out and
-  -- advises to use the CASCADE option
-  SELECT isolate_tenant_to_new_shard('lineitem', 135, 'CASCADE');
-
-The CASCADE option isolates a table and all tables :ref:`co-located <colocation>` with it.
-
 .. _dealing_with_node_failures:
 
 Dealing With Node Failures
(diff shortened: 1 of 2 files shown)


sumedhpathak:
I recall us discussing making a separate section for this. Does this really fall under cluster management? Or is it Table Management?

admin_guide/cluster_management.rst
```diff
@@ -42,6 +42,31 @@ The Citus master only stores metadata about the table shards and does not store
 
 However, in some write heavy use cases where the master becomes a performance bottleneck, users can add another master. As the metadata tables are small (typically a few MBs in size), it is possible to copy over the metadata onto another node and sync it regularly. Once this is done, users can send their queries to any master and scale out performance. If your setup requires you to use multiple masters, please `contact us <https://www.citusdata.com/about/contact_us>`_.
 
+Multi-tenancy
```
sumedhpathak: The title here doesn't really give much information. I'd rather we have something more explicit, like "Isolating tenants in a multi-tenant table" (Or something like that).

admin_guide/cluster_management.rst
```diff
@@ -42,6 +42,31 @@ The Citus master only stores metadata about the table shards and does not store
...
+#############
+
+Citus provides ways to customize table distribution for the multi-tenant use case. By default Citus assigns table rows to shards using the table's distribution column and method. Many multi-tenant SaaS providers need to override this logic for particular tenants for a few reasons:
+
+* Preferential treatment. Providers may have a large (or important) customer whose data warrants dedicated resources. For instance, they want to ensure that customer_id = 10 gets its premium share.
+* Protection against noisy neighbors. Providers may have a noisy customer whose requests are hurting the performance of other customers assigned to the same shard. Providers would like to ensure the noisy customer receives fewer resources by isolating it on smaller instances.
+* Supporting pre-production testing or internal development. Providers are free to experiment with new queries in an isolated test node in the production cluster. Problems in the test node won't affect customers. Once everything looks good then the changes can be applied throughout the cluster.
```
sumedhpathak: @samay-sharma Is this a reason that comes up for this feature? Do the other two reasons resonate from an SE perspective?

begriffs:
My notes for tenant isolation say, "add section to admin section about multi-tenancy." Maybe you're thinking vacuum analyze? The notes for that one say, "Find a good place for it in docs, maybe admin -> Table Management." Unless the idea is that they both go under a "Table Management" section, which contains subsections for isolation and for vacuum.

admin_guide/cluster_management.rst
```diff
@@ -42,6 +42,31 @@ The Citus master only stores metadata about the table shards and does not store
 
 However, in some write heavy use cases where the master becomes a performance bottleneck, users can add another master. As the metadata tables are small (typically a few MBs in size), it is possible to copy over the metadata onto another node and sync it regularly. Once this is done, users can send their queries to any master and scale out performance. If your setup requires you to use multiple masters, please `contact us <https://www.citusdata.com/about/contact_us>`_.
 
+Multi-tenancy
+#############
+
+Citus provides ways to customize table distribution for the multi-tenant use case. By default Citus assigns table rows to shards using the table's distribution column and method. Many multi-tenant SaaS providers need to override this logic for particular tenants for a few reasons:
```
sumedhpathak: This paragraph is a bit different from Metin's original documentation of this feature. Primarily, it doesn't override any table-distribution logic, or allow customization of this logic. It still uses hash-partitioning, but just isolates the hash-value for the provided tenant into its own shard. Given this, Citus still assigns rows to shards using the distribution column and method.

admin_guide/cluster_management.rst
```diff
@@ -42,6 +42,31 @@ The Citus master only stores metadata about the table shards and does not store
...
+
+  -- For example
+  SELECT isolate_tenant_to_new_shard('lineitem', 135);
+
+  -- If the given table has co-located tables, the query above errors out and
+  -- advises to use the CASCADE option
+  SELECT isolate_tenant_to_new_shard('lineitem', 135, 'CASCADE');
+
+The CASCADE option isolates a table and all tables :ref:`co-located <colocation>` with it.
+
```
sumedhpathak: We can add a separate paragraph here to use the Enterprise function `master_move_shard_placement` to move isolated shards on to a different node. We should clarify that by default the isolated shard gets created on the same node.

admin_guide/cluster_management.rst
```diff
@@ -42,6 +42,31 @@ The Citus master only stores metadata about the table shards and does not store
...
+Multi-tenancy
+#############
+
+Citus provides ways to customize table distribution for the multi-tenant use case. By default Citus assigns table rows to shards using the table's distribution column and method. Many multi-tenant SaaS providers need to override this logic for particular tenants for a few reasons:
+
+* Preferential treatment. Providers may have a large (or important) customer whose data warrants dedicated resources. For instance, they want to ensure that customer_id = 10 gets its premium share.
+* Protection against noisy neighbors. Providers may have a noisy customer whose requests are hurting the performance of other customers assigned to the same shard. Providers would like to ensure the noisy customer receives fewer resources by isolating it on smaller instances.
+* Supporting pre-production testing or internal development. Providers are free to experiment with new queries in an isolated test node in the production cluster. Problems in the test node won't affect customers. Once everything looks good then the changes can be applied throughout the cluster.
+
+Citus Enterprise provides a UDF to override the shard placement for a tenant.
```
sumedhpathak: Can we make this (or add a separate) `Note`, that this feature is only available in Citus Enterprise?

admin_guide/cluster_management.rst
```diff
@@ -127,3 +127,45 @@ place for the master node, users can dynamically reconstruct this metadata from
 shard information available on the worker nodes. To learn more about the metadata
 tables and their schema, you can visit the :ref:`metadata_tables` section of our documentation.
 
+Tenant Isolation
+################
+
+Citus places table rows into worker shards based on the hashed value of the rows' distribution column.  Multiple distribution column values often fall into the same shard. In the Citus multi-tenant use case this means that tenants often share shards.
```
begriffs: This is only partially true because there are distribution methods other than hash, but I wasn't sure whether I should complicate the paragraph with all that.

sumedhpathak: Agreed to leave it out. We're not emphasizing append, so its OK to just skip that.

sumedhpathak: Is there an extra space between these sentences:?
```distribution column.  Multiple```?

File: admin_guide/cluster_management.rst [hunk 7]
```diff
@@ -127,3 +127,45 @@ place for the master node, users can dynamically reconstruct this metadata from
...
+
+Many multi-tenant SaaS providers face the problem that some tenants grow much bigger than the others, or require special resource guarantees. In Citus Enterprise Edition large tenants can be *isolated* to their own dedicated shards to better control the resource allocation between small and large tenants.
+
+Every shard is marked in Citus metadata with the range of hashed values it contains (more info in the reference for :ref:`pg_dist_shard <pg_dist_shard>`). The Citus UDF :code:`isolate_tenant_to_new_shard` removes rows with a given distribution value from their current shard placement and creates a shard which will hold them separately from other rows. Furthermore, it takes a :code:`CASCADE` option which isolates not just the table but all tables :ref:`co-located <colocation>` with it.
```
sumedhpathak: Two notes:

1. We say: "Removes rows from current shard placement and creates a shard..." We use shard-placement in one, and shard in the other. Can we use just top-level "shard" as the concept here?

2. "Furthermore, it takes a :code:`CASCADE` option which isolates not just the table but all tables :ref:`co-located <colocation>` with it." Instead, can we say "isolates tenant from not just the provided table, but also isolates the same tenant from all co-located tables"? We're not isolating the table, but the tenant from the table.

admin_guide/cluster_management.rst
```diff
@@ -127,3 +127,45 @@ place for the master node, users can dynamically reconstruct this metadata from
...
+  -- For example
+
+  SELECT isolate_tenant_to_new_shard('lineitem', 135);
+
+  -- If the given table has co-located tables, the query above errors out and
+  -- advises to use the CASCADE option
+
+  SELECT isolate_tenant_to_new_shard('lineitem', 135, 'CASCADE');
+
+Once a tenant is isolated to a dedicated shard it can be moved to a separate node in the Citus cluster for true hardware isolation. As mentioned, the :code:`isolate_tenant_to_new_shard` function returns the newly created shard id, and this id can be used to move the shard:
```
sumedhpathak: Maybe clarify that the function will create a shard, but places it on the same node as the original shard? That can be a preamble for the rest of this. We don't note that point anywhere.

Could we also note that this move function (along with tenant-isolation) both belong in Citus Enterprise only?

admin_guide/cluster_management.rst
```diff
@@ -127,3 +127,45 @@ place for the master node, users can dynamically reconstruct this metadata from
...
+  SELECT isolate_tenant_to_new_shard('table_name', tenant_id);
+
+  -- For example
+
+  SELECT isolate_tenant_to_new_shard('lineitem', 135);
+
+  -- If the given table has co-located tables, the query above errors out and
+  -- advises to use the CASCADE option
+
+  SELECT isolate_tenant_to_new_shard('lineitem', 135, 'CASCADE');
```
sumedhpathak: Could we run this and also show output which is the new shard-id? (Then use this in the below move example)?

admin_guide/cluster_management.rst
```diff
@@ -127,3 +127,64 @@ place for the master node, users can dynamically reconstruct this metadata from
...
+
+  Tenant isolation is a feature of **Citus Enterprise Edition** only.
+
+Citus places table rows into worker shards based on the hashed value of the rows' distribution column. Multiple distribution column values often fall into the same shard. In the Citus multi-tenant use case this means that tenants often share shards.
+
+Many multi-tenant SaaS providers face the problem that some tenants grow much bigger than the others, or require special resource guarantees. In Citus Enterprise Edition large tenants can be *isolated* to their own dedicated shards to better control the resource allocation between small and large tenants.
+
+Every shard is marked in Citus metadata with the range of hashed values it contains (more info in the reference for :ref:`pg_dist_shard <pg_dist_shard>`). The Citus UDF :code:`isolate_tenant_to_new_shard(table_name, tenant_id)` moves a tenant into a dedicated shard in three steps:
+
+1. Creates a new shard for :code:`table_name` which (a) includes rows whose distribution column has value :code:`tenant_id` and (b) excludes all other rows.
```
begriffs: Maybe this 1, 2, 3 is too much gory detail but I thought it might clarify how things work.

admin_guide/cluster_management.rst
begriffs: (done)

admin_guide/cluster_management.rst
```diff
@@ -127,3 +127,64 @@ place for the master node, users can dynamically reconstruct this metadata from
 shard information available on the worker nodes. To learn more about the metadata
 tables and their schema, you can visit the :ref:`metadata_tables` section of our documentation.
 
+Tenant Isolation
+################
+
+.. note::
+
+  Tenant isolation is a feature of **Citus Enterprise Edition** only.
```
lfittl: I wonder if this should say:

```markdown
a feature of **Citus Enterprise Edition** and **Citus Cloud** only.
```

sumedhpathak: @lfittl Agreed. @begriffs Could you make this change as well?

begriffs:
Note to self: change references from `cloud` to `cloud_overview` after rebasing on master.

admin_guide/cluster_management.rst
```diff
@@ -82,3 +82,64 @@ place for the master node, users can dynamically reconstruct this metadata from
...
+Tenant Isolation
+################
+
+.. note::
+
+  Tenant isolation is a feature of **Citus Enterprise Edition** and :ref:`Citus Cloud <cloud>` only.
+
+Citus places table rows into worker shards based on the hashed value of the rows' distribution column. Multiple distribution column values often fall into the same shard. In the Citus multi-tenant use case this means that tenants often share shards.
+
+Many multi-tenant SaaS providers face the problem that some tenants grow much bigger than the others, or require special resource guarantees. In Citus Enterprise Edition large tenants can be *isolated* to their own dedicated shards to better control the resource allocation between small and large tenants.
```
samay-sharma: Could we add more context on the motivation. Maybe say something like:- "When you have a large number of tenants, not all tenants are equally sized. Generally, we've seen that data follows a Zipf's distribution where you would have a few tenants which have a large amount of data and a large number of smaller sized tenants. Once your bigger tenants grow to a large size, it might be worthwhile to move them to a different node. This would allow you to provide better resource allocation and guarantees to larger tenants and also give more resources to other smaller tenants.  To enable this, Citus enterprise edition allows you to isolate ......

begriffs:
@samay-sharma how do you like my edit for motivation?

sumedhpathak:
@samay-sharma ping to take a quick look at the edits

sumedhpathak:
@begriffs Changes looks reasonable. 🚢

PR #252, opened 2017-01-19
Update docs for create_reference_table

begriffs:
Fixes #229

PR #272, opened 2017-02-01
Explain upgrading a broadcast table to reference status

### LINKED ISSUE #261: Document `upgrade_to_reference_table`

begriffs:
Includes a detail omitted from #252

Closes #261

### DIFF
--- a/dist_tables/ddl.rst
+++ b/dist_tables/ddl.rst
@@ -82,6 +82,12 @@ The above method distributes tables into multiple horizontal shards, but it's al
 
 Other queries, such as one calculating tax for a shopping cart, can join on the :code:`states` table with no network overhead.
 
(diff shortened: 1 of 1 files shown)


dist_tables/ddl.rst
```diff
@@ -82,6 +82,12 @@ The above method distributes tables into multiple horizontal shards, but it's al
 
 Other queries, such as one calculating tax for a shopping cart, can join on the :code:`states` table with no network overhead.
 
+In addition to distributing a table as a single replicated shard, the :code:`create_reference_table` UDF marks it as a reference table in the Citus metadata tables. Citus automatically performs two-phase commits (`2PC <https://en.wikipedia.org/wiki/Two-phase_commit_protocol>`_) for modifications to tables marked this way, which provides strong consistency guarantees. If you have an existing distributed table which has a shard count of one and is replicated across all nodes, you can upgrade it to be a recognized reference table by running
```
sumedhpathak: Also, one correction: Looking at the code, you need to have an existing table which has a shard count of 1. The user doesn't need to have it explicitly replicated on all nodes, as the code seems to be doing that:
https://github.com/citusdata/citus/blob/4929e2b5035012fca6903e9b108d748204ba2c3c/src/backend/distributed/utils/reference_table_utils.c#L110

PR #277, opened 2017-02-08
Update tutorials to Citus 6.1

begriffs:
Built new binaries using Citus v6.1.0-rc2 and added them to S3. Tested tutorial instructions on OS X and Ubuntu.

### DIFF
--- a/tutorials/tut-hash-distribution.rst
+++ b/tutorials/tut-hash-distribution.rst
@@ -14,9 +14,9 @@ To do the tutorial you'll need a single-machine Citus cluster with a master and 
 .. raw:: html 
 
   <ul>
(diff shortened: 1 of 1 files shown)


sumedhpathak:
@begriffs Did you run into any issues on Mac? I'm getting an error with the ingest script on Mac's. The Linux one seems to work fine on an AWS instance.

begriffs:
I did run it on Mac. Encountered an intermittent failure that seems to be related to the Wikipedia API. See if the problem clears up if you run it again.

sumedhpathak:
Ok, Looks good. I was distracted for a fair bit by the script not working on my Mac in the public wi-fi network. It seems to work reliably at home. Also tested the Linux bundle on an AWS machine.

begriffs:
I opened https://github.com/citusdata/citus-tutorial/issues/10 about a way to potentially make the script more robust.

PR #295, opened 2017-02-22
Update docker instructions

### LINKED ISSUE #292: Docker installation steps have discrepancies between docs and citus/readme.md
I ran through the Docker installation steps in our docs and noticed a few discrepancies between them and our Readme.md. Also, I had already installed following the steps in our readme.md and the steps in our docs didn't work for me.

1. We give the user the option to install Docker Engine or Docker Compose in our docs. What's the difference? Can we recommend the option that we think is the best?
2. In our Readme.md, we recommend Docker Compose. We provide separate steps for Linux or Mac. We also ask the user to run ``eval`` on Mac. We don't provide those steps in our docs. Why?

begriffs:
Fixes #292 (I will subsequently open a pull request to update the Citus readme as well).

Tested the Docker installation instructions under Ubuntu 16.04 and CentOS 7.3.

### DIFF
--- a/installation/single_machine_docker.rst
+++ b/installation/single_machine_docker.rst
@@ -8,9 +8,11 @@ This section describes setting up a Citus cluster on a single machine using dock
 **1. Install docker and docker-compose**
 
 * Install the `Docker Platform <https://www.docker.com/products/overview#/install_the_platform>`_.
-* On most platforms this will install docker-compose as well, but on Linux you'll need to install docker-compose manually by downloading the binary and putting it in your path:
+* (Linux only) On most platforms the previous step will install docker-compose as well as Docker, but on Linux you'll need to install docker-compose manually by downloading the binary and putting it in your path:
 
   .. code-block:: bash
+
+    # Linux ONLY
 
     curl -L https://github.com/docker/compose/releases/download/1.11.1/docker-compose-`uname -s`-`uname -m` > /usr/local/bin/docker-compose
     chmod +x /usr/local/bin/docker-compose
--- a/installation/single_machine_docker.rst
+++ b/installation/single_machine_docker.rst
@@ -7,7 +7,13 @@ This section describes setting up a Citus cluster on a single machine using dock
 
 **1. Install docker and docker-compose**
 
-Follow the installation instructions for `Docker Engine <https://docs.docker.com/engine/installation/>`_ and `Docker Compose <https://docs.docker.com/compose/install/>`_ on your platform.
+* Install the `Docker Platform <https://www.docker.com/products/overview#/install_the_platform>`_.
+* On most platforms this will install docker-compose as well, but on Linux you'll need to install docker-compose manually by downloading the binary and putting it in your path:
+
+  .. code-block:: bash
+
+    curl -L https://github.com/docker/compose/releases/download/1.11.1/docker-compose-`uname -s`-`uname -m` > /usr/local/bin/docker-compose
+    chmod +x /usr/local/bin/docker-compose
 
 **2. Start the Citus Cluster**
 
--- a/installation/single_machine_docker.rst
+++ b/installation/single_machine_docker.rst
@@ -21,7 +21,7 @@ Citus uses docker-compose to run and connect containers holding the database mas
 
 .. code-block:: bash
 
-  wget https://raw.githubusercontent.com/citusdata/docker/master/docker-compose.yml
+  curl -L https://raw.githubusercontent.com/citusdata/docker/master/docker-compose.yml > docker-compose.yml
   docker-compose -p citus up -d
 
 The first time you start the cluster it builds its containers. Subsequent startups take a matter of seconds.
--- a/installation/single_machine_docker.rst
+++ b/installation/single_machine_docker.rst
@@ -20,7 +20,7 @@ Linux
 .. code-block:: bash
 
   curl -sSL https://get.docker.com/ | sudo bash
-  sudo usermod -aG docker $USER && exec su -l $USER
+  sudo usermod -aG docker $USER && newgrp docker
   sudo systemctl start docker
 
   sudo curl -L https://github.com/docker/compose/releases/download/1.11.1/docker-compose-`uname -s`-`uname -m` -o /usr/local/bin/docker-compose
--- a/installation/single_machine_docker.rst
+++ b/installation/single_machine_docker.rst
@@ -21,6 +21,7 @@ Linux
 
   curl -sSL https://get.docker.com/ | sudo bash
   sudo usermod -aG docker $USER && exec su -l $USER
+  sudo systemctl start docker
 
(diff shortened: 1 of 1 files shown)


begriffs:
Here you go, @samay-sharma, give it a try.

installation/single_machine_docker.rst
```diff
@@ -7,15 +7,23 @@ This section describes setting up a Citus cluster on a single machine using dock
...
 
-Follow the installation instructions for `Docker Engine <https://docs.docker.com/engine/installation/>`_ and `Docker Compose <https://docs.docker.com/compose/install/>`_ on your platform.
+* Install the `Docker Platform <https://www.docker.com/products/overview#/install_the_platform>`_.
+* (Linux only) On most platforms the previous step will install docker-compose as well as Docker, but on Linux you'll need to install docker-compose manually by downloading the binary and putting it in your path:
+
+  .. code-block:: bash
+
+    # Linux ONLY
+
+    curl -L https://github.com/docker/compose/releases/download/1.11.1/docker-compose-`uname -s`-`uname -m` > /usr/local/bin/docker-compose
```
begriffs: We could alternately link to the instructions in https://github.com/docker/compose/releases/latest

This has the advantage of not hardcoding a docker-compose version, but the disadvantage of being harder to follow than our reproducing the command here.

installation/single_machine_docker.rst
```diff
@@ -7,15 +7,23 @@ This section describes setting up a Citus cluster on a single machine using dock
...
 
 **2. Start the Citus Cluster**
 
 Citus uses docker-compose to run and connect containers holding the database master node, workers, and a persistent data volume. To create a local cluster download our docker-compose configuration file and run it
 
 .. code-block:: bash
 
-  wget https://raw.githubusercontent.com/citusdata/docker/master/docker-compose.yml
+  curl -L https://raw.githubusercontent.com/citusdata/docker/master/docker-compose.yml > docker-compose.yml
   docker-compose -p citus up -d
```
samay-sharma: After installing docker, I had to start docker before I was able to do docker-compose. Could we add a step to the instructions which tells users to do that?

begriffs: Come to think of it, when I tested on Ubuntu and CentOS the installer also started the daemon automatically. It's only the Mac installation that requires a start step. So I added a "Mac Only" step to the list.

installation/single_machine_docker.rst
jasonmp85: It's super annoying that we have to ask users to add a repo to their box (to download Engine) only to turn around and ask them to curl some binary into their path.

docker/compose#2235 is a request to add `docker-compose` to the repository. I have _no clue_ why this isn't already the case, but dropping a reference here in order to register my displeasure with upstream.

installation/single_machine_docker.rst
```diff
@@ -7,15 +7,31 @@ This section describes setting up a Citus cluster on a single machine using dock
...
+* Install the `Docker Platform <https://www.docker.com/products/overview#/install_the_platform>`_.
+* (Mac only) Start the Docker daemon
+
+  .. code-block:: bash
+
+    # Mac ONLY
+
+    open /Applications/Docker.app
+
+* (Linux only) On most platforms the previous step will install docker-compose as well as Docker, but on Linux you'll need to install docker-compose manually by downloading the binary and putting it in your path:
```
ozgune: Small rewording.

On Linux, you need to manually install docker-compose (you can find the latest docker-compose version [here](link)). For this, you need to download the binary and put it in your path:

installation/single_machine_docker.rst
ozgune: Super minor unrelated point for the note below (line 41).

If you have PostgreSQL running on your machine -> If you already have PostgreSQL running on your machine

installation/single_machine_docker.rst
```diff
@@ -5,24 +5,44 @@ Docker
...
+* Start Docker by clicking on the application's icon.
+
+Linux
+~~~~~
+
+.. code-block:: bash
+
+  curl -sSL https://get.docker.com/ | sudo bash
+  sudo usermod -aG docker $USER && exec su -l $USER
+  sudo systemctl start docker
```
begriffs: In my CentOS test Docker did not start automatically after non-root installation. Adding `systemctrl start` fixed that, and also causes no harm in Ubuntu where the service was already started.

The `systemctl` works for Linux distros that use systemd, which I think is recent versions of all major distros. @jasonmp85 does this seem like a good bet to you?

installation/single_machine_docker.rst
```diff
@@ -5,24 +5,44 @@ Docker
...
+* Install the `Docker Platform <https://www.docker.com/products/overview#/install_the_platform>`_.
+* Start Docker by clicking on the application's icon.
+
+Linux
+~~~~~
+
+.. code-block:: bash
+
+  curl -sSL https://get.docker.com/ | sudo bash
+  sudo usermod -aG docker $USER && exec su -l $USER
```
ozgune: I spinned up a Redhat instance on AWS and got stock at the ``exec su -l $USER`` step.

This is because I have ``sudo`` on AWS instances but not ``su``.

begriffs:
Waiting to test this on a few EC2 machines until the S3 outage (!) abates.

begriffs:
@ozgune these instructions worked for me on Red Hat, Ubuntu and Debian on AWS. I was logged in as ec2-user, ubuntu, and admin for each system respectively (not root). The only thing that went wrong is that Debian on AWS did not come pre-bundled with curl and I had to install that. However I'm pretty confident that curl is a common enough tool that our users will either have it or know how to get it.

jasonmp85:
> The only thing that went wrong is that Debian on AWS did not come pre-bundled with curl and I had to install that.

This is also true of the Debian-based Docker images themselves. They're pretty barebones and don't include `man` or `curl` or `which` or many other common things. Weird that the AWS image itself doesn't have any `curl`, but not something I haven't come across frequently.

I don't think it's e.g. worth investigating `wget`'s availability as an alternative, since if that's the sole place you ran into this we're probably safe, and it's probably a common enough annoyance that users of that platform know where and how to get `curl`. (i.e. the colloquial name for the idiom is `curl | bash`, not `wget | bash`).

begriffs:
The debian one was actually a community AMI. (ami-0343ae47)

ozgune:
@begriffs -- Sounds good, let's 🚀 

I had two side notes. First, we'll need to resolve a minor conflict with the tutorial changes before merging this.

Second, I played around with a few AMIs and things looked good. On SUSE 11, I ran into the following issue. I think that's perfectly acceptable since Docker seems to only support SUSE 12+. The error message here was also pretty descriptive. JFYI:

<pre>
ec2-user@ip-10-203-133-221:~> curl -sSL https://get.docker.com/ | sudo bash

  Either your platform is not easily detectable, is not supported by this
  installer script (yet - PRs welcome! [hack/install.sh]), or does not yet have
  a package for Docker.  Please visit the following URL for more detailed
  installation instructions:

    https://docs.docker.com/engine/installation/
</pre>

PR #296, opened 2017-02-23
Document get_shard_id_for_distribution_column

### LINKED ISSUE #225: Document how determine the shard to which a dist col value maps

begriffs:
Fixes #225

I documented the function in the UDF reference section because I don't know a concrete activity or user goal that requires its use. If someone knows more about the context we can move my change to a more relevant part of the docs.

### DIFF
--- a/admin_guide/cluster_management.rst
+++ b/admin_guide/cluster_management.rst
@@ -145,3 +145,32 @@ The new shard(s) are created on the same node as the shard(s) from which the ten
     102240,
     'source_host', source_port,
     'dest_host', dest_port);
+
+Diagnostics
+###########
+
+.. _row_placements:
+
+Finding Row Placements
+----------------------
+
+Each row of a distributed table is assigned to a shard, and that shard is placed on a worker node in the Citus cluster. To determine which worker node contains a specific row, we put together two pieces of information: the :ref:`shard id <get_shard_id>` associated with the row's distribution column value, and the shard placements on workers. We can combine this into a single query. Here is how to find the placement of a row of :code:`my_table` whose distribution column has value 4:
+
+.. code-block:: postgresql
+
+  SELECT *
+    FROM pg_dist_shard_placement
+   WHERE shardid = (
+     SELECT get_shard_id_for_distribution_column('my_table', 4)
+   );
+
+The output contains the host and port of the worker database.
+
+::
+
+  ┌─────────┬────────────┬─────────────┬───────────┬──────────┬─────────────┐
+  │ shardid │ shardstate │ shardlength │ nodename  │ nodeport │ placementid │
+  ├─────────┼────────────┼─────────────┼───────────┼──────────┼─────────────┤
+  │  102009 │          1 │           0 │ localhost │     5433 │           2 │
+  └─────────┴────────────┴─────────────┴───────────┴──────────┴─────────────┘
+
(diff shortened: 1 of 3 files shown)


sumedhpathak:
Maybe we should also add this to Cluster Management? I don't know what the title would be yet. 
There we'd want a more end-to-end solution. The UDF and example you've shown just shows the shard-id. We can then also show a query for that Id in pg_dist_shard/pg_dist_shard_placement to show the physical location of that shard?

We could also add an FAQ entry (e.g.: How can I find which shard a partition column maps to?)
which could link to that information.

begriffs:
See what you think about this for a start. It adds slightly more context.

admin_guide/cluster_management.rst
```diff
@@ -145,3 +145,32 @@ The new shard(s) are created on the same node as the shard(s) from which the ten
     102240,
     'source_host', source_port,
     'dest_host', dest_port);
+
+Diagnostics
+###########
+
+.. _row_placements:
+
+Finding Row Placements
```
begriffs: "Row placement" is a nonstandard term, I could go with "Finding rows in worker nodes" or something.

sumedhpathak: How about something like "which shard contains data for a particular tenant?" That might be the FAQ question. The section here might be: "Finding which shard contains data for a specific tenant".

This is clearly more tenant specific, but that *might* be OK.

admin_guide/cluster_management.rst
```diff
@@ -145,3 +145,32 @@ The new shard(s) are created on the same node as the shard(s) from which the ten
...
+----------------------
+
+Each row of a distributed table is assigned to a shard, and that shard is placed on a worker node in the Citus cluster. To determine which worker node contains a specific row, we put together two pieces of information: the :ref:`shard id <get_shard_id>` associated with the row's distribution column value, and the shard placements on workers. We can combine this into a single query. Here is how to find the placement of a row of :code:`my_table` whose distribution column has value 4:
```
sumedhpathak: If we go with the above title suggestion, we could amend this paragraph a bit to not only determine which shard contains a specific row, but which shard contains data for a specific tenant.

faq/faq.rst
```diff
@@ -78,3 +78,8 @@ In which situations are uniqueness constraints supported on distributed tables?
 Citus is able to enforce a primary key or uniqueness constraint only when the constrained columns contain the distribution column. In particular this means that if a single column constitutes the primary key then it has to be the distribution column as well.
 
 This restriction allows Citus to localize a uniqueness check to a single shard and let PostgreSQL on the worker node do the check efficiently.
+
+How do I find which worker nodes contain which rows of a distributed table?
```
sumedhpathak: See above title suggestion.

I'd also suggest maybe adding a very small paragraph here? Like, Citus provides some UDF's, which allow you to map a partition-column value (or tenant) to a particular shard. See <link> for more details.

admin_guide/cluster_management.rst
```diff
@@ -145,3 +145,32 @@ The new shard(s) are created on the same node as the shard(s) from which the ten
...
+
+.. code-block:: postgresql
+
+  SELECT *
+    FROM pg_dist_shard_placement
+   WHERE shardid = (
+     SELECT get_shard_id_for_distribution_column('my_table', 4)
```
sumedhpathak: I'll also restate my preference for real-world examples. Could we provide an example using our multi-tenant schema? 

E.g.: Finding which orders shard contains orders for store Gap.com (which has tenant-id 4) or something like that.

PR #297, opened 2017-02-23
Distributed vacuum

### LINKED ISSUE #239: Document new VACUUM/ANALYZE support
_N.B. — I will mostly say `VACUUM` here; `ANALYZE` also applies in most cases_

Connected to citusdata/citus#719 (original issue)
Connected to citusdata/citus#1013 (implementation)

## Why are we implementing it?

MVCC requires `VACUUM` operations. While auto-`VACUUM` mechanisms are probably good enough for most cases, some users may prefer to manually `VACUUM` particular tables. Until now, that required logging into each worker.

For rables with particular work loads, or companies with certain "safe" hours to schedule a `VACUUM`, it might make more sense to manually `VACUUM` a table rather than 

## Communication goals

Reference, mostly. Just some coverage of the fact that `VACUUM <dist_table>` will actually send that to workers.

We don't seem to have a "General Maintenance" section, but it would probably belong there. I noticed `TRUNCATE` is also missing; it should belong in a similar section, potentially.

## How does this work?

If `citus.enable_ddl_propagation` is set to `on`, using `VACUUM` against a distributed table will send a `VACUUM` command to every one of that table's placements (one connection per placement). This is done in parallel.

All options are supported (including the `column_list` parameter) except for `VERBOSE`. The master `VACUUM` command also runs (on the master), and does so before any workers are notified.

Unqualified `VACUUM` commands (i.e. those without a table specified) do not propagate: a warning informs the user as much.

### Example sql

```sql
VACUUM my_table;
ANALYZE my_table;
```

See [the PostgreSQL documentation](https://www.postgresql.org/docs/9.6/static/sql-analyze.html) for more.

### Corner cases, gotchas

None worthy of noting.

Commit: citusdata/citus@6f95875191da97a454e3563bb0fbab7a8cbd010d
[Tests](https://github.com/citusdata/citus/blob/6f95875191da97a454e3563bb0fbab7a8cbd010d/src/test/regress/sql/multi_utilities.sql#L44-L175)

begriffs:
Fixes #239 

I use the word "worker" when referring to auto-vacuum worker processes, and also for worker nodes. Hope it's not confusing, let me know what you think.

### DIFF
--- a/admin_guide/cluster_management.rst
+++ b/admin_guide/cluster_management.rst
@@ -145,3 +145,24 @@ The new shard(s) are created on the same node as the shard(s) from which the ten
     102240,
     'source_host', source_port,
     'dest_host', dest_port);
+
+Database Bloat
+##############
+
+It’s not just user queries which scale in a distributed database, vacuuming does too. In PostgreSQL big busy tables have great potential to bloat, both from lower sensitivity to PostgreSQL's vacuum scale factor, and generally because of the extent of their row churn. Splitting a table into distributed shards means both that individual shards are smaller tables and that auto-vacuum workers can parallelize over different parts of the table on different machines. Ordinarily auto-vacuum can only run one worker per table.
+
+Due to the above, auto-vacuum operations on a Citus cluster are probably good enough for most cases. However for tables with particular workloads, or companies with certain "safe" hours to schedule a vacuum, it might make more sense to manually vacuum a table rather than leaving all the work to auto-vacuum.
+
+To vacuum a table, simply run this on the coordinator node:
+
+.. code-block:: postgresql
+
+  VACUUM my_table;
+
+If :code:`citus.enable_ddl_propagation` is set to :code:`on` (the default value) then using vacuum against a distributed table will send a vacuum command to every one of that table's placements (one connection per placement). This is done in parallel. All options are supported (including the :code:`column_list` parameter) except for :code:`VERBOSE`. The vacuum command also runs on the coordinator, and does so before any workers nodes are notified. Note that unqualified vacuum commands (i.e. those without a table specified) do not propagate to worker nodes.
+
+Analyzing distributes in the same way:
+
+.. code-block:: postgresql
+
+  ANALYZE my_table;
--- a/admin_guide/cluster_management.rst
+++ b/admin_guide/cluster_management.rst
@@ -145,24 +145,3 @@ The new shard(s) are created on the same node as the shard(s) from which the ten
     102240,
     'source_host', source_port,
     'dest_host', dest_port);
-
-Database Bloat
-##############
-
-It’s not just user queries which scale in a distributed database, vacuuming does too. In PostgreSQL big busy tables have great potential to bloat, both from lower sensitivity to PostgreSQL's vacuum scale factor, and generally because of the extent of their row churn. Splitting a table into distributed shards means both that individual shards are smaller tables and that auto-vacuum workers can parallelize over different parts of the table on different machines. Ordinarily auto-vacuum can only run one worker per table.
-
-Due to the above, auto-vacuum operations on a Citus cluster are probably good enough for most cases. However for tables with particular workloads, or companies with certain "safe" hours to schedule a vacuum, it might make more sense to manually vacuum a table rather than leaving all the work to auto-vacuum.
-
-To vacuum a table, simply run this on the coordinator node:
-
-.. code-block:: postgresql
-
-  VACUUM my_table;
-
-If :code:`citus.enable_ddl_propagation` is set to :code:`on` (the default value) then using vacuum against a distributed table will send a vacuum command to every one of that table's placements (one connection per placement). This is done in parallel. All options are supported (including the :code:`column_list` parameter) except for :code:`VERBOSE`. The vacuum command also runs on the coordinator, and does so before any workers nodes are notified. Note that unqualified vacuum commands (i.e. those without a table specified) do not propagate to worker nodes.
-
-Analyzing distributes in the same way:
-
-.. code-block:: postgresql
-
-  ANALYZE my_table;
--- a/admin_guide/table_management.rst
+++ b/admin_guide/table_management.rst
@@ -1,0 +1,36 @@
+.. _cluster_management:
+
+Table Management
+$$$$$$$$$$$$$$$$$$
+
+Vacuuming Distributed Tables
+############################
+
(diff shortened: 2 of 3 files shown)


admin_guide/cluster_management.rst
```diff
@@ -145,3 +145,24 @@ The new shard(s) are created on the same node as the shard(s) from which the ten
     102240,
     'source_host', source_port,
     'dest_host', dest_port);
+
+Database Bloat
+##############
+
+It’s not just user queries which scale in a distributed database, vacuuming does too. In PostgreSQL big busy tables have great potential to bloat, both from lower sensitivity to PostgreSQL's vacuum scale factor, and generally because of the extent of their row churn. Splitting a table into distributed shards means both that individual shards are smaller tables and that auto-vacuum workers can parallelize over different parts of the table on different machines. Ordinarily auto-vacuum can only run one worker per table.
```
sumedhpathak: What did you have in mind when you mention 'lower sensitivity to PostgreSQL's vacuum..'?

admin_guide/cluster_management.rst
```diff
@@ -145,3 +145,24 @@ The new shard(s) are created on the same node as the shard(s) from which the ten
...
+
+Due to the above, auto-vacuum operations on a Citus cluster are probably good enough for most cases. However for tables with particular workloads, or companies with certain "safe" hours to schedule a vacuum, it might make more sense to manually vacuum a table rather than leaving all the work to auto-vacuum.
+
+To vacuum a table, simply run this on the coordinator node:
+
+.. code-block:: postgresql
+
+  VACUUM my_table;
+
+If :code:`citus.enable_ddl_propagation` is set to :code:`on` (the default value) then using vacuum against a distributed table will send a vacuum command to every one of that table's placements (one connection per placement). This is done in parallel. All options are supported (including the :code:`column_list` parameter) except for :code:`VERBOSE`. The vacuum command also runs on the coordinator, and does so before any workers nodes are notified. Note that unqualified vacuum commands (i.e. those without a table specified) do not propagate to worker nodes.
```
sumedhpathak: Maybe we can remove the note on `enable_ddl_propagation` if the default value is `on`? For one, I'm not sure if Vacuum qualifies as DDL, and so may be a bit confusing. We could just start with "Using vacuum against a distributed table..."

admin_guide/cluster_management.rst
```diff
@@ -145,3 +145,24 @@ The new shard(s) are created on the same node as the shard(s) from which the ten
...
+
+It’s not just user queries which scale in a distributed database, vacuuming does too. In PostgreSQL big busy tables have great potential to bloat, both from lower sensitivity to PostgreSQL's vacuum scale factor, and generally because of the extent of their row churn. Splitting a table into distributed shards means both that individual shards are smaller tables and that auto-vacuum workers can parallelize over different parts of the table on different machines. Ordinarily auto-vacuum can only run one worker per table.
+
+Due to the above, auto-vacuum operations on a Citus cluster are probably good enough for most cases. However for tables with particular workloads, or companies with certain "safe" hours to schedule a vacuum, it might make more sense to manually vacuum a table rather than leaving all the work to auto-vacuum.
+
+To vacuum a table, simply run this on the coordinator node:
+
+.. code-block:: postgresql
+
+  VACUUM my_table;
```
sumedhpathak: Minor: Maybe `my_distributed_table`?

admin_guide/cluster_management.rst
```diff
@@ -145,3 +145,24 @@ The new shard(s) are created on the same node as the shard(s) from which the ten
     102240,
     'source_host', source_port,
     'dest_host', dest_port);
+
+Database Bloat
```
sumedhpathak: Instead of 'Database Bloat' as a title, could we use something like just "Vacuuming Distributed Tables"?

admin_guide/cluster_management.rst
```diff
@@ -145,3 +145,24 @@ The new shard(s) are created on the same node as the shard(s) from which the ten
     102240,
     'source_host', source_port,
     'dest_host', dest_port);
+
```
sumedhpathak: This doesn't really belong in "Cluster Management". I think we had previously discussed adding a separate section in the Admin Guide for "Table Management", which probably is a more suitable location. Could we do that?

admin_guide/cluster_management.rst
sumedhpathak: I think this paragraph is really nice. It captures the benefits of parallelism for Vacuuming in Citus. 
I have two comments:
- We don't define what 'bloat' is. Can we briefly describe that somewhere?
- Maybe in order to do that, could we also link to PostgreSQL's Vacuum documentation? https://www.postgresql.org/docs/current/static/routine-vacuuming.html

admin_guide/cluster_management.rst
sumedhpathak: Could we also link to PostgreSQL's Vacuum reference section when we talk about options supported?
https://www.postgresql.org/docs/current/static/sql-vacuum.html

admin_guide/cluster_management.rst
```diff
@@ -145,3 +145,24 @@ The new shard(s) are created on the same node as the shard(s) from which the ten
...
+To vacuum a table, simply run this on the coordinator node:
+
+.. code-block:: postgresql
+
+  VACUUM my_table;
+
+If :code:`citus.enable_ddl_propagation` is set to :code:`on` (the default value) then using vacuum against a distributed table will send a vacuum command to every one of that table's placements (one connection per placement). This is done in parallel. All options are supported (including the :code:`column_list` parameter) except for :code:`VERBOSE`. The vacuum command also runs on the coordinator, and does so before any workers nodes are notified. Note that unqualified vacuum commands (i.e. those without a table specified) do not propagate to worker nodes.
+
+Analyzing distributes in the same way:
```
sumedhpathak: Is it worth adding more context on Analyze? (I'm assuming this section isn't called Database Bloat now). Maybe a similar note on what Analyze does, why it might be useful, and links to Postgres documentation?

admin_guide/cluster_management.rst
begriffs: Sure, I can add an intro about bloat.

The vacuum scale factor I mentioned is a database config parameter that affects when autovacuum decides to start working on a table: it starts for every n rows written where n = autovacuum_vacuum_scale_factor * [reltuples in pg_class for table] + autovacuum_vacuum_threshold. So autovacuum won't begin until many rows get updated on a big table unless the scale factor is decreased.

admin_guide/table_management.rst
```diff
@@ -0,0 +1,36 @@
+.. _cluster_management:
```
sumedhpathak: Should this be `_table_management`?

begriffs: Good catch. I don't think we need a label at all for this one actually.

PR #301, opened 2017-02-28
Move dataset pull from DDL to DML section where its used.

sumedhpathak:
This had bugged me for a while, so opening a PR for it. Just moving the `wget` of the data files we use from DDL to DML where they are actually used in the `COPY` example.  The previous location was in the DDL section, where those files are never used.

### DIFF
--- a/dist_tables/dml.rst
+++ b/dist_tables/dml.rst
@@ -3,7 +3,7 @@
 Ingesting, Modifying Data (DML)
 ###############################
 
-The following code snippets use the distributed tables example, see :ref:`ddl`.
+The following code snippets use the Github events example, see :ref:`ddl`.
 
 Inserting Data
 --------------
--- a/dist_tables/dml.rst
+++ b/dist_tables/dml.rst
@@ -36,7 +36,7 @@ First download our example github_events dataset by running:
 
 Then, you can copy the data using psql:
 
-.. code-block:: sql
+.. code-block:: postgresql
 
     \COPY github_events FROM 'github_events-2015-01-01-0.csv' WITH (format CSV)
 
--- a/dist_tables/dml.rst
+++ b/dist_tables/dml.rst
@@ -3,7 +3,13 @@
 Ingesting, Modifying Data (DML)
 ###############################
 
-The following code snippets use the distributed tables example dataset, see :ref:`ddl`.
+The following code snippets use the distributed tables example, see :ref:`ddl`.
+We also use the github events dataset to illustrate the commands below. You can download that dataset by running:
+
+::
+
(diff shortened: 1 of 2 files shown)


dist_tables/dml.rst
```diff
@@ -3,7 +3,13 @@
 Ingesting, Modifying Data (DML)
 ###############################
 
-The following code snippets use the distributed tables example dataset, see :ref:`ddl`.
+The following code snippets use the distributed tables example, see :ref:`ddl`.
+We also use the github events dataset to illustrate the commands below. You can download that dataset by running:
```
begriffs: It looks like the bulk loading section is the only place that uses this dataset. What about moving the wget/gzip code block down there like

```rst
...you can directly use PostgreSQL’s \\COPY command. First download our example dataset:

.. code-block:: bash

  wget http://examples.citusdata.com/github_archive/github_events-2015-01-01-{0..5}.csv.gz
  gzip -d github_events-2015-01-01-*.gz

Then copy the data using psql:

.. code-block:: postgresql

  \COPY github_events FROM 'github_events-2015-01-01-0.csv' WITH (format CSV)
```

sumedhpathak: Makes sense. Making that change.

dist_tables/dml.rst
```diff
@@ -26,9 +26,17 @@ $$$$$$$$$$$$
...
-::
+.. code-block:: bash
+
+    wget http://examples.citusdata.com/github_archive/github_events-2015-01-01-{0..5}.csv.gz
+    gzip -d github_events-2015-01-01-*.gz
+
+
+Then, you can copy the data using psql:
+
+.. code-block:: postgresql
```
sumedhpathak: As a note, the change to a specific language hasn't seemed to make any difference to actual syntax highlighting.

begriffs: Hm, maybe the backslash confuses the highlighter. Oh well it was worth a try.

dist_tables/dml.rst
```diff
@@ -3,7 +3,7 @@
 Ingesting, Modifying Data (DML)
 ###############################
 
-The following code snippets use the distributed tables example dataset, see :ref:`ddl`.
+The following code snippets use the distributed tables example, see :ref:`ddl`.
```
begriffs: (Nitpick) After taking a fresh look at this, "distributed tables example" seems kind of vague. What about "...use the Github events schema defined in :ref:`ddl`"

begriffs:
Aside from the nitpick this LGTM.

dist_tables/dml.rst
sumedhpathak: Agreed. Changed.

PR #304, opened 2017-03-03
Document query propagation control (citus_tools)

### LINKED ISSUE #287: Document citus_tools

Some, but not all, statements propagate to workers. Citus provides utilities to do things on individual workers or shards. People really want to use the lower level functions to run SQL on all worker nodes etc.

### UDFs:
run_command_on_workers
run_command_on_placements
run_command_on_colocated_placements
run_command_on_shards

* Create UDFs on workers
```
SELECT run_command_on_workers($cmd$CREATE FUNCTION ..... $cmd$);
```

* Get size of shards (to determine how balanced the cluster is)
```
SELECT * from run_command_on_shards('<table_name>',
$cmd$SELECT pg_size_pretty(pg_total_relation_size('%I'))$cmd$, true)
--The third argument is the **parallel factor**.
--true runs the command across the worker shards in parallel.
--(This argument is optional and it is true by default)
```
* Total relation size:
```
SELECT pg_size_pretty(sum(result::int)) from
run_command_on_shards('<table_name>',
$cmd$SELECT pg_total_relation_size('%I')$cmd$, false);
--false runs the command across the worker shards in a sequential(non-parallel) way.
```
* Total indexes size:
```
SELECT pg_size_pretty(sum(result::int)) from run_command_on_shards('<table_name>',
$cmd$SELECT pg_indexes_size('%I')$cmd$);
```
* Get size of worker databases
```
SELECT * from run_command_on_workers($cmd$SELECT pg_size_pretty(pg_database_size('citus'))$cmd$);
```

* Create triggers on shard
https://gist.github.com/saicitus/74e6de56b77a26d22cc576004d8242b1 (Structure/idea is similar, the trigger creation in the link might not work out of the box because of some syntax issues, could you try fixing them)

### Example sql

```sql
select run_command_on_workers($cmd$ show work_mem; $cmd$);
```

### Corner cases, gotchas

* Can run only queries which return a single column and single row

begriffs:
Opening a PR for discussion on the progress so far. Do we want to add these functions to the UDF reference section as well?

Fixes #287


sumedhpathak:
This shouldn't belong in the SQL section. The commands *could* be used for SQL, but primarily are used for other utility activities (creating functions on workers, changing config settings etc). It's also used for some DDL like `create index`.

We had also discussed whether we want to make these functions public, or just create an unlinked page which can be used/sent at the discretion of Sales Engineering. @samay-sharma What are your thoughts here.

I think keeping them out of the UDF configuration section is also OK for now.

dist_tables/querying.rst
```diff
@@ -120,3 +120,126 @@ Citus employs a two stage optimizer when planning SQL queries. The first phase i
...
+  The :code:`run_command_on_workers` function and other manual propagation commands in this section can run only queries which return a single column and single row.
+
+Running on all Shards
+---------------------
+
+Each distributed table :code:`foo` has shards across the workers which are each tables with a naming convention of :code:`foo_n` where :code:`n` is a number. In order to run a query against a shard, the query must be run on its containing worker, and the shard name must be interpolated into the query string as an identifier (:code:`%I`).
+
+.. code-block:: postgresql
+
+  -- Get the size of table shards to determine how balanced the cluster
```
sumedhpathak: Are there better examples than table-size? We will have utility functions in 6.2 for distributed table size, so we may want to highlight a use-case other than that.

dist_tables/querying.rst
```diff
@@ -120,3 +120,126 @@ Citus employs a two stage optimizer when planning SQL queries. The first phase i
...
+Running on all Placements
+-------------------------
+
+When using Citus- rather than streaming-replication for :ref:`dealing_with_node_failures`, each shard has replicas stored as tables on other workers. Shards and their replicas are all called *placements*. (PostgreSQL streaming replication, on the other hand, works at the database level rather than the shard level, so the notion of shard replica placements is not applicable there.)
+
+Queries that update rows ought to be run on all placements rather than simply all shards. Ordinarily update queries go through the coordinator node which ensures they apply across placements. The functions in this section bypass the normal coordinator logic, so skipping any replicas in a manual update will leave the cluster inconsistent. Conversely, read-only queries should be run on shards rather than placements or they will return duplicate results.
+
+The following are equivalent:
+
+.. code-block:: postgres
```
sumedhpathak: This says `postgres` instead of `postgresql`.  Do you know which is the 'correct' one?

begriffs: I think they are synonyms, but agree that I should keep it consistent.

For reference these are the available [Pygments lexers](http://pygments.org/docs/lexers/) that enable highlighting. [This doc](https://build-me-the-docs-please.readthedocs.io/en/latest/Using_Sphinx/ShowingCodeExamplesInSphinx.html#pygments-lexers) confirms that Pygments is the underlying engine.

begriffs:
I moved this content to a `/extra/query_propagation.html` section that is not listed in the table of contents. Thus we can share the link with people but nobody will stumble on the section while browsing, nor will google know to index it.

begriffs:
[from our Slack conversation] this section should have a section about gotchas. I'll pull it back to the "please revise" column.

extra/query_propagation.rst
```diff
@@ -0,0 +1,128 @@
...
+---------------------
+
+Each distributed table :code:`foo` has shards across the workers which are each tables with a naming convention of :code:`foo_n` where :code:`n` is a number. In order to run a query against a shard, the query must be run on its containing worker, and the shard name must be interpolated into the query string as an identifier (:code:`%I`).
+
+.. code-block:: postgresql
+
+  -- Get the size of table shards to determine how balanced the cluster
+  -- is. The third argument controls whether the queries are run in
+  -- parallel and defaults to true
+  SELECT *
```
begriffs: @samay-sharma, @lfittl these examples of getting table and index size will be less impressive in 6.2 when we have UDFs to do it directly. Can you think of other relevant uses for `run_command_on_shards`?

lfittl: Maybe some other local table stats?

e.g. number of rows as estimated by pg_class.reltuples

extra/query_propagation.rst
```diff
@@ -0,0 +1,123 @@
...
+The following are equivalent:
+
+.. code-block:: postgresql
+
+  -- ordinary query going through the coordinator
+  UPDATE my_distributed_table
+     SET some_col = some_col + 1;
+
+  -- vs manually updating each placement
+  SELECT run_command_on_placements(
```
sumedhpathak: Is this the right example we have in mind for run_command_on_placements? Are there other examples we can think of which might be more useful?

extra/query_propagation.rst
```diff
@@ -0,0 +1,123 @@
...
+#############################
+
+When the user issues a query, the Citus coordinator partitions it into smaller query fragments where each query fragment can be run independently on a worker shard. This allows Citus to distribute each query across the cluster.
+
+However the way queries are partitioned into fragments (and which queries are propagated at all) varies by the type of query. In some advanced situations it is useful to manually control this behavior. Citus provides utility functions to propagate SQL to workers, shards, or placements.
+
+Running on all Workers
+----------------------
+
+The least granular level of execution is broadcasting a statement for execution on all workers. This is useful for viewing a properties of entire worker databases or creating UDFs uniformly throughout the cluster. For example:
```
sumedhpathak: Grammar: "viewing a properties of entire.." -> "viewing properties.."

extra/query_propagation.rst
```diff
@@ -0,0 +1,123 @@
...
+  SELECT run_command_on_workers($cmd$ SHOW work_mem; $cmd$);
+
+.. note::
+
+  The :code:`run_command_on_workers` function and other manual propagation commands in this section can run only queries which return a single column and single row.
+
+Running on all Shards
+---------------------
+
+Each distributed table :code:`foo` has shards across the workers which are each tables with a naming convention of :code:`foo_n` where :code:`n` is a number. In order to run a query against a shard, the query must be run on its containing worker, and the shard name must be interpolated into the query string as an identifier (:code:`%I`).
```
sumedhpathak: Can each opening paragraph follow a similar structure? The opening paragraph for `run_command_on_workers` is nice, as it talks about what the command does, and why its useful.

This paragraph deviates a bit from that as it first talks about what shards are and their naming. It gives more of a 'how', rather than 'what' and 'why'. A more simple description could be something like: "This functions takes the given command and runs it across all shards for the given distributed table. It could be useful for ...". We don't need to dive into a discussion of shard name extension.

extra/query_propagation.rst
```diff
@@ -0,0 +1,123 @@
...
+        SELECT reltuples
+          FROM pg_class
+         WHERE relname = '%I';
+      $cmd$
+    );
+
+Running on all Placements
+-------------------------
+
+When using Citus- rather than streaming-replication for :ref:`dealing_with_node_failures`, each shard has replicas stored as tables on other workers. Shards and their replicas are all called *placements*. (PostgreSQL streaming replication, on the other hand, works at the database level rather than the shard level, so the notion of shard replica placements is not applicable there.)
```
sumedhpathak: I don't think we need to discuss Statement vs Streaming replication here. We could just link to the reference section for shard-placements if someone wants to know what they are.

Could we follow the similar "what the function does", and "why it might be useful" pattern?
The function executes the given command on each placement/replica of every shard <link> of the given distributed table", "it can be useful for running data-modification commands, which need to be run on every replica to ensure consistency".

extra/query_propagation.rst
```diff
@@ -0,0 +1,123 @@
...
+
+Queries that update rows ought to be run on all placements rather than simply all shards. Ordinarily update queries go through the coordinator node which ensures they apply across placements. The functions in this section bypass the normal coordinator logic, so skipping any replicas in a manual update will leave the cluster inconsistent. Conversely, read-only queries should be run on shards rather than placements or they will return duplicate results.
```
sumedhpathak: I'd simplify this and merge with above to follow the pattern. We probably don't need to discuss the bypassing coordinator logic here. This probably is something we should discuss at the very top-level, as all of these commands bypass the coordinator logic.

extra/query_propagation.rst
```diff
@@ -0,0 +1,123 @@
...
+
+The following are equivalent:
+
+.. code-block:: postgresql
+
+  -- ordinary query going through the coordinator
```
sumedhpathak: We don't support this query as-is going through the co-ordinator. We support this via master_modify_multiple_shards. So let's take out this as an equivalent query.

extra/query_propagation.rst
```diff
@@ -0,0 +1,123 @@
...
+
+  -- vs manually updating each placement
+  SELECT run_command_on_placements(
+    'my_distributed_table',
+    $cmd$
+      UPDATE %I SET some_col = some_col + 1
+    $cmd$
+  );
+
+Whereas this next query leads to **inconsistency** for Citus-replication with replication factor greater than one:
```
sumedhpathak: Let's remove this example, as the initial paragraph should be enough to set this context. I'd rather not have query examples which we recommend not to run or could lead to inconsistency.

extra/query_propagation.rst
```diff
@@ -0,0 +1,123 @@
+Controlling Query Propagation
+#############################
+
+When the user issues a query, the Citus coordinator partitions it into smaller query fragments where each query fragment can be run independently on a worker shard. This allows Citus to distribute each query across the cluster.
+
+However the way queries are partitioned into fragments (and which queries are propagated at all) varies by the type of query. In some advanced situations it is useful to manually control this behavior. Citus provides utility functions to propagate SQL to workers, shards, or placements.
```
sumedhpathak: We can clarify here that this bypasses any coordinator logic, locking or other consistency checks. We should also note that these are in-place only as a last-resort to allow for statements which Citus otherwise does not run natively. If used incorrectly, it could lead to data inconsistency, deadlocks etc.

extra/query_propagation.rst
```diff
@@ -0,0 +1,112 @@
...
+
+For example, suppose a distributed table has an :code:`updated_at` field, and we want to "touch" all rows so that they are marked as updated at a certain time. An ordinary UPDATE statement on the coordinator requires a filter by the distribution column, but we can manually propagate the update across all shards and replicas:
+
+.. code-block:: postgresql
+
+  -- note we're using a hard-coded date rather than
+  -- a function such as "now()" because the query will
+  -- run at slightly different times on each replica
+
+  SELECT run_command_on_placements(
```
begriffs: Still perhaps a little contrived, but better than incrementing a column.

sumedhpathak:
Also, over the last week, I've received the question of being able to propagate worker commands at least 2-3 times. @begriffs @samay-sharma What do you think of also adding `run_command_on_workers` to our public docs?

begriffs:
I'm fine with documenting any of these functions publicly since we provide a warning about their hazards. But I'll defer to @samay-sharma's judgement.

extra/query_propagation.rst
begriffs: Also a better job for `master_modify_multiple_shards`...

dist_tables/dml.rst
```diff
@@ -162,3 +162,18 @@ Maximizing Write Performance
 ----------------------------
 
 Both INSERT and UPDATE/DELETE statements can be scaled up to around 50,000 queries per second on large machines. However, to achieve this rate, you will need to use many parallel, long-lived connections and consider how to deal with locking. For more information, you can consult the :ref:`scaling_data_ingestion` section of our documentation.
+
+Running a Query on All Workers
+------------------------------
+
+Broadcasting a statement for execution on all workers is useful for viewing properties of entire worker databases or creating UDFs uniformly throughout the cluster. For example:
```
sumedhpathak: As you note, we expect this UDF to be used primarily for cluster management. Could you move this section there?

PR #305, opened 2017-03-03
Link to cloud signup page rather than the console

begriffs:
It provides more context for new users


ozgune:
Looks good to me. Let's 🚀

PR #309, opened 2017-03-07
Remove docker subsections from table of contents

### LINKED ISSUE #306: Docker installation steps -- make headers uniform with other installation instructions
We forgot address a [small point](https://github.com/citusdata/citus_docs/pull/295#pullrequestreview-23608944) about Docker installation steps. We structured these steps' headers in a way that isn't consistent with the rest of our Single-Machine cluster.

In particular, when I click on the following page, I see one link for OS X, Red Hat, and Ubuntu. I see seven links for Docker.

https://docs.citusdata.com/en/v6.1/installation/development.html

Could we make the Docker steps uniform with those of other installation steps?

begriffs:
Use text bolding rather than true subsections to divide the docker page.

Fixes #306

### DIFF
--- a/installation/single_machine_docker.rst
+++ b/installation/single_machine_docker.rst
@@ -5,17 +5,14 @@ Docker
 
 This section describes setting up a Citus cluster on a single machine using docker-compose.
 
-1. Install Docker Engine and docker-compose
--------------------------------------------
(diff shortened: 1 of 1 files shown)


ozgune:
Looks good to me. Let's 🚀

PR #311, opened 2017-03-08
Use newer DDL in MX section

### LINKED ISSUE #264: Use the new API in Citus MX section
Currently, in Citus MX section of the docs, we use the old API `master_create_distributed_table` but we should use `create_distributed_table`, alongside some `start_metadata_sync_to_node` calls to workers.

begriffs:
Note that I didn't mention `start_metadata_sync_to_node` because this guide is for using Cloud on MX which takes care of that for the user. Also worth noting that when I first created the `events` table on the coordinator it took a while for it to appear in the `\d` list on the worker. For a while I saw only the shards (`events_102008` etc) and had to wait before I could proceed with the next step in this guide of inserting data.

Fixes #264

### DIFF
--- a/cloud/mx.rst
+++ b/cloud/mx.rst
@@ -55,8 +55,7 @@ To create a distributed events table with a JSONB column and a BRIN index, we ca
     primary key (device_id, event_id)
   );
   CREATE INDEX event_time_idx ON events USING BRIN (event_time);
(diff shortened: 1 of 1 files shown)


sumedhpathak:
@marcocitus Is this change sufficient for using the new API?

sumedhpathak:
Based on @aamederen 's comment in the issue, the change looks good to me. 🚢

PR #312, opened 2017-03-08
mmms does not require dist col in where clause

### LINKED ISSUE #307: Update description for master_modify_multiple_shards
The following [link](https://docs.citusdata.com/en/v6.1/reference/user_defined_functions.html?highlight=master_modify#master-modify-multiple-shards) link states:
```
As the function uses shard metadata to decide whether or not a shard needs to be updated,
it requires the WHERE clause in the query to be on the distribution column. 
```
This above is false.

**This UDF run UPDATES/DELETES on all the shards, it doesn't require WHERE clause on the distribution column. It is basically used for running the above DMLs cross-shard.**

**Example:** Suppose you want to delete older data from your table (which is distributed by say **tenant_id** column)
```
DELETE FROM <table_name> WHERE time_created<'xxxxxxxxx'
```
The above DELETE is cross shard because there is no **=** filter on tenant_id.
You can use master_modify_multiple_shards to run this query:
```
SELECT master_modify_multiple_shards('DELETE FROM <table_name> WHERE time_created<''xxxxxxxxx''');

### PR DESCRIPTION
**Pull request by begriffs**
Fixes #307

### DIFF
--- a/reference/user_defined_functions.rst
+++ b/reference/user_defined_functions.rst
@@ -257,7 +257,7 @@ The first example deletes all the shards for the github_events table since no de
 master_modify_multiple_shards
 $$$$$$$$$$$$$$$$$$$$$$$$$$$$$
 
-The master_modify_multiple_shards() function is used to run a query against all shards which match the criteria specified by the query. As the function uses shard metadata to decide whether or not a shard needs to be updated, it requires the WHERE clause in the query to be on the distribution column. Depending on the value of citus.multi_shard_commit_protocol, the commit can be done in one- or two-phases.
+The master_modify_multiple_shards() function is used to run a query against all shards which match the criteria specified by the query. Depending on the value of citus.multi_shard_commit_protocol, the commit can be done in one- or two-phases.
 
 Limitations:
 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: reference/user_defined_functions.rst [hunk 1]
```diff
@@ -257,7 +257,7 @@ The first example deletes all the shards for the github_events table since no de
 master_modify_multiple_shards
 $$$$$$$$$$$$$$$$$$$$$$$$$$$$$
 
-The master_modify_multiple_shards() function is used to run a query against all shards which match the criteria specified by the query. As the function uses shard metadata to decide whether or not a shard needs to be updated, it requires the WHERE clause in the query to be on the distribution column. Depending on the value of citus.multi_shard_commit_protocol, the commit can be done in one- or two-phases.
+The master_modify_multiple_shards() function is used to run a query against all shards which match the criteria specified by the query. Depending on the value of citus.multi_shard_commit_protocol, the commit can be done in one- or two-phases.
```
sumedhpathak: While we're modifying this description, could we update "used to run a query" to "used to run data modification statements which could span multiple shards"? We support `DELETE`, `UPDATE` and `TRUNCATE`, so we could possibly mention those as well. Otherwise looks good. Feel free to merge after making that change.

PR #313, opened 2017-03-08
Match Docker instructions in Citus README

begriffs:


installation/single_machine_docker.rst
```diff
@@ -16,18 +16,18 @@ This section describes setting up a Citus cluster on a single machine using dock
 
 .. code-block:: bash
 
-  curl -sSL https://get.docker.com/ | sudo bash
-  sudo usermod -aG docker $USER && newgrp docker
+  curl -sSL https://get.docker.com/ | sh
```
sumedhpathak: FWIW I get this right up front when I run this command:
``
modprobe: FATAL: Module aufs not found in directory /lib/modules/4.4.0-64-generic
``

Installation appears to work fine, but its a bit disconcerting to see that message.

begriffs: That is weird, which Linux distro did you try?

sumedhpathak: Ubuntu 16.04

begriffs: That scary fake error isn't good. Looks like it may be related to running Ubuntu on Amazon hardware specifically. The fix may be running `sudo modprob aufs` but I don't think they'll be updating the install script to do it. https://github.com/docker/docker/issues/14658

sumedhpathak: I can try again with Amazon Linux

jasonmp85: I've seen the same error in Ubuntu. I don't think it's meaningful.

jasonmp85: (Which is annoying since it's all **`FATAL`**)

PR #314, opened 2017-03-09
Warn about the need for timely ALTER EXTENSION

### LINKED ISSUE #281: Inform users of mandatory need to run ALTER EXTENSION UPDATE on minor/major updates.
## How does this work? (devs)

This comes from https://github.com/citusdata/citus/issues/907. Brian summarized the conversation there as:

* We add a fail-safe preventing you from running ALTER EXTENSION before you've restarted
* For minor releases we allow the user to restart and then ALTER at their leisure. (And restrict ourselves to the kinds of changes which allow such)
* For major releases we fail-safe and prevent the user from running any distributed queries until ALTER has been run.

begriffs:
Fixes #281

### DIFF
--- a/admin_guide/upgrading_citus.rst
+++ b/admin_guide/upgrading_citus.rst
@@ -63,14 +63,17 @@ Step 1. Update Citus Package
 Step 2. Apply Update in DB
 ~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-After installing the new package, run the extension upgrade script:
+After installing the new package and restarting the database, run the extension upgrade script.
 
 .. code-block:: bash
 
+  # you must restart PostgreSQL before running this
   psql -c 'ALTER EXTENSION citus UPDATE;'
 
-  psql -c '\dx'
   # you should see the newer Citus version in the list
+  psql -c '\dx'
+
+During a major version upgrade Citus will refuse to run distributed queries until ALTER EXTENSION is executed.
 
 Step 3. (upgrade from 5.x only) Add Co-Location Metadata
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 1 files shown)


admin_guide/upgrading_citus.rst
```diff
@@ -63,14 +63,17 @@ Step 1. Update Citus Package
...
 .. code-block:: bash
 
+  # you must restart PostgreSQL before running this
   psql -c 'ALTER EXTENSION citus UPDATE;'
 
-  psql -c '\dx'
   # you should see the newer Citus version in the list
+  psql -c '\dx'
+
+During a major version upgrade Citus will refuse to run distributed queries until ALTER EXTENSION is executed.
```
begriffs: I tried having this as a `.. note::` but it's close to another note and looked too busy. Could go back to that though if you think this point merits the extra attention.

sumedhpathak: I think it should be a note since it does merit attention. Also, its not just until ALTER EXTENSION, but even if you run `yum install` and don't *restart* the DB, we'll error out (albeit on new connections). So we should not that after `yum install` you should (a) restart and (b) alter-extension, and without both Citus will error out.
We can note that this is because object and function definitions are specific to a version, and on version mismatch can cause other issues.

begriffs: Noted

admin_guide/upgrading_citus.rst
```diff
@@ -63,14 +63,20 @@ Step 1. Update Citus Package
...
   psql -c 'ALTER EXTENSION citus UPDATE;'
 
-  psql -c '\dx'
   # you should see the newer Citus version in the list
+  psql -c '\dx'
+
+
+.. note::
+
+  During a major version upgrade, from the moment of yum installing a new version, Citus will refuse to run distributed queries until ALTER EXTENSION is executed. This is to protect your data, as Citus object and function definitions are specific to a version. After a yum install you should (a) restart and (b) run alter extension.
```
sumedhpathak: "Citus will refuse to run distributed queries until ALTER EXTENSION is executed." -> "Citus will refuse to run distributed queries until the *database is restarted* and ALTER EXTENSION is executed."

PR #317, opened 2017-03-11
Remove OS X installation steps

### LINKED ISSUE #300: Remove installation steps for Homebrew
Citus currently provides ten different ways for installation. We talked about simplifying our packaging in 6.1 so that we can focus our energy on doing fewer things (https://github.com/citusdata/citus/issues/1014).

We decided to remove instructions to install from Homebrew from our documentation and stop updating Homebrew with each release. We therefore should remove the following section from our docs: https://docs.citusdata.com/en/latest/installation/single_machine_osx.html

ozgune:
We're now guiding our Mac users to Docker installation steps. We therefore decided to remove installation steps for  Homebrew in #300. This change resolves #300.

Two questions on whether we'd like to make Docker more prominent with this change:

* Move Docker installation steps to the top, or
* Rename "Docker" to say "Docker (Mac or Linux)"

@begriffs / @sumedhpathak, what do you think?


begriffs:
I'd say both changes would be good: renaming and moving to the top. For those people comfortable with Docker it's probably the easiest way to get started.

sumedhpathak:
@begriffs Could you review the change? (And merge if looks good)

Issue #285, opened 2017-02-16
How to rename columns for distributed tables


begriffs:
### Link to relevant commits and regression tests if applicable

begriffs:
Associated with https://github.com/citusdata/citus/issues/633

begriffs:
Given the DDL feature support matrix in #320, what else will we need to cover for this issue?

begriffs:
I believe this issue is redundant given the PR about DDL propagation.

sumedhpathak:
@begriffs Looks like we have an example for column rename. This is currently pushed, but a bit premature given column rename is in 6.2. The other examples are OK. However, this issue can be closed.

PR #320, opened 2017-03-15
Document DDL propagation

### LINKED ISSUE #159: DDL propagation support (Create index, alter table)
This recently came up in our Slack channel. I'm copy/pasting the conversation below.

brian.freeman.slate [7:48 PM]  If I issue a `DROP INDEX` on the master node, does that propagate to all the workers? if not, how do I drop an index on a distributed table?

brian.freeman.slate [7:53 PM]  According to the docs: Currently `CREATE INDEX` and `ALTER TABLE` are the only DDL changes that Citus propagates automatically.
This is from http://docs.citusdata.com/en/v5.2/reference/configuration.html?highlight=index#citus-enable-ddl-propagation-boolean

scott [7:58 PM]  pretty sure drop table does too

robin.thomas [7:58 PM]  https://github.com/citusdata/citus/blob/v5.2.0/src/test/regress/sql/multi_index_statements.sql#L26

begriffs:
Fixes #159 

Note that `create index` is not listed as DDL in the PostgreSQL manual. Should I include `create index` along with my other additions in this section or in a new subsection of Querying Distributed Tables (SQL)?

### DIFF
--- a/dist_tables/ddl.rst
+++ b/dist_tables/ddl.rst
@@ -148,3 +148,33 @@ You can use the standard PostgreSQL DROP TABLE command to remove your distribute
 ::
 
     DROP TABLE github_events;
+
+DDL Propagation
+---------------
+
+Modifying Tables
+~~~~~~~~~~~~~~~~
+
+  Y Adding a Column
+  Y Removing a Column
+      Except the dist column
+  Y Removing a Constraint
+      N Check Constraints
+          And gives an unpolished error (ERROR:  42704: constraint "http_request_1min_check_102104" of relation "http_request_1min_102104" does not exist)
+        Not-Null Constraints
+        Unique Constraints
+      Y Primary Keys
+      Y Foreign Keys
+  Y Adding a Constraint
+      N Check Constraints
+      Y Not-Null Constraints
+      N Unique Constraints
+      N Primary Keys
+      Y Foreign Keys
+          Watch out for lack of existing uniqueness constraints (ERROR:  42830: there is no unique constraint matching given keys for referenced table "http_request_102008")
+          Uniqueness cannot be propagated after distribution
+      Exclusion Constraints
+  Y Changing a Column's Default Value
+  Y Changing a Column's Data Type (except for dist column)
+  N Renaming a Column
+  N Renaming a Table
--- a/dist_tables/ddl.rst
+++ b/dist_tables/ddl.rst
@@ -149,32 +149,69 @@ You can use the standard PostgreSQL DROP TABLE command to remove your distribute
 
     DROP TABLE github_events;
 
-DDL Propagation
----------------
+.. _ddl_prop_support:
 
 Modifying Tables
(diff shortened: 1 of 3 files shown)


dist_tables/ddl.rst
```diff
@@ -148,3 +148,70 @@ You can use the standard PostgreSQL DROP TABLE command to remove your distribute
...
++============+===================================+================================+
+| YES        | Adding a Column                   |                                |
++------------+-----------------------------------+--------------------------------+
+| YES        | Removing a Column                 | Except the distribution column |
++------------+-----------------------------------+--------------------------------+
+| YES        | Changing a Column's Default Value |                                |
++------------+-----------------------------------+--------------------------------+
+| YES        | Changing a Column's Data Type     | Except the distribution column |
++------------+-----------------------------------+--------------------------------+
+| NO         | Renaming a Column                 |                                |
```
begriffs: Note that I tested each of these commands on Citus 6.1-17 so some of these YES/NO settings will not be up to date for 6.2. For instance renaming a column will be supported in 6.2 (see #285).

One option is to cherry pick this feature support matrix into 6.1 and add another commit on master to update the entries for 6.2. Or just list the correct data for 6.2 and don't include this section in 6.1.

begriffs:
Includes renaming a column in the list so fixes #285

sumedhpathak:
Just recapping our notes here:

- Separate them into 3 larger sections: Adding/Modifying columns, Adding/Removing Constraints, Adding/Removing Indices.
- Add examples for each: Especially valuable for highlighting special cases like needing to alter before table creation.
- Add links to the relevant Postgres commands, so users can see full syntax.

dist_tables/ddl.rst
```diff
@@ -148,3 +148,135 @@ You can use the standard PostgreSQL DROP TABLE command to remove your distribute
...
+  CREATE INDEX clicked_at_idx ON clicks USING BRIN (clicked_at);
+
+  -- Removing an index
+
+  DROP INDEX clicked_at_idx;
+
+Manual Modification
+~~~~~~~~~~~~~~~~~~~
+
+If you attempt to alter a table and get an error such as "ERROR:  0A000: cannot create constraint," it means Citus does not yet support auto-propagation for the operation. In this case you can propagate the changes manually using this general four-step outline:
```
sumedhpathak: Is this the actual error we get? It seems specific to constraints, and I'd expect we should have a more proper Citus error around not being able to propagate the command. Could we remove this, and replace with something like "Currently other DDL commands are not auto-propagated, however there are different workarounds to manually propagate these to worker nodes" (or something like that).

sumedhpathak:
Another note: If we don't intend to backpatch this immediately, we can also include content for CREATE INDEX CONCURRENTLY, since that has been checked-in, and will be part of the 6.2 release. Else, we should open up a separate issue, and fix it in a different PR.

PR #321, opened 2017-03-15
Finding the dist col

### LINKED ISSUE #308: Get partition column of a distributed table
Basically if a user wants to get what is the distribution column of a distributed table.

Suppose a user forgets the name of the partition column and our error messages indicates say **add filter on partition column**, the user would want to see the partition column of a table.

A SQL statement which queries the meta-data table and gets the result.

### Example sql
```
SELECT column_to_column_name(logicalrelid, partkey) 
FROM pg_dist_partition where logicalrelid='<table_name>'::regclass;
```

begriffs:
Fixes #308

### DIFF
--- a/admin_guide/cluster_management.rst
+++ b/admin_guide/cluster_management.rst
@@ -176,3 +176,19 @@ The output contains the host and port of the worker database.
   │  102009 │          1 │           0 │ localhost │     5433 │           2 │
   └─────────┴────────────┴─────────────┴───────────┴──────────┴─────────────┘
 
+Finding the distribution column for a table
+-------------------------------------------
+
+Each distributed table in Citus has a "distribution column." For more information about what this is and how it works, see :ref:`Distributed Data Modeling <distributed_data_modeling>`. There are many situations where it is important to know which column it is. Some operations require joining or filtering on the distribution column, and you may encounter error messages with hints like, "add a filter to the distribution column."
+
+The :code:`pg_dist_*` tables on the coordinator node contain diverse metadata about the distributed database. In particular :code:`pg_dist_partition` holds information about the distribution column (formerly called *partition* column) for each table.
+
+.. code-block:: postgresql
+
+  -- a list of all tables and their distribution columns
+
+  SELECT column_to_column_name(logicalrelid, partkey)
+    FROM pg_dist_partition;
+
+  -- to search for a specific table add a condition like
+  -- WHERE logicalrelid='<table_name>'::regclass;
--- a/admin_guide/cluster_management.rst
+++ b/admin_guide/cluster_management.rst
@@ -187,14 +187,29 @@ The :code:`pg_dist_*` tables on the coordinator node contain diverse metadata ab
 
 .. code-block:: postgresql
 
+  -- create example table
+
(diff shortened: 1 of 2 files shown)


admin_guide/cluster_management.rst
```diff
@@ -176,3 +176,19 @@ The output contains the host and port of the worker database.
   │  102009 │          1 │           0 │ localhost │     5433 │           2 │
   └─────────┴────────────┴─────────────┴───────────┴──────────┴─────────────┘
 
+Finding the distribution column for a table
+-------------------------------------------
+
+Each distributed table in Citus has a "distribution column." For more information about what this is and how it works, see :ref:`Distributed Data Modeling <distributed_data_modeling>`. There are many situations where it is important to know which column it is. Some operations require joining or filtering on the distribution column, and you may encounter error messages with hints like, "add a filter to the distribution column."
+
+The :code:`pg_dist_*` tables on the coordinator node contain diverse metadata about the distributed database. In particular :code:`pg_dist_partition` holds information about the distribution column (formerly called *partition* column) for each table.
```
sumedhpathak: Maybe as a follow-up, we could add a line like "You can use a convenient utility function, which allows you to view the distribution column for a given table. An example query is..."

admin_guide/cluster_management.rst
```diff
@@ -176,3 +176,19 @@ The output contains the host and port of the worker database.
...
+
+.. code-block:: postgresql
+
+  -- a list of all tables and their distribution columns
+
+  SELECT column_to_column_name(logicalrelid, partkey)
```
sumedhpathak: Could you also add some example output along with the query?

sumedhpathak:
Apart from some minor feedback, should we also add this as an FAQ question, linking to this section?
"How do I find which column my table is partitioned on" or some such.

admin_guide/cluster_management.rst
```diff
@@ -176,3 +176,25 @@ The output contains the host and port of the worker database.
...
+
+Each distributed table in Citus has a "distribution column." For more information about what this is and how it works, see :ref:`Distributed Data Modeling <distributed_data_modeling>`. There are many situations where it is important to know which column it is. Some operations require joining or filtering on the distribution column, and you may encounter error messages with hints like, "add a filter to the distribution column."
+
+The :code:`pg_dist_*` tables on the coordinator node contain diverse metadata about the distributed database. In particular :code:`pg_dist_partition` holds information about the distribution column (formerly called *partition* column) for each table. You can use a convenient utility function to look up the distribution column name from the low-level details in the metadata. Here's an example and its output:
+
+.. code-block:: postgresql
+
+  -- get distribution column name for products table
+
+  SELECT column_to_column_name(logicalrelid, partkey)
```
sumedhpathak: Ok, this is a bit minor:

a) Maybe alias the SELECT value to something like "distribution_column"? e.g.:
``SELECT column_to_column_name(logicalrelid, partkey) as distribution_column``...

b) Maybe also show the sequence of steps leading up to this? 
So (1) CREATE TABLE products (shows the schema), (2) SELECT create_distributed_table('products', 'store_id'); (3) The column_to_column_name query.

PR #322, opened 2017-03-16
Distributing coordinator data

### LINKED ISSUE #302: Migrate data in create_distributed_table

To make migrating existing applications that use PostgreSQL less painful by allowing you to distribute existing tables that contain data.

Currently, migrating an existing application from regular PostgreSQL tables to distributed tables involves copying out the data out of a local table and then copying it back into a distributed table. This has a few substantial drawbacks:

- The user needs to turn off writes in the application
- The user needs to have a machine available with a high bandwidth connection to the Citus coordinator
- Copying data in an out can take a lot of time
- In some cases, the data needs to be temporarily written to a file which requires extra storage space
- In some cases, the distributed tables need to have a different name or schema

In cases where the current PostgreSQL server will act as the Citus coordinator, these problems can now be avoided by calling `create_distributed_table` directly on the existing PostgreSQL table. `create_distributed_table` now copies data from local table to shards. Once. Writes on the table are blocked while the data is migrated and pending writes are handled as distributed queries once `create_distributed_table` commits (or back to local if `create_distributed_table` fails). Reads can continue as normal and will become distributed queries once `create_distributed_table` commits.

We don't immediately delete the local data, which allows users to revert if something goes wrong during migration.

Note: `create_reference_table` also loads data automatically.

Scaling out an existing PostgreSQL server without significant downtime by installing Citus, adding worker nodes, and calling `create_distributed_table` on the existing tables.

Two-stage migrations in which the user first migrates from a regular PostgreSQL server to a Citus coordinator node (e.g. on Citus Cloud), and then calls `create_distributed_table` on tables to distribute the tables.

The main focus should be on describing migration from regular to distributed tables, this could be part of a larger document which also describes things like pg_dump and Amazon Data MIgration Service.

The basic functionality can be described in Creating Distributed Tables. There should be a new section on Data Migration.

## How does this work? (devs)

When you call create_distributed_table we effectively do the equivalent of a COPY on a distributed table, but instead of the rows coming from a file, we read them from the local table.

### Example sql

We can now show examples in which we load data prior to running `create_distributed_table`:

```
\COPY github_events FROM 'github_events-2015-01-01-0.csv' WITH (format CSV)

SELECT create_distributed_table('github_events', 'repo_id');
NOTICE:  Copying data from local table...
 create_distributed_table 
--------------------------
 
(1 row)
```

We also have some abilities that we did not have before:

```
CREATE TABLE series AS SELECT i FROM generate_series(1,1000000) i;
SELECT create_distributed_table('series', 'i');
NOTICE:  Copying data from local table...
 create_distributed_table 
--------------------------
 
(1 row)
```

### Corner cases, gotchas

The data is left on disk after `create_distributed_table` and its size will remain visible in `\d+`. 

In case of emergency (e.g. your application turns out to not work properly with distributed tables), this allows you to revert back to the local table. There is no clean way of doing this right now, but you could use the following commands:

```
SELECT master_drop_all_shards('series', 'pubic', 'series');
DELETE FROM pg_dist_partition WHERE logicalrelid = 'series'::regclass;
```

To free up the disk space after distributing data, you need to truncate the local table without truncating the distributed table. For this you can set `citus.enable_ddl_propagation` to off as of https://github.com/citusdata/citus/issues/1265 . It is recommend you do this in a transaction block to avoid accidentally truncating the distributed table (e.g. due to a reconnect, or using pgbouncer in transaction-pooling mode). 

```
BEGIN;
SET LOCAL citus.enable_ddl_propagation TO off;
TRUNCATE series;
-- SELECT * FROM series LIMIT 1; should still returns results
END;
```

When distributing a number of tables with foreign keys between them, it's best to drop the foreign keys before running `create_distributed_table` and recreating them after distributing the tables. They cannot always be enforced when one table is distributed and the other is not.

Foreign- or partitioned tables are not supported.

No

https://github.com/citusdata/citus/pull/1117
https://github.com/citusdata/citus/pull/1265

begriffs:
Fixes #302

### DIFF
--- a/dist_tables/ddl.rst
+++ b/dist_tables/ddl.rst
@@ -106,7 +106,7 @@ Writes on the table are blocked while the data is migrated, and pending writes a
 
   When distributing a number of tables with foreign keys between them, it's best to drop the foreign keys before running :code:`create_distributed_table` and recreating them after distributing the tables. Foreign keys cannot always be enforced when one table is distributed and the other is not.
 
-When migrating data from an external database, such as from Amazon RDS to Citus Cloud, first distribute the Citus table(s), then copy the data into the table.
+When migrating data from an external database, such as from Amazon RDS to Citus Cloud, first create the Citus distributed tables via :code:`create_distributed_table`, then copy the data into the table.
 
 .. _colocation_groups:
 
--- a/dist_tables/ddl.rst
+++ b/dist_tables/ddl.rst
@@ -106,23 +106,29 @@ Writes on the table are blocked while the data is migrated, and pending writes a
 
   When distributing a number of tables with foreign keys between them, it's best to drop the foreign keys before running :code:`create_distributed_table` and recreating them after distributing the tables. Foreign keys cannot always be enforced when one table is distributed and the other is not.
 
-When migrating data from an external database, such as from Amazon RDS to Citus Cloud, first transfer the data into the coordinator node, then distribute it throughout the cluster:
+When migrating data from an external database, such as from Amazon RDS to Citus Cloud, first distribute the Citus table(s), then copy the data into the table.
 
 .. code-block:: postgresql
 
   -- Suppose we have a CSV file to import, and have a destination table
   -- on the coordinator with the appropriate schema
 
-  -- First fill the table on the coordinator
-  \COPY github_events FROM 'github_events-2015-01-01-0.csv' WITH (format CSV)
-
-  -- Then distribute to the workers
+  -- First distribute the table to the workers
   SELECT create_distributed_table('github_events', 'repo_id');
+
+::
+
   NOTICE:  Copying data from local table...
    create_distributed_table
    --------------------------
 
    (1 row)
+
+.. code-block:: psql
+
+  -- Then fill the table. Rows will automatically distribute to workers.
+  \COPY github_events FROM 'github_events-2015-01-01-0.csv' WITH (format CSV)
+
 
 Both approaches work, but migrating from an external database to Citus has some drawbacks compared to turning an existing database directly into a Citus coordinator:
 
--- a/dist_tables/ddl.rst
+++ b/dist_tables/ddl.rst
@@ -83,6 +83,55 @@ If you have an existing distributed table which has a shard count of one, you ca
 
   SELECT upgrade_to_reference_table('table_name');
 
+Distributing Coordinator Data
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+If an existing PostgreSQL database is converted into the coordinator node for a Citus cluster, the data in its tables can be distributed efficiently and with minimal interruption to an application.
+
+The :code:`create_distributed_table` function described earlier works on both empty and non-empty tables, and for the latter automatically distributes table rows throughout the cluster. You will know if it does this by the presence of the message, "NOTICE:  Copying data from local table..." For example:
+
+.. code-block:: postgresql
+
+  CREATE TABLE series AS SELECT i FROM generate_series(1,1000000) i;
+  SELECT create_distributed_table('series', 'i');
+  NOTICE:  Copying data from local table...
+   create_distributed_table
+   --------------------------
+
+   (1 row)
+
+Writes on the table are blocked while the data is migrated, and pending writes are handled as distributed queries once the function commits. (If the function fails then the queries become local again.) Reads can continue as normal and will become distributed queries once the function commits.
+
+.. note::
+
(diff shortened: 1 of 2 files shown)


dist_tables/ddl.rst
```diff
@@ -83,6 +83,55 @@ If you have an existing distributed table which has a shard count of one, you ca
...
+
+   (1 row)
+
+Writes on the table are blocked while the data is migrated, and pending writes are handled as distributed queries once the function commits. (If the function fails then the queries become local again.) Reads can continue as normal and will become distributed queries once the function commits.
+
+.. note::
+
+  When distributing a number of tables with foreign keys between them, it's best to drop the foreign keys before running :code:`create_distributed_table` and recreating them after distributing the tables. Foreign keys cannot always be enforced when one table is distributed and the other is not.
+
+When migrating data from an external database, such as from Amazon RDS to Citus Cloud, first transfer the data into the coordinator node, then distribute it throughout the cluster:
```
begriffs: There's probably a more sophisticated way to export from RDS to another DB than dumping a CSV file. If so maybe I should mention it?

dist_tables/ddl.rst
```diff
@@ -83,6 +83,55 @@ If you have an existing distributed table which has a shard count of one, you ca
...
+
+  -- Then distribute to the workers
+  SELECT create_distributed_table('github_events', 'repo_id');
+  NOTICE:  Copying data from local table...
+   create_distributed_table
+   --------------------------
+
+   (1 row)
+
+Both approaches work, but migrating from an external database to Citus has some drawbacks compared to turning an existing database directly into a Citus coordinator:
```
begriffs: Am I painting too bleak a picture here?

dist_tables/ddl.rst
begriffs: Created an issue for more sophisticated RDS migration: #326

dist_tables/ddl.rst
```diff
@@ -83,6 +83,61 @@ If you have an existing distributed table which has a shard count of one, you ca
...
+
+Writes on the table are blocked while the data is migrated, and pending writes are handled as distributed queries once the function commits. (If the function fails then the queries become local again.) Reads can continue as normal and will become distributed queries once the function commits.
+
+.. note::
+
+  When distributing a number of tables with foreign keys between them, it's best to drop the foreign keys before running :code:`create_distributed_table` and recreating them after distributing the tables. Foreign keys cannot always be enforced when one table is distributed and the other is not.
+
+When migrating data from an external database, such as from Amazon RDS to Citus Cloud, first distribute the Citus table(s), then copy the data into the table.
+
+.. code-block:: postgresql
```
sumedhpathak: This full example isn't as relevant here anymore. We aren't talking about migration strategies in this section, and only discussing behavior of `create_distributed_table` when data exists in that table.

dist_tables/ddl.rst
```diff
@@ -83,6 +83,61 @@ If you have an existing distributed table which has a shard count of one, you ca
...
+
+   (1 row)
+
+.. code-block:: psql
+
+  -- Then fill the table. Rows will automatically distribute to workers.
+  \COPY github_events FROM 'github_events-2015-01-01-0.csv' WITH (format CSV)
+
+
+Both approaches work, but migrating from an external database to Citus has some drawbacks compared to turning an existing database directly into a Citus coordinator:
```
sumedhpathak: Same here. Let's remove the example, and also these drawbacks. We aren't discussing various migration optins here, and this also isn't the right location to discuss pros/cons of those options.

dist_tables/ddl.rst
```diff
@@ -83,6 +83,31 @@ If you have an existing distributed table which has a shard count of one, you ca
...
+
+   (1 row)
+
+Writes on the table are blocked while the data is migrated, and pending writes are handled as distributed queries once the function commits. (If the function fails then the queries become local again.) Reads can continue as normal and will become distributed queries once the function commits.
+
+.. note::
+
+  When distributing a number of tables with foreign keys between them, it's best to drop the foreign keys before running :code:`create_distributed_table` and recreating them after distributing the tables. Foreign keys cannot always be enforced when one table is distributed and the other is not.
+
+When migrating data from an external database, such as from Amazon RDS to Citus Cloud, first distribute the Citus table(s), then copy the data into the table.
```
sumedhpathak: Minor: Instead of "first distribute the Citus tables" -> "first create the Citus distributed tables via `create_distributed_table`"? (or something similar).

sumedhpathak:
I seem to recall us adding an FAQ entry for this. (e.g. "how do I migrate an existing Postgres table to Citus". @samay-sharma useful to add something like that?

begriffs:
I don't see the FAQ item in the "latest" version of the docs. Would you like me to add that prior to merging, or merge and create an issue (or merge and ignore)?

PR #330, opened 2017-03-31
Provide more info about HA in cloud docs

begriffs:
Re-uses the contents of our blog post from 2017-03-23. Separates the little security section into its own page.

Note that the animated gifs aren't transparent which looks a little off against the colored background of the docs. I investigated making them transparent but it caused old animation frames to show through newer ones. I could investigate other [frame disposal methods](https://www.imagemagick.org/Usage/anim_basics/#dispose) to make it work, or maybe whoever made the animation (@craigkerstiens?) could re-render it with transparency. Or maybe it just doesn't matter too much. 😀


sumedhpathak:
So based on our discussion, I think we want to maybe incorporate some of this information in the node addition/upgrade/side-grade PR. Some of this information isn't directly related to HA.

begriffs:
Does this mean we'll be merging this PR, or that the way this information applies to #337 makes the current PR invalid?

sumedhpathak:
@begriffs I think we should also revise this PR, and not merge as-is. Some of this content belongs in side-grades/upgrades.

begriffs:
@sumedhpathak I removed the part at the end about upgrades and will work on that in its own PR. Other than that, look good?

sumedhpathak:
@begriffs  Are you working on the PR for side-grades/upgrades? I'm curious to see how that PR will look, along with this one.

begriffs:
Yep, that one is #356

sumedhpathak:
Could you trigger a Heroku deployment for this one too? Thanks.

PR #335, opened 2017-04-06
Workaround for insert into serial distribution

begriffs:
FAQ entry for inserting values into a distributed table whose partition column has serial type.

Note that this FAQ entry should be removed when https://github.com/citusdata/citus/pull/1302 gets merged since these inserts will be supported out of the box.


sumedhpathak:
https://github.com/citusdata/citus/pull/1302 is now merged. :) We don't need this PR. Yay! @begriffs We can close.

PR #340, opened 2017-04-18
UDFs for inactive nodes

### LINKED ISSUE #282: Decouple replication of reference tables from master_add_node

The existing master_add_node function automatically replicates reference tables to the new node. However this is a costly operation. Citus users can now choose to call master_add_inactive_node() instead, and master_activate_node() at a later, more convenient, time.

https://docs.citusdata.com/en/latest/admin_guide/cluster_management.html#adding-a-worker

begriffs:
Fixes #282


sumedhpathak:
Looks good! 🚢

PR #341, opened 2017-04-18
Update FAQ

begriffs:
Noticed two things that seemed outdated when reviewing the faq. One was no mention of reference tables, and the other was a more real-time focus for SQL support.

### DIFF
--- a/faq/faq.rst
+++ b/faq/faq.rst
@@ -55,7 +55,7 @@ Other queries which, by contrast, combine data from multiple nodes, do not suppo
 * Window Functions
 * CTEs
 * Set operations
-* Transactional semantics
+* Transactional semantics for queries that span across multiple shards
 
 How do I choose the shard count when I hash-partition my data?
 --------------------------------------------------------------
--- a/faq/faq.rst
+++ b/faq/faq.rst
@@ -43,21 +43,21 @@ ingest workflows.
 Can I join distributed and non-distributed tables together in the same query?
 -----------------------------------------------------------------------------
 
-If you want to do joins between small dimension tables (regular Postgres tables) and large tables (distributed), then you could distribute the small tables by creating 1 shard and N (with N being the number of workers) replicas. Citus will then be able to push the join down to the worker nodes. If the local tables you are referring to are large, we generally recommend to distribute the larger tables to reap the benefits of sharding and parallelization which Citus offers. For a deeper discussion into the way Citus handles joins, please see our :ref:`joins` documentation.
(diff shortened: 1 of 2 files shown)


faq/faq.rst
```diff
@@ -43,21 +43,21 @@ ingest workflows.
...
 
-Are there any PostgreSQL features not supported by CitusDB?
------------------------------------------------------------
+Are there any PostgreSQL features not supported by Citus?
+---------------------------------------------------------
 
 Since Citus provides distributed functionality by extending PostgreSQL, it uses the standard PostgreSQL SQL constructs. This includes the support for wide range of data types (including semi-structured data types like jsonb, hstore), full text search, operators and functions, foreign data wrappers, etc.
 
-PostgreSQL has a wide SQL coverage; and Citus does not support that entire spectrum out of the box when querying distributed tables. Some constructs which aren't supported natively for distributed tables are:
+PostgreSQL has a wide SQL coverage; and Citus does not support that entire spectrum out of the box when querying distributed tables that span multiple worker nodes. Some constructs which aren't supported natively in this case are:
```
sumedhpathak: Could we invert this? We start by saying Citus doesn't support these out of the box, and then later note we do for multi-tenancy. Could we start by saying we support all these for multi-tenant apps? And maybe a small paragraph around limited support for real-time?

faq/faq.rst
```diff
@@ -43,21 +43,19 @@ ingest workflows.
...
-PostgreSQL has a wide SQL coverage; and Citus does not support that entire spectrum out of the box when querying distributed tables. Some constructs which aren't supported natively for distributed tables are:
+Other queries which, by contrast, combine data from multiple nodes, do not support the entire spectrum of PostgreSQL features. However they still enjoy broad SQL coverage, including semi-structured data types (like jsonb, hstore), full text search, operators, functions, and foreign data wrappers. Note that the following constructs aren't supported natively for cross-node queries:
 
 * Window Functions
 * CTEs
 * Set operations
-* Transactional semantics for queries that span across multiple shards
-
-It is important to note that you can still run all of those queries on regular PostgreSQL tables in the Citus cluster. As a result, you can address many use cases through a combination of rewriting the queries and/or adding some extensions. We are working on increasing the distributed SQL coverage for Citus to further simplify these queries. So, if you run into an unsupported construct, please contact us and we will do our best to help you out.
+* Transactional semantics
```
sumedhpathak: I'd retain the qualifier of queries that span multiple shards. If there are none of those, they'd work in real-time use-cases as well.

begriffs: This list is preceded by, "Note that the following constructs aren't supported natively for cross-node queries"

But I can add the qualifier back on that individual line as well just so people don't miss it.

PR #349, opened 2017-04-25
Citus supports concurrent index creation

### LINKED ISSUE #339: Citus 6.2 supports create index concurrently
#320 warned that `CREATE INDEX CONCURRENTLY` is not supported. But now it will be, so remove that warning on the master branch. (The original PR was merged in order to backport easily to 6.1)

begriffs:
Remove the warning

Fixes #339

### DIFF
--- a/dist_tables/ddl.rst
+++ b/dist_tables/ddl.rst
@@ -284,11 +284,11 @@ Not-null constraints can always be applied because they require no lookups betwe
 Adding/Removing Indices
 ~~~~~~~~~~~~~~~~~~~~~~~
 
-Citus supports adding and removing `indices <https://www.postgresql.org/docs/current/static/sql-createindex.html>`_, but not with the :code:`CONCURRENTLY` option.
+Citus supports adding and removing `indices <https://www.postgresql.org/docs/current/static/sql-createindex.html>`_:
 
 .. code-block:: postgresql
 
-  -- Adding an index (does not support CONCURRENTLY)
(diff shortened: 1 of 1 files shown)


dist_tables/ddl.rst
```diff
@@ -284,11 +284,11 @@ Not-null constraints can always be applied because they require no lookups betwe
 Adding/Removing Indices
 ~~~~~~~~~~~~~~~~~~~~~~~
 
-Citus supports adding and removing `indices <https://www.postgresql.org/docs/current/static/sql-createindex.html>`_, but not with the :code:`CONCURRENTLY` option.
+Citus supports adding and removing `indices <https://www.postgresql.org/docs/current/static/sql-createindex.html>`_:
 
 .. code-block:: postgresql
 
-  -- Adding an index (does not support CONCURRENTLY)
```
sumedhpathak: This looks good. Maybe we also add an explicit example for CONCURRENTLY? This has come up several times, and is valuable especially for large data sets. We can add an associated note that regular create-index holds a lock, but CONCURRENTLY won't. This is useful when we expect the index creation to take a while and for multi-tenant 'system-of-record' type databases where we don't want to take downtime.
See: https://www.postgresql.org/docs/9.1/static/sql-createindex.html#SQL-CREATEINDEX-CONCURRENTLY.

begriffs: Added CONCURRENTLY example.

PR #356, opened 2017-05-02
Cloud scaling

### LINKED ISSUE #337: Document node size upgrade and, node scale-out
Citus Cloud provides self-service scaling up and out of nodes. Each of these has slightly different behavior, but do accomplish a similar upgrade process. 

## Communication goals

Orientation and how-to as well

This likely lives it it's own section under cloud around upgrades. In the future we may want to add content about citus version and postgres upgrades as well. 

## How does this work? (devs)

In the case of an instance upgrade. We bring stand-by's of the new instance type online. Once those instances are fully caught up in transactions to the primary for each node we remove the IP from the primary, attach it to the stand-by, bring the stand-by out of recovery so it is now acting as the primary. We do this for each node as it's caught up. The impact is seconds for each node, though an overall cluster size increase could be minutes depending on when each is ready.

If you have a maintenance window set we run this operation during that time frame. If no maintenance window is set we run it as soon as the stand-bys are ready.

For scaling out. We bring new nodes online and set them up with Citus. You'll see within the Citus dashboard that new nodes are being added to the cluster. Once all new nodes are added you then need to run the rebalancer to introduce them to the cluster and have data redistributed to them. *Add more context on rebalancing and co-location*

https://www.citusdata.com/blog/2017/03/23/a-look-into-disaster-recovery-and-high-availability-and-how-they-work-with-postgres-on-citus-cloud/

begriffs:
Fixes #337


cloud/scaling.rst
```diff
@@ -0,0 +1,34 @@
...
+
+Clicking the node change link provides two sliders:
+
+.. image:: ../images/cloud-nodes-slider.png
+
+After you adjust the sliders and accept the changes, Citus Cloud begins applying the changes subject to the user-specified maintenance window:
+
+.. image:: ../images/cloud-maintenance-window.png
+
+Resizing nodes creates a PostgreSQL follower for each node, where the followers have the new hardware instance type. Creating follower nodes uses the same replication mechanism that powers Cloud's high-availability feature. Once the resize is initiated, it takes an average of forty minutes per hundred gigabytes of data for the follower's data to be fully synchronized. After the synchronization is complete, Cloud does a quick failover from the existing primary nodes to their followers which takes about one minute.
```
begriffs: @sumedhpathak this paragraph contains the same sentiment as was removed in https://github.com/citusdata/citus_docs/pull/330/commits/8f4890f43583460d2424c3047a2b69c3c33bd008 about reusing "HA plumbing" for instance upgrades. I can expand on it more if you like.

sumedhpathak:
@begriffs Could you trigger a Heroku deployment for this? Thanks.

sumedhpathak:
So there are two methods of scaling: Scaling up, or Scaling out. The distinction isn't clear from the text between the two. Can we have two separate sub-sections, one which shows scaling out (adding nodes, with screenshots and followed by rebalance-table-shards? And another one for Scaling up (increasing node size), again with screen-shots (assuming they are applicable)?

I think that would make this much clearer for me. Thanks!

begriffs:
See if you think this adjustment helps distinguish the two cases.

cloud/scaling.rst
```diff
@@ -0,0 +1,42 @@
...
+The first slider, **count**, scales out the cluster by adding new nodes. The second, **RAM**, scales it up by changing the amount of RAM on existing nodes.
+
+After you adjust the sliders and accept the changes, Citus Cloud begins applying the changes subject to the user-specified maintenance window:
+
+.. image:: ../images/cloud-maintenance-window.png
+
+Scaling Up (increasing node size)
+---------------------------------
+
+Resizing node RAM works by creating a PostgreSQL follower for each node, where the followers are provisioned with the desired amount of RAM. It takes an average of forty minutes per hundred gigabytes of data for the primary nodes' data to be fully synchronized on the followers. After the synchronization is complete, Citus Cloud does a quick failover from the existing primary nodes to their followers which takes about one minute. The creation and switchover process uses the same well-tested replication mechanism that powers Cloud's high-availability feature.
```
sumedhpathak: Could we show example screen-shots of an increase of the second slider here? (i.e. only bumping that up), and maybe the first slider below? I'm not sure if it adds to the clarity, but was wondering if they would.

cloud/scaling.rst
```diff
@@ -0,0 +1,42 @@
...
+
+Clicking the node change link provides two sliders:
+
+.. image:: ../images/cloud-nodes-slider.png
+
+The first slider, **count**, scales out the cluster by adding new nodes. The second, **RAM**, scales it up by changing the amount of RAM on existing nodes.
+
+After you adjust the sliders and accept the changes, Citus Cloud begins applying the changes subject to the user-specified maintenance window:
+
+.. image:: ../images/cloud-maintenance-window.png
```
sumedhpathak: I had some questions here:

- Does the maintenance window apply for scaling out as well? I can understand for scaling up, since it requires a switchover, but scaling out shouldn't require any downtime? 

- What considerations should we communicate to the user about the maintenance window? We discuss a bit below about time it takes to synchronize (for scaling up), but how much downtime should they expect? Is that downtime only during the switchover?

- Is there subsequent information we want to communicate? Do I get emails when its done? Does it show in the dashboard as "configuring" or something?

cloud/scaling.rst
sumedhpathak: "The creation and switchover process uses the same well-tested replication mechanism that powers Cloud's high-availability feature." Can we link to the HA section for cloud here?

cloud/scaling.rst
```diff
@@ -0,0 +1,42 @@
...
+
+Scaling Up (increasing node size)
+---------------------------------
+
+Resizing node RAM works by creating a PostgreSQL follower for each node, where the followers are provisioned with the desired amount of RAM. It takes an average of forty minutes per hundred gigabytes of data for the primary nodes' data to be fully synchronized on the followers. After the synchronization is complete, Citus Cloud does a quick failover from the existing primary nodes to their followers which takes about one minute. The creation and switchover process uses the same well-tested replication mechanism that powers Cloud's high-availability feature.
+
+Scaling Out (adding new nodes)
+------------------------------
+
+Node addition completes in five to ten minutes, which is faster than node resizing because the new nodes are created without data. To take advantage of the new nodes you still must adjust manually rebalance the shards, meaning move some shards from existing nodes to the new ones. Citus does not automatically do this on node creation because shard rebalancing takes locks that prevent inserts. The choice of when to run the shard rebalancer is left to the database administrator.
```
sumedhpathak: "Citus does not automatically do this on node creation because shard rebalancing takes locks that prevent inserts." 
This is much more aggressive than what the rebalancer actually does. It doesn't prevent all inserts, but only those for the shard that it is currently moving. It moves data a shard at a time (or a group of colocated shards), and inserts to other shards can continue normally.

sumedhpathak:
@begriffs Could you trigger a Heroku deployment? Thanks

begriffs:
Looking into it. The heroku deployment is in a weird state. I can try force pushing the top commit again.

cloud/scaling.rst
```diff
@@ -0,0 +1,58 @@
+Scaling
+#######
+
+Citus Cloud provides self-service scaling to deal with increased load. The web interface makes it easy to either add new worker nodes or increase existing nodes' memory capacity.
```
sumedhpathak: "or increase existing nodes' memory capacity". Not just memory right? Also CPU count/speed? Maybe just reword to "or scale up existing nodes"

cloud/scaling.rst
```diff
@@ -0,0 +1,58 @@
...
+
+For most cases either approach to scaling is fine and will improve performance. However there are times you may want to choose one over the other. If the cluster is reaching its disk limit then adding more nodes is the best choice. Alternately, if there is still a lot of headroom on disk but performance is suffering, then scaling node RAM is the way to go.
```
sumedhpathak: Same comment about "scaling node RAM" as above. Is there a reason you chose to only highlight RAM?

sumedhpathak: Ok, looks like the Sliders only denote RAM, but moving them also increases CPU count?

cloud/scaling.rst
```diff
@@ -0,0 +1,58 @@
...
+
+.. image:: ../images/cloud-formation-configuration.png
+
+Clicking the node change link provides two sliders:
+
+.. image:: ../images/cloud-nodes-slider.png
+
+The first slider, **count**, scales out the cluster by adding new nodes. The second, **RAM**, scales it up by changing the amount of RAM on existing nodes.
+
+For example, just drag the slider for node count:
```
sumedhpathak: I would have liked to see these different sliders in the two different sections, but I think it looks good for now.

cloud/scaling.rst
begriffs: If it's just changing an ec2 instance type then makes sense that CPU is increases as well. I'll confirm with the Cloud team.

PR #357, opened 2017-05-02
Add FAQ about RDS/Heroku

### LINKED ISSUE #355: Can I run Citus on Heroku or RDS?

It's a frequently asked question, so add to our FAQ.

The answer is something like:

> At this time Heroku and RDS do not support running Citus directly on top of Heroku Postgres or Amazon RDS Postgres. It would be up to them and their roadmaps on if/when they enable it. If you are looking for something similar, Citus Cloud is our database-as-a-service which we fully manage for you. It runs on top of AWS (like both RDS and Heroku Postgres) and should provide a very similar product experience, except with Citus.

begriffs:
Fixes #355


begriffs:
Cool I'll cherry pick it too.

PR #360, opened 2017-05-02
How to set auth methods for workers

### LINKED ISSUE #350: How to enable db security on the workers?

For convenience, our multi-node setup docs instruct people to use "trust" authentication in `pg_hba.conf` in for local network addresses. However many users have asked how to enable stronger security.

In a multi-tenant use case using row-level security to keep information strictly hidden between clients, stronger permissions are required to ensure that clients can connect to workers as only their own role.
(@lfittl this is only my guess, is it accurate?)

A new "Security" subsection under Cloud

## How does this work? (devs)

It is possible to require passwords on the workers. If you put worker nodes' password  into a .pgpass file on the master, the master will use the password to connect. Some queries (some joins we call repartition joins) also require that the workers connect to each other, for those queries to work you'll have to put the .pgpass file onto the workers as well.

https://www.postgresql.org/docs/current/static/libpq-pgpass.html

begriffs:
Fixes #350

I thought that the idea of worker permissions is relevant to more than just Cloud, so I added this to the general Cluster Administration section.

One thing that is lacking, however, is specific instructions for Cloud on how to ssh into the workers or scp the .pgpass and pg_hba.conf files into place. I actually don't know how to do this, perhaps I must ssh into the coordinator and from there into workers. Not sure. Is this something we should cover?

### DIFF
--- a/admin_guide/cluster_management.rst
+++ b/admin_guide/cluster_management.rst
@@ -165,6 +165,33 @@ Broadcasting a statement for execution on all workers is useful for viewing prop
 
 The :code:`run_command_on_workers` function can run only queries which return a single column and single row.
 
+Worker Security
+###############
+
+For your convenience getting started, our multi-node installation instructions direct you to set up the :code:`pg_hba.conf` on the workers with its `authentication method <https://www.postgresql.org/docs/current/static/auth-methods.html>`_ set to "trust" for local network connections. This is also the default configuration on Citus Cloud. It offers reasonable security, but more may be desired.
+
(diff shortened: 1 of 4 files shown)


lfittl:
@begriffs On Citus Cloud we setup pg_hba.conf for you - for creating new roles one simply uses the user interface. All users must login with a password over SSL, no exceptions / custom configuration allowed.

begriffs:
I'm moving this PR back into the Review column since there are no worker auth settings that can be changed in Cloud.

PR #365, opened 2017-05-04
Multi-tenant "reference architecture"

begriffs:
I put quotes on the title because we'll see where it ends up fitting best in the docs.

Opening this PR very early so we can all take a look.


ref_arch/multi_tenant.rst
```diff
@@ -0,0 +1,217 @@
...
+
+We will need to modify our schema, but there is one other caveat to note about distributed systems. Enforcing uniqueness or foreign key constraints in Citus requires that they include the distribution column. Our tables don't currently do that, for instance in the ads table we specify
+
+::
+
+  -- not efficiently enforceable
+
+  campaign_id bigint REFERENCES campaigns (id)
+
+This constraint includes only the campaign id, not the company (tenant) id. In order to verify the constraint Citus might have to consult multiple workers because it's not guaranteed in all situations that the ad in question is co-located with its campaign.
```
begriffs: Actually I don't feel I have explained the justification very well.

ref_arch/multi_tenant.rst
```diff
@@ -0,0 +1,393 @@
...
+We'll build the backend for an application that tracks online advertising performance and provides an analytics dashboard on top. It's a natural fit for a multi-tenant SaaS application because each company running ad campaigns is concerned with the performance of only its own ads. The queries' requests for data are partitioned per company.
+
+Before jumping ahead to all that, let's consider a simplified schema for this application, designed for use on a single-machine PostgreSQL database. The application must keep track of multiple companies, each of which runs advertising campaigns. Campaigns have many ads, and each ad has associated records of its clicks and impressions.
+
+Here is the initial schema for our application. We'll be updating it slightly later in this guide, so don't run it quite yet.
+
+::
+
+  -- Don't try executing this schema as-is, we need to make
+  -- a few changes later on to prepare it for Citus
```
samay-sharma: Could we align more with message discussed above?

ref_arch/multi_tenant.rst
```diff
@@ -0,0 +1,393 @@
...
+    id bigserial PRIMARY KEY,
+    ad_id bigint REFERENCES ads (id),
+    seen_at timestamp without time zone NOT NULL,
+    site_url text NOT NULL,
+    cost_per_impression_usd numeric(20,10),
+    user_ip inet NOT NULL,
+    user_data jsonb NOT NULL
+  );
+
+This schema supports querying the performance of ads and campaigns. It is designed for a single-machine database, and will require adjustment in a distributed environment. To see why, we must become familar with how Citus distributes data and executes queries.
```
samay-sharma: (Minor) :- Could we choose a word other than adjustment? maybe "modifications"?

ref_arch/multi_tenant.rst
```diff
@@ -0,0 +1,393 @@
...
+
+Applications connect to a certain PostgreSQL server in the Citus cluster called the *coordinator node.* The connection is established using an ordinary PostgreSQL `connection URI <https://www.postgresql.org/docs/current/static/libpq-connect.html#AEN45527>`_. However the actual data and processing is stored on and will happen in other machines called *worker nodes.*
+
+The coordinator examines each client query and determines what data the query needs, and which worker nodes have the data. The coordinator then splits the query into simplified *query fragments*, and sends them to worker nodes for processing. When the workers' results are ready, the coordinator puts them together into a final result and forwards it to the application.
```
samay-sharma: Could we alter this description to fir better with MT workloads? This is a description of the real-time executor i.e. splits into query fragments, co-ordinator puts together the results etc., however Multi tenancy mostly uses router.

Could we say something like "data is send to multiple nodes, and when queries come in, Citus figures out the right node to send the query to and fetches the results and returns them...."

ref_arch/multi_tenant.rst
```diff
@@ -0,0 +1,393 @@
...
+
+DIAGRAM: diagram of query execution
+
+Distributed Data in Citus
+-------------------------
+
+Using Citus effectively requires choosing the right pattern for distributing data and doing processing across workers. Citus runs fastest when the data distribution maximizes worker parallelism and minimizes network overhead for the application's most common queries. To minimize network overhead, related data items should be stored together on the same worker node. In multi-tenant applications this means that all data for a given tenant should be stored on the same worker. (Multiple tenants can be stored on the same worker for better hardware utilization, but no single tenant's data should span multiple workers.)
```
samay-sharma: Maybe remove the "maximizes worker parallelism".

ref_arch/multi_tenant.rst
```diff
@@ -0,0 +1,393 @@
...
+
+DIAGRAM: show id pulling from all workers, and company_id routed to one
+
+If we distribute by the campaign id, then campaign shards will be spread across multiple workers irrespective of company. Finding the top ten monthly campaign budgets for a company requires asking all workers for their local top ten and doing a final sort and filter on the coordinator. If we distribute by :code:`company_id`, on the other hand, then Citus can detect by the presence of :code:`WHERE company_id = 5` that all relevant information will be on a single worker. Citus can route the entire query to that worker for execution and pass the results through verbatim.
+
+The order/limit query slightly favors distribution by :code:`company_id`. However JOIN queries differ more dramatically.
+
+.. note::
+
+  In our normalized schema above, the ads table does not have a company_id column because it can retrieve that information through the campaigns table. If we want to distribute the ads table by company id however, we would need to denormalize the schema slightly and add that column. The query below assumes we have done this, and we'll talk more about this technique later.
```
samay-sharma: "If we want to distribute the ads table by company id however, we would need to denormalize the schema slightly and add that column." -> "To distribute the ads table by company id, we can add the company_id column to it"

ref_arch/multi_tenant.rst
```diff
@@ -0,0 +1,393 @@
...
+  AND campaigns.company_id = 5
+  AND campaigns.state = 'running'
+  GROUP BY campaigns.id, campaigns.name, campaigns.monthly_budget
+  ORDER BY total_impressions, total_clicks;
+
+DIAGRAM: show id repartitioning, and company_id routing
+
+For this query, distributing by campaign id is quite bad. Workers must use a lot of network traffic to pull related information together for the join, in a process called *repartitioning.* Routing the query for execution in a single worker avoids the overhead, and is possible when distributing by :code:`company_id`. The placement of related information together on a worker is called *co-location.*
+
+These queries indicate a general design pattern: distributing shards by tenant id (such as the company id) allows Citus to route queries to individual workers for efficient processing. This fits multi-tenant applications which join structured information together per-tenant.
```
samay-sharma: This is something I had written up in some other place which we can use to describe benefits of choosing tenant id as the sharding column. Feel free to re-use as required.

```
The key idea is to pick the item at the top of the hierarchy as
the sharding key. In B2B scenarios, this would typically be your
customer or business. This provides two benefits:-

* All tables would be related to the concept of tenant.
This would ensure all tables are colocated.

* Since the application is multi-tenant, all queries would be
scoped to a tenant. This means, they would go to a single node as
data for a tenant would be on a single node.

These properties dramatically reduce the cost associated
with distributed transactions, joins, and foreign key constraints.
It also allows Postgres to optimize a single-tenant query as it sees appropriate.
```

samay-sharma: Also, other than joins, I'd explain the benefits of sharding by tenant for transactions and foreign keys. We probably don't need to do diagrams etc but a description of why that's advisable and how it helps would be useful.

ref_arch/multi_tenant.rst
```diff
@@ -0,0 +1,393 @@
...
+
+::
+
+  -- not efficiently enforceable
+
+  campaign_id bigint REFERENCES campaigns (id)
+
+This constraint includes only the campaign id, not the company (tenant) id. In order to verify the constraint Citus might have to consult multiple workers because it's not guaranteed in all situations that the ad in question is co-located with its campaign.
+
+To guarantee that they are co-located, ad and campaign must both be distributed by company_id, and this column must appear in the foreign key. Similarly the primary key, implying uniqueness as it does, must be composite and include company_id.
```
samay-sharma: I think we should break this into pieces:-

(a) Explain why we need to have company id on all tables. This could follow from the description I have above or some modification of it.

(b) Explain changes to primary / foreign keys assuming schema changes in (a). i.e. If we don't include the tenant id in the primary key, Citus will need to check all worker nodes to enforce uniqueness on each index, which does not scale well. Adding tenant id to your primary keys allows Citus to ensure this at a shard level which can be completely enforced on the worker nodes.

ref_arch/multi_tenant.rst
```diff
@@ -0,0 +1,393 @@
...
+
+::
+
+  SELECT create_distributed_table('companies',   'id');
+  SELECT create_distributed_table('campaigns',   'company_id');
+  SELECT create_distributed_table('ads',         'company_id');
+  SELECT create_distributed_table('clicks',      'company_id');
+  SELECT create_distributed_table('impressions', 'company_id');
+
+This activates these tables for distributed storage and query execution. The next step is loading sample data into the cluster.
```
samay-sharma: Could we have a slightly more technical description of `create_distributed_table`? Above, we've explained what shards are and what worker nodes are. So, we could may be say something like below (with much better wording) :-

`create_distributed_table` informs Citus which tables are distributed i.e. all incoming queries to these tables will be planned for distributed execution. It also creates shards for these tables on the worker nodes so that incoming data can be routed to them.`

ref_arch/multi_tenant.rst
```diff
@@ -0,0 +1,393 @@
...
+    from 'ads.csv' with csv;
+  \copy clicks
+    from 'clicks.csv' with csv;
+  \copy impressions
+    from 'impressions.csv' with csv;
+
+Querying the Cluster
+--------------------
+
+Tenant applications in Citus can make ordinary queries, as long as the queries include the tenant id as a filter condition. For instance, suppose we are company id 5, how do we determine our total campaign budget?
```
samay-sharma: Change to something like :- "Multi-tenant applications in Citus can run any SQL queries, as long as the queries include the tenant id as a filter condition."

ref_arch/multi_tenant.rst
```diff
@@ -0,0 +1,393 @@
...
+Multi-tenant Applications
+#########################
+
+.. contents::
+
+*Estimated read time: 30 minutes*
+
+Many companies with web applications cater not to end users, but to other businesses with customers of their own. When many clients with similar needs are expected, it makes sense to run a single instance of the application to handle all the clients.
+
+A software-as-a-service (SaaS) provider, for example, can run one instance of its application on one instance of a database and provide web access to multiple customers. In such a scenario, each tenant's data is isolated and remains invisible to other tenants. This is efficient in two ways. First application improvements apply to all clients. Second, sharing a database between tenants uses hardware efficiently.
```
samay-sharma: "In such a scenario, each tenant's data is isolated and remains invisible to other tenants. " -> "In such a scenario, a single software system serves multiple customers, each with their own isolated view of their own data in that system"

samay-sharma:
@begriffs : I added a few higher level comments to the PR. Some of my comments were on sections you modified while addressing @sumedhpathak 's feedback. So, you might have to do a "show outdated" to look at all of them.

I've mostly provided feedback around "what to communicate / how to communicate" and haven't gone into minor verbiage or re-wording feedback yet. We can do that in a final pass once we're all aligned on the high level structure / content.

I'll let you make progress on the remaining sections and address feedback on these sections. Will plan to look at the PR again sometime next week. In the meanwhile, let me know if you have any questions.

ref_arch/multi_tenant.rst
```diff
@@ -0,0 +1,540 @@
...
+Querying the Cluster
+--------------------
+
+Multi-tenant applications in Citus can run any SQL queries, as long as the queries include the tenant id as a filter condition. For instance, consider a simple query to find the top campaigns with highest budget for company five.
+
+::
+
+  -- Top ten campaigns with highest budget for a company
+
+  SELECT name, cost_model, state, monthly_budget
```
samay-sharma: Should we show an explain plan for one of the queries to explain how it works under the covers (i.e. Citus detects that this tenant is on this node and forwards the request etc.)?

ref_arch/multi_tenant.rst
```diff
@@ -0,0 +1,540 @@
...
+  BEGIN;
+  DELETE from campaigns where id = 46 AND company_id = 5;
+  DELETE from ads where campaign_id = 46 AND company_id = 5;
+  COMMIT;
+
+When people scale applications with NoSQL databases they often miss the lack of transactions and joins. We already saw a join query when discussing distribution columns, but here's another to combine information from campaigns and ads.
+
+::
+
+  -- Total campaign budget vs expense this month
```
samay-sharma: Should we add a super complex query (with window functions / distinct / outer joins / where clause subqueries etc.) to inform the reader that any SQL will work?

ref_arch/multi_tenant.rst
```diff
@@ -0,0 +1,540 @@
...
+    from 'campaigns.csv' with csv;
+  \copy ads (id, company_id, campaign_id, name, image_url, target_url,
+             impressions_count, clicks_count, created_at, updated_at)
+    from 'ads.csv' with csv;
+  \copy clicks
+    from 'clicks.csv' with csv;
+  \copy impressions
+    from 'impressions.csv' with csv;
+
+Querying the Cluster
```
samay-sharma: I have two high level comments on this section.

(1) We should communicate the benefits of running these queries to app devs too i.e. what is the implication that all constructs work.

i.e. if you model your data in the above way, very minimal changes are needed to your application to scale as things work very easily. Citus will support all construct which your ORM generates. Then, we can also inform users about the MT gem which applies to ruby devs.

(2) Also, maybe add a concluding paragraph driving home the point that anything including inserts / updates / deletes, complex SQL, transactions, will work if you are scoped to a tenant. (As a takeway). We could then (maybe?) add the content I mentioned above.

What do you think?

ref_arch/multi_tenant.rst
```diff
@@ -0,0 +1,540 @@
...
+::
+
+  SELECT latlon
+    FROM geo_ips, clicks c
+   WHERE addrs >> c.user_ip
+     AND c.clicked_at > current_date - INTERVAL '1 day'
+     AND c.company_id = 5
+     AND c.ad_id = 456;
+
+In Citus we need to find a way to co-locate the :code:`geo_ips` table with clicks for every company. One way would be to add a :code:`company_id` column to :code:`geo_ips` and duplicate the data in the table for every company. This approach is not optimal because it introduces the burden of keeping the data synchronized between the companies if and when it changes. A more convenient way is by designating :code:`geo_ips` as a *reference table.*
```
samay-sharma: Maybe change first sentence to "To run this query efficiently in a distributed setup, we need to find a way to co-locate the :code:`geo_ips` table with clicks for every company. That way, no network traffic needs to be incurred during query time."

ref_arch/multi_tenant.rst
```diff
@@ -0,0 +1,540 @@
...
+
+Reference tables in Citus have exactly one shard, and it is replicated across all worker nodes. This co-locates the information for all tenants' queries. It does require reserving space for a copy of the data on all nodes, but automatically stays in sync during reference table updates. To make a reference table, call :ref:`create_reference_table <create_reference_table>` for a table on the coordinator node:
```
samay-sharma: What do you think about replacing "It does require reserving space for a copy of the data on all nodes, but automatically stays in sync during reference table updates."

with something like below (with better wording)

"Reference tables in Citus automatically stay in sync with modifications as Citus makes sure to forward them to all copies of the shard consistently."

ref_arch/multi_tenant.rst
```diff
@@ -0,0 +1,540 @@
...
+  -- Make synchronized copies of geo_ips on all workers
+
+  SELECT create_reference_table('geo_ips');
+
+After doing this, the join query (presented earlier) on :code:`geo_ips` and :code:`clicks` will perform efficiently.
+
+Online Changes to the Schema
+----------------------------
+
+Citus propagates most DDL statements fron the coordinator node to the workers. This allows the database administrator to alter the database even during use and after tables are distributed in the cluster. Citus uses a two-phase commit protocol to make sure updates happen consistently.
```
samay-sharma: Could we add some context / explain the benefit this provides to users? Something like:

Another challenge with multi-tenant systems is ensuring that the schemas for all the tenants are always in sync and schema changes are consistently reflected to all the tenants.

In Citus, you can use standard postgres DDL commands to change the schema of your tables and Citus will propagate them from the coordinator node to the workers.....

i.e. make it more MT focus as opposed to syntax / SQL focus.

ref_arch/multi_tenant.rst
```diff
@@ -0,0 +1,540 @@
...
+  SELECT
+    user_data->>'browser' AS browser,
+    count(*) AS count
+  FROM clicks
+  WHERE company_id = 5
+  GROUP BY user_data->>'browser'
+  ORDER BY count DESC
+  LIMIT 10;
+
+The database administrator can even create a `partial index <https://www.postgresql.org/docs/current/static/indexes-partial.html>`_ to improve speed for an individual tenant's query patterns. Here is one to improve company 5's filters for browser information:
```
samay-sharma: In addition to partial indexes, could we also talk about GIN indexes and the benefits they provide?

ref_arch/multi_tenant.rst
```diff
@@ -0,0 +1,540 @@
...
+
+.. note::
+
+  Don't forget that even when this process finishes there is more to do! The new node will be available in the system, but at this point no tenants are stored on it so **Citus will not yet run any queries there**.
+
+Node addition takes around five minutes. Refresh the browser until the change-in-progress message disappears. Next select the "Nodes" tab in the Cloud Console. You should see three nodes listed. Notice how the new node has no data on it (data size = 0 bytes):
+
+.. image:: ../images/cloud-node-stats.png
+
+To bring the node into play we can ask Citus to rebalance the shards. This operation moves shards between the currently active nodes to attempt to equalize the amount of data on each. Rebalancing preserves :ref:`colocation`, which means we can tell Citus to rebalance the :code:`companies` table and it will take the hint and rebalance the other tables which are distributed by :code:`company_id`.
```
samay-sharma: Do we want to talk about reads flowing through and writes being locked (on a per-shard group basis) for rebalance? Also, we can say that applications don't need to undergo downtime for this process to happen.

ref_arch/multi_tenant.rst
```diff
@@ -0,0 +1,540 @@
...
+  NOTICE:  00000: Moving shard [id] from [host:port] to [host:port] ...
+
+Refreshing the Nodes tab in the Cloud Console shows that the new node now contains data! It can now help processing requests for some of the tenants.
+
+Dealing with Big Tenants
+------------------------
+
+The previous section describes a general-purpose way to scale a cluster as the number of tenants increases. However there's another technique that becomes important when individual tenants get especially large compared to the others.
+
+As the number of tenants increases, the size of tenant data typically tends to follow a Zipfian distribution. This means there are a few very large tenants, and many smaller ones. Hosting a large tenant together with small ones on a single worker node can degrade the performance for all of them. Standard shard rebalancing won't prevent this mixing of tenants.
```
samay-sharma: "Standard shard rebalancing won't prevent this mixing of tenants."

Can we rephrase to something like "Standard shard rebalancing will rebalance the shards without paying specific attention to the large tenants. This may not always be sufficient to move large tenants to specific nodes." or something which indicates that shard rebalancer may sometimes solve this problem but is not built to do that?

ref_arch/multi_tenant.rst
begriffs: Added a little about GIN indices and jsonb operators.

ref_arch/multi_tenant.rst
begriffs: Didn't get this finished today but will get it tomorrow.

begriffs:
Got most of your comments covered. Have to still make a better repartition diagram, change the focus of "Querying the Cluster," come up with a complex query with window functions etc, add the explain plan for one of those queries, write the "where to go from here," and generate the missing demo data for one of the tables.

samay-sharma:
@begriffs : This reads much better, and I think the high level flow and content is all there. Great work! (I also like your re-partitioning diagram :))

I'm doing a second pass to see if there's any smaller level feedback and will put it back on your plate after that.

begriffs:
Awesome! I'll get the rest of those small issues taken care of. My biggest uncertainty is coming up with a realistic complex query. Got any ideas (even in English, and I can translate to SQL)?

Also @lfittl you mentioned you have a favorite diagram making tool, perhaps we could do a screenshare and you could show me your tricks? Then I'll be able to crank out nice diagrams faster in the future.

lfittl:
@begriffs Sure thing, lets schedule something later this week re: diagramming.

begriffs:
In case we need to re-generate the example datasets, this is how I did it:

* Start up a citus cluster in docker
* `export DATABASE_URL=...`
* Clone the citus-example-ad-analytics project
* `rake db:migrate`
* in psql:
   ```sql
   BEGIN;
   CREATE SEQUENCE clicky;
   ALTER TABLE clicks ADD COLUMN id_new integer;
   SET LOCAL citus.enable_ddl_propagation TO off;
   ALTER TABLE clicks ALTER COLUMN id_new SET DEFAULT nextval('clicky');
   END;
   
   BEGIN;
   CREATE SEQUENCE impressive;
   ALTER TABLE impressions ADD COLUMN id_new integer;
   SET LOCAL citus.enable_ddl_propagation TO off;
   ALTER TABLE impressions ALTER COLUMN id_new SET DEFAULT nextval('impressive');
   END;
   ```
* Change lib/tasks/fake_data.rake:
   ```ruby
   company_count_range    = 75..75
   campaign_count_range   = 6..10
   ad_count_range         = 6..10
   impression_count_range = 20..70
   click_count_range      = 1..20
   ```
* `rake test_data:bulk_load`
* Export from psql:
   ```sql
   \copy companies to '/tmp/companies.csv' DELIMITER ',' CSV;
   \copy ads to '/tmp/ads.csv' DELIMITER ',' CSV;
   \copy campaigns to '/tmp/campaigns.csv' DELIMITER ',' CSV;

   \copy clicks (
   	id_new, company_id, ad_id, clicked_at, site_url, cost_per_click_usd, user_ip, user_data)
   	to '/tmp/clicks.csv' DELIMITER ',' CSV;
   \copy impressions (
   	id_new, company_id, ad_id, seen_at, site_url, cost_per_impression_usd, user_ip, user_data)
   	to '/tmp/impressions.csv' DELIMITER ',' CSV;
   ```

ref_arch/multi_tenant.rst
```diff
@@ -0,0 +1,576 @@
...
+  BEGIN;
+  DELETE from campaigns where id = 46 AND company_id = 5;
+  DELETE from ads where campaign_id = 46 AND company_id = 5;
+  COMMIT;
+
+As a final demonstration of SQL support, this query includes a joins, grouping, aggregation and a window function, and it executes fine. The query ranks the ads in each campaign by the count of their impressions.
+
+::
+
+  SELECT a.campaign_id,
```
samay-sharma: This is a good start, I'll spend a few minutes today evening or tom morning trying to see if I can come up with something better.

ref_arch/multi_tenant.rst
```diff
@@ -0,0 +1,576 @@
...
+     AND i.ad_id      = a.id
+  GROUP BY a.campaign_id, a.id
+  ORDER BY a.campaign_id, n_impressions desc;
+
+In short when queries are scoped to a tenant then inserts, updates, deletes, complex SQL, and transactions all work as expected.
+
+Sharing Data Between Tenants
+----------------------------
+
+Up until now all tables have been distributed by :code:`company_id`, but sometimes there is data that can be shared by all tenants, and doesn't "belong" to any tenant in particular. For instance all companies using this example ad platform might want to get geographical information for their audience based on IP addresses. In a single machine database this could be accomplished by a lookup table for geo-ip, like the following. (A real table would probably use PostGIS but bear with the simplified example.)
```
samay-sharma: For instance , ...

ref_arch/multi_tenant.rst
```diff
@@ -0,0 +1,576 @@
...
+  SELECT latlon
+    FROM geo_ips, clicks c
+   WHERE addrs >> c.user_ip
+     AND c.clicked_at > current_date - INTERVAL '1 day'
+     AND c.company_id = 5
+     AND c.ad_id = 456;
+
+To run this query efficiently in a distributed setup, we need to find a way to co-locate the :code:`geo_ips` table with clicks for every company. That way, no network traffic need be incurred at query time. One way would be to add a :code:`company_id` column to :code:`geo_ips` and duplicate the data in the table for every company. This approach is not optimal because it introduces the burden of keeping the data synchronized between the companies if and when it changes. A more convenient way is by designating :code:`geo_ips` as a *reference table.*
+
+Reference tables are replicated across all worker nodes, and Citus automatically keeps them in sync during modifications. This co-locates the information for all tenants' queries. To make a reference table, call :ref:`create_reference_table <create_reference_table>` for a table on the coordinator node:
```
samay-sharma: could we add an "instead of create_distributed_table" to the end so that users don't end up running this after running create_distributed_table.

ref_arch/multi_tenant.rst
```diff
@@ -0,0 +1,576 @@
...
+    ADD COLUMN caption text;
+
+This updates all the workers as well. Once this command finishes, the Citus cluster will accept queries that read or write data in the new :code:`caption` column.
+
+For a fuller explanation of which DDL commands propagate through the cluster, see :ref:`ddl_prop_support`.
+
+Per-Tenant Customization
+========================
+
+Given that all tenants share a common schema and hardware infrastructure, how can we accommodate tenants which want to store information not needed by others? For example, one of the tenant applications using our advertising database may want to store tracking cookie information with clicks, whereas another tenant may care about browser agents. Traditionally databases using a shared schema approach for multi-tenancy have resorted to creating a fixed number of pre-allocated "custom" columns, or having external "extension tables." However PostgreSQL provides a much easier way with its unstructured column types, notably `JSONB <https://www.postgresql.org/docs/current/static/datatype-json.html>`_.
```
samay-sharma: (Minor) "extension tables." -> the double quotes need to be before the full stop.

ref_arch/multi_tenant.rst
```diff
@@ -0,0 +1,576 @@
...
+
+Multi-tenant databases should be designed for future scale as business grows or tenants want to store more data. Citus can scale out easily by adding new machines without having to make any changes or take application downtime.
+
+Being able to rebalance data in the Citus cluster allows you to grow your data size or number of customers and improve performance on demand. Adding new machines allows you to keep data in memory even when it is much larger than what a single machine can store.
+
+Also, if data increases for only a few large tenants, then you can isolate those particular tenants to separate nodes for better performance.
+
+.. note::
+
+  This section assumes you are using Citus Cloud and have created the schema and loaded data as described earlier. Additionally the Citus Cloud "Dev Plan" will not work because it does not allow scaling nodes; you must use a "Customized Plan."
```
samay-sharma: Should it be you must not use a "Customized Plan"? (dont know, just asking)

ref_arch/multi_tenant.rst
```diff
@@ -0,0 +1,576 @@
...
+
+.. image:: ../images/cloud-nodes-slider.png
+
+Drag the slider to increase node count by one, and click "Resize Formation." While the node is added the Cloud Console will display a message at the top of the screen:
+
+.. image:: ../images/cloud-change-progress.png
+
+.. note::
+
+  Don't forget that even when this process finishes there is more to do! The new node will be available in the system, but at this point no tenants are stored on it so **Citus will not yet run any queries there**.
```
samay-sharma: Maybe add something along the lines of "Below we'll explore how to move existing data to the new nodes"

ref_arch/multi_tenant.rst
```diff
@@ -0,0 +1,576 @@
...
+Refreshing the Nodes tab in the Cloud Console shows that the new node now contains data! It can now help processing requests for some of the tenants.
+
+Dealing with Big Tenants
+------------------------
+
+The previous section describes a general-purpose way to scale a cluster as the number of tenants increases. However there's another technique that becomes important when individual tenants get especially large compared to the others.
+
+As the number of tenants increases, the size of tenant data typically tends to follow a Zipfian distribution. This means there are a few very large tenants, and many smaller ones. Hosting a large tenant together with small ones on a single worker node can degrade the performance for all of them.
+
+Performing standard Citus shard rebalancing may improve the mixing of large and small tenants, but this would be accidental because it is not designed to do so. The rebalancer simply distributes shards to equalize storage usage on nodes, without examining the ratio of shards allocated to one tenant vs another within any given node.
```
samay-sharma: Slight rephrasing for the first sentence "Performing standard Citus shard rebalancing will improve overall performance but it may or may not improve the mixing of large and small tenants".

Also, the second sentence doesn't seem right to me. It says "without examining the ratio of shards allocated to one tenant", where as in MT setup a single tenant will be allocated only one shard. So, maybe we say something like "The rebalancer simply .... , without examining which tenants are allocated on each shard".

ref_arch/multi_tenant.rst
```diff
@@ -0,0 +1,576 @@
...
+
+  ┌─────────────────────────────┐
+  │ isolate_tenant_to_new_shard │
+  ├─────────────────────────────┤
+  │                      102240 │
+  └─────────────────────────────┘
+
+The optional :code:`CASCADE` parameter to :ref:`isolate_tenant_to_new_shard` makes a dedicated shard not only for the :code:`companies` table but for any other tables which are co-located with it. In our case that would be all the other tables except the reference table. If you recall, these tables are distributed by :code:`company_id` and are thus in the same co-location group. (Note that the shards created for the other tables each have their own shard id, they do not share id 102240.)
+
+Creating shards is only half the battle. The new shards -- one per table -- live on the worker nodes from which their data originated. For true hardware isolation we can move them to a separate node in the Citus cluster.
```
samay-sharma: Creating shards is only the first step. (instead of half the battle). I don't want to make tenant isolation sound like a battle 😄

samay-sharma:
@begriffs : My next batch of comments is out :slightly_smiling_face: . I still have the following things left to do:

* Click on all links to verify they work. 
* Review if we can or should add more links to our docs.
* See if I can come up with a representative complex query.
* Try all the instructions using docker / Citus cloud. @begriffs : Could you let me know which one you've been working heavily with and I can try the other one.
* Review diagrams once they are ready.

samay-sharma:
Also, just realized, we have rebalance and tenant isolation which are a part of enterprise edition and won't be available in docker. Do we want to add caveats to the doc indicating that?

ref_arch/multi_tenant.rst
begriffs: Mmmmonster query uery ery ry y❗️

ref_arch/multi_tenant.rst
begriffs: Are you sure? I'm more familiar with the period going inside the quotes. The [APA style guide](http://blog.apastyle.org/apastyle/2011/08/punctuating-around-quotation-marks.html) recommends that style for general usage as well.

ref_arch/multi_tenant.rst
begriffs: I know the dev plan didn't allow scaling, but whether there's a plan other than dev or customized which allows it I'm not sure.

ref_arch/multi_tenant.rst
begriffs: Lol. "We're deep into this devastating siege but the new shards still live on the nodes from which their data originated....Keep rolling your rock, Sisyphus."

ref_arch/multi_tenant.rst
begriffs: That is, I see a few "Predefined Plans" but they seem internal, not sure which ones are available to new users.

samay-sharma:
@begriffs : Regarding the diagrams, we have multiple diagrams representing the same concepts in different parts of the docs / in our presentations. Could we standardize on them and use the same ones everywhere? I've also sent an email to you today with a presentation we've used (And that also has diagrams for similar concepts). Reg. the repartitioning one, I think while resolving the comments we'll end up removing it from here but that's a useful one to keep in case we want to use it somewhere else (with better labelling).

ref_arch/multi_tenant.rst
```diff
@@ -0,0 +1,590 @@
...
+      CHECK (-90  <= latlon[0] AND latlon[0] <= 90 AND
+             -180 <= latlon[1] AND latlon[1] <= 180)
+  );
+  CREATE INDEX ON geo_ips USING gist (addrs inet_ops);
+
+Joining clicks with this table can produce, for example, the locations of everyone who clicked on ad 456.
+
+::
+
+  SELECT latlon
```
samay-sharma: I ran this query just after creating the reference table and received "ERROR: cannot plan queries that include both regular and partitioned relations". Could we make users run this query after they've run the create_reference_table() command to avoid them running into this error?

ref_arch/multi_tenant.rst
```diff
@@ -0,0 +1,590 @@
...
+
+  CREATE INDEX click_user_data_browser
+  ON clicks (user_data->>'browser')
+  WHERE company_id = 5;
+
+Additionally, PostgreSQL supports GIN indices on JSONB. Creating a GIN index on a JSONB column will create an index on every key and value within that JSON document. This speeds up a number of `JSONB operators <https://www.postgresql.org/docs/current/static/functions-json.html#FUNCTIONS-JSONB-OP-TABLE>`_ such as :code:`?`, :code:`?|`, and :code:`?&`.
+
+.. code-block:: postgresql
+
+  CREATE INDEX click_user_data
```
samay-sharma: This command fails for me.

ref_arch/multi_tenant.rst
```diff
@@ -0,0 +1,590 @@
...
+  WHERE company_id = 5
+  GROUP BY user_data->>'browser'
+  ORDER BY count DESC
+  LIMIT 10;
+
+The database administrator can even create a `partial index <https://www.postgresql.org/docs/current/static/indexes-partial.html>`_ to improve speed for an individual tenant's query patterns. Here is one to improve company 5's filters for browser information:
+
+.. code-block:: postgresql
+
+  CREATE INDEX click_user_data_browser
```
samay-sharma: This command fails for me.

ref_arch/multi_tenant.rst
```diff
@@ -0,0 +1,590 @@
...
+  WHERE company_id = 5;
+
+Additionally, PostgreSQL supports GIN indices on JSONB. Creating a GIN index on a JSONB column will create an index on every key and value within that JSON document. This speeds up a number of `JSONB operators <https://www.postgresql.org/docs/current/static/functions-json.html#FUNCTIONS-JSONB-OP-TABLE>`_ such as :code:`?`, :code:`?|`, and :code:`?&`.
+
+.. code-block:: postgresql
+
+  CREATE INDEX click_user_data
+  ON clicks GIN (user_data);
+
+  -- this allows queries like, "which clicks have
```
samay-sharma: Also, could we say this speeds up queries like "which clicks .... ". Because the query should be allowed irrespective of whether there is an index or not?

ref_arch/multi_tenant.rst
```diff
@@ -0,0 +1,590 @@
...
+
+  This section uses features available in Citus Cloud and Citus Enterprise, and will not work in the community edition.
+
+Multi-tenant databases should be designed for future scale as business grows or tenants want to store more data. Citus can scale out easily by adding new machines without having to make any changes or take application downtime.
+
+Being able to rebalance data in the Citus cluster allows you to grow your data size or number of customers and improve performance on demand. Adding new machines allows you to keep data in memory even when it is much larger than what a single machine can store.
+
+Also, if data increases for only a few large tenants, then you can isolate those particular tenants to separate nodes for better performance.
+
+.. note::
```
begriffs: 👍

ref_arch/multi_tenant.rst
begriffs: Totally, noticed the misleading wording myself before seeing your comment here.

ref_arch/multi_tenant.rst
```diff
@@ -0,0 +1,590 @@
...
+
+  # download and ingest datasets from the shell
+
+  for dataset in companies campaigns ads clicks impressions; do
+    curl -O https://examples.citusdata.com/mt_ref_arch/${dataset}.csv
+  done
+
+.. note::
+
+  **If you are using Docker,** you should use the :code:`docker cp` command to copy the files into the Docker container.
```
begriffs: Good point. On my own machine I ran the native psql and `\copy` moved the data inside to postgres, but the docker instructions that we provide have the user run psql inside the container (since it may not exist on the regular user machine).

begriffs:
OK @samay-sharma, I did a big rewrite based on our discussion. Changed it to be higher level and more encouraging, with links to the tech details. Wanted to show app devs how they can keep their data model and make it scale without much trouble.

See how you like it!

ref_arch/multi_tenant.rst
```diff
@@ -0,0 +1,511 @@
...
+This guide is framework-agnostic, so we'll point out some Citus features using SQL. Use your imagination for how these statements would be expressed in your language of choice.
+
+When a Citus sends a query to a single node for execution (called "pushing down" the query), then all SQL functionality is available. The node is an ordinary PostgreSQL server after all. When people scale applications with NoSQL databases they suffer the lack of transactions and joins. However per-tenant transactions work fine in Citus:
+
+::
+
+  -- transactionally remove campaign 46 and all its ads etc
+
+  BEGIN;
+  DELETE from campaigns where id = 46 AND company_id = 5;
```
begriffs: Actually might be cooler to use `ON DELETE CASCADE` in our foreign key, and pick another set of statements for this transaction.

ref_arch/multi_tenant.rst
begriffs: It's all good now, but the changed line is one below the one you marked so github didn't remove your comment.

ref_arch/multi_tenant.rst
begriffs: (changed one line below)

ref_arch/multi_tenant.rst
```diff
@@ -0,0 +1,590 @@
...
+
+As it executes, the command outputs notices of each shard it moves:
+
+.. code-block:: text
+
+  NOTICE:  00000: Moving shard [id] from [host:port] to [host:port] ...
+
+Refreshing the Nodes tab in the Cloud Console shows that the new node now contains data! It can now help processing requests for some of the tenants.
+
+Dealing with Big Tenants
```
begriffs: Incorporated both concerns.

ref_arch/multi_tenant.rst
```diff
@@ -0,0 +1,590 @@
...
+         impressions as i
+   WHERE a.company_id = 5
+     AND i.company_id = a.company_id
+     AND i.ad_id      = a.id
+  GROUP BY a.campaign_id, a.id
+  ORDER BY a.campaign_id, n_impressions desc;
+
+In short when queries are scoped to a tenant then inserts, updates, deletes, complex SQL, and transactions all work as expected.
+
+Sharing Data Between Tenants
```
begriffs: Done.

ref_arch/multi_tenant.rst
```diff
@@ -0,0 +1,511 @@
...
+Dealing with Big Tenants
+------------------------
+
+.. note::
+
+  This section uses features available in Citus Cloud and Citus Enterprise, and will not work in the community edition.
+
+The previous section describes a general-purpose way to scale a cluster as the number of tenants increases. However, users often have two concerns. The first is that their largest tenant won't be able to fit on a single node. The second is that hosting a large tenant together with small ones on a single worker node can degrade the performance for all of them.
+
+The first concern tends to be unwarranted. Investigating data from large SaaS sites reveals that as the number of tenants increases, the size of tenant data typically tends to follow a `Zipfian distribution <https://en.wikipedia.org/wiki/Zipf%27s_law>`_. For instance in a database of 10 tenants, the largest is predicted to account for 60% of the data. When there are 10k tenants, the largest will account for 2% of all data. Even at 10TB for 10k tenants, the largest tenant will require 200GB, which can still fit on a single node.
```
samay-sharma: Maybe slightly re-write:

Regarding the first question, investigating data from ......

We could also add a graph of Zipfs distribution here.

For instance, in a database of 100 tenants, the largest is predicted to account for 23% (please check this number, I've picked it up from https://www.wolframalpha.com/input/?i=frequency+of+largest+item+in+zipf+distribution+with+100+items) of the data. In a more realistic example for a large SaaS company, if there are 10k tenants, the largest will account for ~2% of the data. Even at 10TB of data, the largest tenant will require 200GB, which can pretty easily fit on a single node.

ref_arch/multi_tenant.rst
```diff
@@ -0,0 +1,511 @@
...
+
+There's truth to the second concern, but there's a solution as well. The true part is that standard Citus shard rebalancing will improve overall performance but it may or may not improve the mixing of large and small tenants. The rebalancer simply distributes shards to equalize storage usage on nodes, without examining which tenants are allocated on each shard.
```
samay-sharma: Slight re-write: "Another question is regarding performance when large and small tenants are on the same node. Standard shard rebalancing ....."

ref_arch/multi_tenant.rst
```diff
@@ -0,0 +1,511 @@
...
+
+To improve resource allocation and make guarantees of tenant QoS it is worthwhile to move large tenants to dedicated nodes.  Citus Enterprise Edition and Citus Cloud provide the tools to do this.
```
samay-sharma: I think we can just say Citus here (We've already brought up the enterprise edition / cloud thing above)

ref_arch/multi_tenant.rst
samay-sharma: Reason for re-write is: it is a valid concern for some customers and I don't want to dismiss it outright.

ref_arch/multi_tenant.rst
```diff
@@ -0,0 +1,511 @@
...
+
+Citus appears to applications as a single PostgreSQL database, but it internally routes queries to an adjustible number of physical servers (nodes) which can process requests in parallel. When configuring Citus, the database administrator specifies how the routing will happen.
+
+Multi-tenant applications have a nice property that we can take advantage of: queries usually always request information for one tenant at a time, not a mix of tenants. For instance, when a visitor is shopping at an online store runing on an ecommerce platform, the search results are just for that store; competitors' products are not mixed in.
+
+Because application queries are restricted to a single tenant, such as a store or company, one approach for making multi-tenant application queries fast is to store *all* data for a given tenant on the same node. This minimizes network overhead between the nodes. The joins, key constraints, and transactions can happen entirely within a node and do not require information from other nodes.
```
begriffs: Yeah it is a big section of raw text, but the existing diagrams use a different example than our ad analytics so it might be kind of confusing to drop them in there verbatim.

ref_arch/multi_tenant.rst
```diff
@@ -0,0 +1,511 @@
...
+::
+
+  ALTER TABLE ads
+    ADD COLUMN caption text;
+
+This updates all the workers as well. Once this command finishes, the Citus cluster will accept queries that read or write data in the new :code:`caption` column.
+
+For a fuller explanation of how DDL commands propagate through the cluster, see :ref:`ddl_prop_support`.
+
+When Data Differs Across Tenants
```
begriffs: (Done)

begriffs:
OK, got your suggestions in there. Just two notes -- I didn't verify the zipf numbers, and I still can add another statement to tee up the nosql comments as you mentioned. We can resume this next week, but should be in a good place to share with Ozgun and others.

use_case_guide/multi_tenant.rst
```diff
@@ -0,0 +1,517 @@
...
+
+However, a single relational database instance has traditionally had trouble scaling to the volume of data needed for a large multi-tenant application. Developers were forced to relinquish the benefits of the relational model when data exceeded the capacity of a single database node.
+
+Citus allows users to write multi-tenant applications as if they are connecting to a single PostgreSQL database, when in fact the database is a horizontally scalable cluster of machines. Client code requires minimal modifications and can continue to use full SQL capabilities.
+
+This guide takes a sample multi-tenant application and describes how to model it for scalability with Citus. Along the way we examine typical challenges for multi-tenant applications like per-tenant customization, isolating tenants from noisy neighbors, and scaling hardware to accommodate more data. PostgreSQL and Citus provide all the tools needed to handle these challenges, so let's get building.
```
samay-sharma: can we change per-tenant customization to differing data across tenants?

use_case_guide/multi_tenant.rst
```diff
@@ -0,0 +1,517 @@
...
+
+Multi-tenant applications have a nice property that we can take advantage of: queries usually always request information for one tenant at a time, not a mix of tenants. For instance, when a salesperson is searching prospect information in a CRM, the search results are specific to his employer; other businesses' leads and notes are not included.
+
+Because application queries are restricted to a single tenant, such as a store or company, one approach for making multi-tenant application queries fast is to store *all* data for a given tenant on the same node. This minimizes network overhead between the nodes and allows Citus to support all your application's joins, key constraints and transactions efficiently. They can happen entirely within a node and do not require information from other nodes. With this, you can scale across multiple nodes without having to totally re-write or re-architect your application.
```
samay-sharma: This minimizes network overhead between the nodes and allows Citus to support all your application's joins, key constraints and transactions efficiently. ~~They can happen entirely within a node and do not require information from other nodes.~~ With this, you can scale across multiple nodes without having to totally re-write or re-architect your application.

use_case_guide/multi_tenant.rst
```diff
@@ -0,0 +1,517 @@
...
+
+.. image:: ../images/zipf.png
+
+For instance, in a database of 100 tenants, the largest is predicted to account for 23% of the data. In a more realistic example for a large SaaS company, if there are 10k tenants, the largest will account for ~2% of the data. Even at 10TB of data, the largest tenant will require 200GB, which can pretty easily fit on a single node.
+
+Another question is regarding performance when large and small tenants are on the same node. Standard shard rebalancing will improve overall performance but it may or may not improve the mixing of large and small tenants. The rebalancer simply distributes shards to equalize storage usage on nodes, without examining which tenants are allocated on each shard.
+
+To improve resource allocation and make guarantees of tenant QoS it is worthwhile to move large tenants to dedicated nodes. Citus provides the tools to do this.
+
+In our case, let's imagine that our old friend company id=5 is very large. We can isolating the data for this tenant happens in two somewhat low-level steps. We'll present the commands here, and you can consult :ref:`tenant_isolation` for the theory behind them.
```
samay-sharma: We can ~~isolating~~ isolate the data for this tenant ~~happens~~ in two ~~somewhat low-level~~ steps.

use_case_guide/multi_tenant.rst
```diff
@@ -0,0 +1,517 @@
...
+
+First consolidate the tenant's data into a bundle (called a shard) suitable to move. The CASCADE option also applies this change to the rest of our tables distributed by :code:`company_id`.
```
samay-sharma: We're actually not consolidating data. We're just isolating the tenants data.

So, can we say "First, separate the tenant's data into its own shard".

use_case_guide/multi_tenant.rst
```diff
@@ -0,0 +1,517 @@
...
+    102240,
+    'source_host', source_port,
+    'dest_host', dest_port);
+
+You can confirm the shard movement by querying :ref:`pg_dist_shard_placement <placements>` again.
+
+Where to Go From Here
+---------------------
+
+Now that you know the basics of designing a schema for scalability in a multi-tenant application, you can continue further in our docs to put this knowledge to use. If you have an existing schema and want to migrate it for Citus, see :ref:`Multi-Tenant Transitioning <transitioning_mt>`. To adjust a front-end application, specifically Ruby on Rails, read :ref:`rails_migration`. Finally, try :ref:`Citus Cloud <cloud_overview>`, the easiest way to manage a Citus cluster, available with discounted developer plan pricing.
```
samay-sharma: With this, you now know how to use Citus to power your multi-tenant application for scalability. If you have an existing application and want to migrate it to Citus, see MT transitioning. To ...

use_case_guide/multi_tenant.rst
```diff
@@ -0,0 +1,517 @@
...
+
+Because application queries are restricted to a single tenant, such as a store or company, one approach for making multi-tenant application queries fast is to store *all* data for a given tenant on the same node. This minimizes network overhead between the nodes and allows Citus to support all your application's joins, key constraints and transactions efficiently. They can happen entirely within a node and do not require information from other nodes. With this, you can scale across multiple nodes without having to totally re-write or re-architect your application.
+
+We do this in Citus by making sure every table in our schema has a column to clearly mark which tenant owns which rows. In the ad analytics application the tenants are companies, so we must ensure all tables have a :code:`company_id` column.
+
+We can tell Citus to use this column to read and write rows to the same node when the rows are marked for the same company. In Citus' terminology :code:`company_id` will be the *distribution column*, which you can learn more about in :ref:`Distributed Data Modeling <distributed_data_modeling>`.
```
samay-sharma: Could we link here to distributing by tenant instead of the general distributed data modeling? (https://citus-doc-preview-pr-365.herokuapp.com/sharding/data_modeling.html#distributing-by-tenant-id)

use_case_guide/multi_tenant.rst
```diff
@@ -0,0 +1,517 @@
...
+
+  SELECT create_reference_table('geo_ips');
+
+Reference tables are replicated across all worker nodes, and Citus automatically keeps them in sync during modifications. Notice that we call :ref:`create_reference_table <create_reference_table>` rather than :code:`create_distributed_table`.
+
+Now joining clicks with this table can execute efficiently. We can ask, for example, the locations of everyone who clicked on ad 456.
+
+::
+
+  SELECT latlon
```
samay-sharma: Could we add some data to the reference table so that it returns results?

use_case_guide/multi_tenant.rst
samay-sharma: you can consult tenant isolation to learn more about them.

use_case_guide/multi_tenant.rst
```diff
@@ -0,0 +1,517 @@
...
+
+    for dataset in companies campaigns ads clicks impressions; do
+      docker cp ${dataset}.csv citus_master:.
+    done
+
+Being an extension of PostgreSQL, Citus supports bulk loading with the COPY command. Use it to ingest the data you downloaded, and make sure that you specify the correct file path if you downloaded the file to some other location.
+
+.. code-block:: psql
+
+  \copy companies from 'companies.csv' with csv;
```
samay-sharma: There's a transition here where we exit from psql and then enter psql again. Can we make sure to have this transition smoother i.e. indicate to user that he has to exit and then re-enter psql? Or, have him download the files first, before he runs the psql commands?

use_case_guide/multi_tenant.rst
```diff
@@ -0,0 +1,517 @@
...
+.. I would include a simple update statement or select statement to tee up the NoSQL point.
+
+A common pain point for users scaling applications with NoSQL databases is the lack of transactions and joins. However, transactions work as you'd expect them to in Citus:
+
+::
+
+  -- transactionally remove campaign 46 and all its ads etc
+
+  BEGIN;
+  DELETE from campaigns where id = 46 AND company_id = 5;
```
samay-sharma: This does not delete any rows. Could we pick a campaign id which exists for company_id = 5?

use_case_guide/multi_tenant.rst
```diff
@@ -0,0 +1,517 @@
...
+
+Given that all tenants share a common schema and hardware infrastructure, how can we accommodate tenants which want to store information not needed by others? For example, one of the tenant applications using our advertising database may want to store tracking cookie information with clicks, whereas another tenant may care about browser agents. Traditionally databases using a shared schema approach for multi-tenancy have resorted to creating a fixed number of pre-allocated "custom" columns, or having external "extension tables." However PostgreSQL provides a much easier way with its unstructured column types, notably `JSONB <https://www.postgresql.org/docs/current/static/datatype-json.html>`_.
+
+Notice that our schema already has a JSONB field in :code:`clicks` called :code:`user_data`. Each tenant can use it for flexible storage.
+
+Supposing company 5 uses the field to track the user agent ad clicks, they can later query to find which browsers click most often:
+
+.. code-block:: postgresql
+
+  SELECT
```
samay-sharma: Same thing with this, the browsers are not populated so it just shows a total count without any browser information. Could we write a query which returns some results or randomly add browser information to the data?

samay-sharma:
@begriffs : I added a few minor rephrasing suggestions and 3-4 comments on the data / queries not returning results. Other than that, looks good to me :). Along with addressing these, I think there are 3 things left now:

(a) Adding a diagram in the section for Scaling out relational model.
(b) Verifying Zipf distribution numbers.
(c) Adding the UPDATE command.

In the meanwhile, @ozgune, I think this is ready for your input. Could you give it a quick read and provide feedback?

begriffs:
OK, I'm learning more about Zipf. There's not actually one distribution called Zipfian, it's actually a parameterized family. When you see that a certain dataset has a Zipfian shape you need to find the parameter value that makes the theoretical distribution fit your data.

Your Wolfram calculations assumed a value of ρ=0.1:
```
P[X=1] for X~ZipfDistribution[100,0.1]

= 0.233753
```

If you pick ρ=1 instead then the answer jumps to 0.611627, so it's important we find the correct value.

use_case_guide/multi_tenant.rst
```diff
@@ -0,0 +1,538 @@
...
+
+    for dataset in companies campaigns ads clicks impressions geo_ips; do
+      docker cp ${dataset}.csv citus_master:.
+    done
+
+Being an extension of PostgreSQL, Citus supports bulk loading with the COPY command. Use it to ingest the data you downloaded, and make sure that you specify the correct file path if you downloaded the file to some other location. Back inside psql run this:
+
+.. code-block:: psql
+
+  \copy companies from 'companies.csv' with csv
```
sumedhpathak: Minor note. This seems to highlight only the first copy for me. The others appear grayed out.
screen shot 2017-06-30 at 11 50 27 am

samay-sharma:
@begriffs : I went through it and it looks great! I'm merging this into master and have opened #399 with some minor feedback which we can address as a follow up PR. Thanks for all the effort.

sumedhpathak:
🍾 🎆 Thanks @begriffs and @samay-sharma ! The content looks great!

Issue #366, opened 2017-05-08
6.2 release general checklist


begriffs:
- [x] s/6\\.1/6.2/g
- [ ] Verify / Update major-version upgrade steps (Joe)
    - ❌ 5.2.2->6.2 fails. To reproduce: https://gist.github.com/begriffs/935b97623d52c691246840e1d3e042c4
- [x] Verify / Update minor-version upgrade steps (Joe)
    - From 6.1-17 to 6.2-4
- [x] Check / Document new UDF's added in 6.2
- [x] Check / Document new config options added in 6.2
- [x] Update links from Readme and Citus web site to newer version of docs if required
- [x] Update links to new CloudFormation template (Sumedh)
- [x] Verify Installation steps with new packages (6.2-4)
    - Ubuntu 16.04
    - Ubuntu 14.04
    - Debian Jessie
    - Debian Wheezy
    - Amazon Linux 2017.03.0
    - CentOS 7.3
    - CentOS 6.9
    - RHEL 7.2
    - RHEL 6.5
    - ❌ Fedora 25 **No package citus62_96 available.**
    - Fedora 24
- [ ] Update all installation steps to use new package name (CloudFormation, Website instructions) (Sumedh)
- [x] Create new 6.2 release branch and set as default ver in RTD (last step)

begriffs:
@sumedhpathak here's a list of UDFs that are either not in the most recent docs at all, or are there but do not have entries in the reference section.

```sql
DROP FUNCTION master_drop_sequences(text[], text, bigint);
CREATE FUNCTION master_drop_sequences(sequence_names text[])

CREATE FUNCTION upgrade_to_reference_table(table_name regclass)


CREATE FUNCTION master_disable_node(nodename text, nodeport integer)

CREATE FUNCTION master_dist_local_group_cache_invalidate()
    IS 'register node cache invalidation for changed rows';


CREATE FUNCTION worker_apply_sequence_command(text)
    IS 'create a sequence which products globally unique values';

CREATE FUNCTION isolate_tenant_to_new_shard(table_name regclass, tenant_id "any", cascade_option text DEFAULT '')

CREATE FUNCTION stop_metadata_sync_to_node(nodename text, nodeport integer)

CREATE FUNCTION column_to_column_name(table_name regclass, column_var_text text)
    IS 'convert the textual Var representation to a column name';

CREATE FUNCTION lock_shard_resources(lock_mode int, shard_id bigint[])
[...]

PR #369, opened 2017-05-16
Bump version number to 6.2

begriffs:
Except CF template link

Affects #366


begriffs:
I made sure all the UDFs that we want to document are now included in the reference section. Sometimes there is a more in-depth discussion in existing places in the docs, in which case I provide links from each UDF in the reference to that existing documentation.

installation/multi_machine_aws.rst
```diff
@@ -41,7 +41,7 @@ Below, we explain in detail the steps required to setup a multi-node Citus clust
...
+<p>To begin, you can start a Citus cluster using CloudFormation by clicking <a href="https://console.aws.amazon.com/cloudformation/home?region=us-east-1#/stacks/new?stackName=Citus&templateURL=https:%2F%2Fcitus-deployment.s3.amazonaws.com%2Faws%2Fcitus6%2Fcloudformation%2Fcitus-6.2.0.json" onclick="trackOutboundLink('https://console.aws.amazon.com/cloudformation/home?region=us-east-1#/stacks/new?stackName=Citus&templateURL=https:%2F%2Fcitus-deployment.s3.amazonaws.com%2Faws%2Fcitus6%2Fcloudformation%2Fcitus-6.2.0.json'); return false;">here</a>. This will take you directly to the AWS CloudFormation console.</p>
```
sumedhpathak: The CF template itself hasn't been updated yet. @jasonmp85 @byucesoy I presume this will happen after packaging. We can then copy this over to S3 and update this link.

sumedhpathak:
@begriffs Could you trigger a Heroku deployment? Thanks!

reference/user_defined_functions.rst
```diff
@@ -389,6 +415,35 @@ Example
           7 |       7 | new-node |    12345 | default  | f           | t
     (1 row)
 
+master_disable_node
```
sumedhpathak: We didn't discuss this during prioritization, so I thought we wouldn't document this. If we do want to, could you review the function comment here: https://github.com/citusdata/citus/blob/master/src/backend/distributed/utils/node_metadata.c#L171? It has a bit more detail on what it does under the covers. An example is that it also removes reference-table placements from the node.

PR #373, opened 2017-05-17
Bump Citus documentation to 6.2.0

byucesoy:


begriffs:
Hey thanks Burak. I'm actually working on this already in #369 so I'm going to close this PR.

byucesoy:
I opened this PR because we have a checklist(https://github.com/citusdata/citus/issues/1415) for release items and this is one of the items. It only consist of changing cf template link and the version in conf.py. 

We are changing this two places as a part of packaging process for several releases. How should we arrange work division here?

@jasonmp85

begriffs:
Ah I see. I was using a checklist as well, https://github.com/citusdata/citus_docs/issues/366, which I guess  overlaps yours.

Actually we could re-open this PR for the CF template link change because I wasn't sure what the correct new value should be.

begriffs:
...or I can copy the CF template change into my own PR because it has a few other upgrade things in it as well.

begriffs:
(Copied the code)

PR #383, opened 2017-05-22
Remove outdated section on recovering from coordinator failure

lithp:
This section only applies to append-distributed tables. I think we've removed all other references to append-distributed tables from our docs so this should go as well.


ozgune:
@lithp -- Quick note. We still have a section in our docs on append distributed tables, but they aren't as visible as they used to be (they are in the Reference section).

lithp:
@ozgune would you prefer that we add a note stating this only applies to append-distributed tables?

sumedhpathak:
Looks good to me. 🚢

PR #386, opened 2017-05-22
Document citus.enable_version_checks

### LINKED ISSUE #374: Add node on citus.enable_version_check
We currently don't allow *any* queries, distributed or not, after a major version upgrade. This could be a problem if the upgrade itself errors out. The example we noticed was conflicting functions.

In this case, the user cannot take any remedial actions to resolve the situation. Thus, we have to document `citus.enable_version_check`, and note that in-case of upgrade errors, to disable the version check, allowing queries to run.

sumedhpathak:
Document enable_version_checks configuration parameter, and also add a note in the upgrade section to disable this if necessary. 
Fixes #374

### DIFF
--- a/admin_guide/upgrading_citus.rst
+++ b/admin_guide/upgrading_citus.rst
@@ -77,7 +77,7 @@ After installing the new package and restarting the database, run the extension 
 .. note::
 
   During a major version upgrade, from the moment of yum installing a new
-  version, Citus will refuse to run distributed queries until is restarted and
+  version, Citus will refuse to run distributed queries until the server is restarted and
   ALTER EXTENSION is executed. This is to protect your data, as Citus object and
   function definitions are specific to a version. After a yum install you
   should (a) restart and (b) run alter extension. In rare cases if you
--- a/admin_guide/upgrading_citus.rst
+++ b/admin_guide/upgrading_citus.rst
@@ -76,7 +76,15 @@ After installing the new package and restarting the database, run the extension 
 
 .. note::
 
-  During a major version upgrade, from the moment of yum installing a new version, Citus will refuse to run distributed queries until is restarted and ALTER EXTENSION is executed. This is to protect your data, as Citus object and function definitions are specific to a version. After a yum install you should (a) restart and (b) run alter extension.
+  During a major version upgrade, from the moment of yum installing a new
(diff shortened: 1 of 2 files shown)


reference/configuration.rst
```diff
@@ -7,14 +7,28 @@ There are various configuration parameters that affect the behaviour of Citus. T
...
+.. _enable_version_checks:
+
+citus.enable_version_checks (bool)
+$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$
+
+Upgrading Citus version requires a server restart (to pick up the new
+shared-library), as well as running an ALTER EXTENSION UPDATE command. The
+failure to execute both steps could potentially cause errors or crashes. Citus
+thus validates the version of the code and that of the extension match, and
+error's out if they don't.
```
begriffs: errors out

reference/configuration.rst
```diff
@@ -7,14 +7,28 @@ There are various configuration parameters that affect the behaviour of Citus. T
...
+
+This value defaults to true, and is effective on the coordinator. In rare cases
+this parameter may set to false, thus disabling the check, and might be suitable
```
begriffs: I was initially confused by this sentence. What about something like:

In rare cases, complex upgrade processes may require setting this parameter to false.

admin_guide/upgrading_citus.rst
```diff
@@ -76,7 +76,15 @@ After installing the new package and restarting the database, run the extension
 
 .. note::
 
-  During a major version upgrade, from the moment of yum installing a new version, Citus will refuse to run distributed queries until is restarted and ALTER EXTENSION is executed. This is to protect your data, as Citus object and function definitions are specific to a version. After a yum install you should (a) restart and (b) run alter extension.
+  During a major version upgrade, from the moment of yum installing a new
+  version, Citus will refuse to run distributed queries until is restarted and
```
begriffs: Just noticed that this PR preserves one of my own typos: "queries until is restarted." Want to fix it up while you're at it?

sumedhpathak:
Addressed feedback.

begriffs:
Looks good!

begriffs:
🚢

PR #387, opened 2017-05-23
Fix all sphinx warnings

begriffs:
Among other things, removes the warning about "toctree contains reference to nonexisting document."


begriffs:
@sumedhpathak I confirmed in the preview that this doesn't change the way anything looks except for enabling proper syntax highlighting in some blocks. I'm going to merge it because I don't like seeing warnings when I run the sphinx build locally.

PR #390, opened 2017-06-06
Multi-row insert not yet supported

### LINKED ISSUE #353: Can I do multi-row inserts?

Let people know they cannot yet do multi-row insert statements in Citus. But we do have insert into select and COPY as workarounds.

New FAQ entry.

begriffs:
Fixes #353 

We might also consider adding it to the FAQ, but it's getting pretty long. After a certain size I think an FAQ becomes less effective.


sumedhpathak:
I think we could hold off on this, since we'll be adding support in 7.0.

begriffs:
OK makes sense.

Issue #346, opened 2017-04-20
How to specify correct roles for Cloud users


begriffs:
## Why are we implementing it? (sales eng)

Organizations using Citus Cloud want to give different levels of access to different systems. We pass that responsibility to the Postgres roles underneath.

### What are the typical use cases?

The most common case is creating a user with read-only access for use by a web/reporting tool.

## Communication goals (e.g. detailed howto vs orientation)

Provide instructions for:
* add role through web interface
* psql in as the admin to customize role access

@lfittl are these the steps? Any other things to add about use cases?

### Good locations for content in docs structure

A new "Users and Security" section in Cloud

### Corner cases, gotchas

By default new users cannot access anything!

lfittl:
@begriffs Thats basically it - with roles I would also mention how PostgreSQL's grants work, and the following gotcha:

* Any GRANT thats on a specific distributed table will be propagated to the data nodes
* GRANTs that are system-wide (e.g. for all tables in a schema) need to be applied individually to every data node. You can use Citus tools to do so

It might be worth including a specific example of how to setup a read-only user, since that is the most common use case.

begriffs:
Thanks for the notes. Is it still true that new users cannot access anything? The placeholder text in this dialog box makes it look like the user will have read only access:

![screen shot 2017-04-20 at 12 40 34 pm](https://cloud.githubusercontent.com/assets/911911/25244517/b53c938c-25c6-11e7-9d05-1ec520d68cb7.png)

lfittl:
@begriffs The place holder text is wrong in that regard. Once we have docs up I'll likely just link to them and update that text to say `my_user_name`

begriffs:
(Closed by a PR)

PR #391, opened 2017-06-07
How to connect to Cloud

### LINKED ISSUE #344: How to connect to the Cloud database

Teach people the connection options for optimal security and speed.

A new subsection of Cloud about getting started, probably after a section about creating formations.

## How does this work? (devs)

To connect to a Cloud formation, users must use a PostgreSQL connection string. We'll begin by explaining the anatomy of the string, such as where the username, password, host, db etc are written. Show where to find the string in the Cloud console. Note too that we require SSL.

Users have the option to connect directly via port 5432 or through pgBouncer on 6432. The regular port imposes a hard limit of 300 connections to limit memory consumption, but pgBouncer allows any number of connections while effectively throttling the number of active connections into the db itself. @lfittl are there any pgBouncer limits I should know about, or other gotchas?

### Corner cases, gotchas

Some shells (zsh?) require escaping query params on the connection string such as `\?ssl_required=`.

begriffs:
Fixes #344


begriffs:
Just included a commit that fixes #346. It made sense to include it in this PR because the new section logically depends on the user connecting to the cluster as the earlier part of this PR explains.

cloud/index.rst
```diff
@@ -23,12 +23,125 @@ The key items you'll care about for each node:
...
+Users and Permissions
+#####################
+
+As we saw above, every new Citus Cloud formation includes a user account called :code:`citus`. This account is great for creating tables and other DDL, but it has too much power for certain applications.
+
+We'll want to create new roles for specialized purposes. For instance, a user with read-only access is perfect for a web/reporting tool. The Cloud console allows us to create a new user, and will set up a new password automatically. Go to the "Roles" tab and click "Create New Role."
+
+.. image:: ../images/cloud-roles-tab.png
+
+It pops up a dialog where we will fill in the role name, which we can call :code:`reports`.
```
begriffs: Maybe this is a weird name for a role...

cloud/index.rst
```diff
@@ -23,12 +23,125 @@ The key items you'll care about for each node:
 - Memory - The memory on each node varies based on the size of node you select
 - Cores - The cores on each node varies based on the size of node you select
 
-High Availability
-~~~~~~~~~~~~~~~~~
+Connecting to Citus Cloud
```
sumedhpathak: The overview section doesn't seem like the right section for this content. Looking at the Menu bar on the left, I wouldn't expect information about connecting and roles in Overview.

sumedhpathak: Should we maybe rename the current security section to "Connection and Security" and move both these sections in there? Any other options?

cloud/index.rst
```diff
@@ -23,12 +23,125 @@ The key items you'll care about for each node:
...
+.. image:: ../images/cloud-role-dialog.png
+
+After creating the role on a fresh formation, there will be three roles:
+
+.. code-block:: text
+
+  ┌─[ RECORD 1 ]────────────────────────────────────────────────────────────┐
+  │ Role name  │ citus                                                      │
+  │ Attributes │                                                            │
+  │ Member of  │ {reports}                                                  │
```
sumedhpathak: Do you know why Citus shows up as a member of reports? Is that a group or a role?

cloud/index.rst
```diff
@@ -23,12 +23,125 @@ The key items you'll care about for each node:
...
+
+  GRANT SELECT ON mytable TO reports;
+
+You can confirm the privileges by consulting the information schema:
+
+.. code-block:: postgresql
+
+  SELECT grantee, privilege_type
+    FROM information_schema.role_table_grants
+   WHERE table_name='mytable';
```
sumedhpathak: Minor: Spacing between the equality sign?

cloud/index.rst
```diff
@@ -23,12 +23,125 @@ The key items you'll care about for each node:
...
+Applications connect to Citus the same way they would PostgreSQL, using a `connection URI <https://www.postgresql.org/docs/current/static/libpq-connect.html#AEN45571>`_. This is a string which includes network and authentication information, and has the form:
 
-Note that your data is replicated to S3 with and without enabling high availability. This allows disaster recovery and reduces the risk of data loss. Although the data is safe either way, we suggest enabling high availability if you cannot tolerate up to one hour of downtime in the rare occurrence of an instance failure. We do not offer a SLA on uptime.
+::
+
+  postgresql://[user[:password]@][host][:port][/dbname][?param1=value1&...]
+
+The connection string for each Cloud Formation is provided on the Overview tab in Citus Console.
+
+.. image:: ../images/cloud-overview-1.png
```
sumedhpathak: What information does "created" have, that we've obfuscated?

cloud/index.rst
begriffs: It's a role that cloud creates for each formation, and uses in the connection string it provides. The role has enough permissions to create db objects, but it's not a superuser.

cloud/index.rst
begriffs: Oh, just the creation date and time. Didn't want to date the docs, wanted to keep them feeling fresh in case this section stays around unmodified for some time.

cloud/index.rst
begriffs: That sounds like a good plan.

cloud/index.rst
sumedhpathak: I see. The obfuscation looks a little off though. It appears we're trying to hide some privileged information in a docs screenshot. So I'd prefer to just leave the date as-is (even if it gets a bit stale).

cloud/index.rst
sumedhpathak: Wait, so "reports" is a role we create, and also something that Cloud creates for each formation? We created a role called "reports" in the previous step, and in this step it appears to be related to the "citus" role for some reason?

begriffs: Cloud creates the `citus` role automatically, then we ask Cloud to create `reports` through the GUI. As part of creating the new role the cloud console must internally run `grant reports to citus`. That's why citus is a member of reports. That membership allows a user who connected as the citus role in psql to switch into the reports role in the same connection by running `set role reports;`

cloud/index.rst
begriffs: I can even just erase the Created label and field entirely, then it won't look redacted, it just won't exist. Probably be easier to doctor the image like that than to regenerate it.

cloud/index.rst
sumedhpathak: Ok, got it. Thanks for the explanation.

begriffs:
K, I took care of the items you pointed out. I can also add a little more about the three roles if you think it might confuse people as-is.

PR #396, opened 2017-06-28
Remove explicit setting of replication factor in real-time dashboard reference architecture

### LINKED ISSUE #208: Locking problem in technical solution
Currently the insert script blocks `expire_old_request_data` from running. If the user leaves the insert script running in the background they might not understand why `expire_old_request_data` appears to hang forever.

sumedhpathak:
Setting it to 2 causes locking problems with the expiration script and data ingest. 
Fixes #208

### DIFF
--- a/use_case_guide/real_time_dashboards.rst
+++ b/use_case_guide/real_time_dashboards.rst
@@ -69,11 +69,7 @@ data across your cluster after adding new worker nodes.
 
 .. NOTE::
 
-  Citus Cloud uses `streaming replication
-  <https://www.postgresql.org/docs/current/static/warm-standby.html>`_ to achieve high
-  availability and thus maintaining shard replicas would be redundant. In a
-  production environment, you should set ``citus.shard_replication_factor`` to 2
-  or higher for fault tolerance.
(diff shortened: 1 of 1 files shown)


ref_arch/real_time_dashboards.rst
```diff
@@ -69,16 +67,13 @@ recommend :ref:`using 2-4x as many shards <faq_choose_shard_count>` as
...
-using the other replicas.
-
 .. NOTE::
 
-  In Citus Cloud you must use a replication factor of 1 (instead of the 2 used here). As
   Citus Cloud uses `streaming replication
   <https://www.postgresql.org/docs/current/static/warm-standby.html>`_ to achieve high
-  availability maintaining shard replicas would be redundant.
+  availability and thus maintaining shard replicas would be redundant. In a
+  production environment, you should set ``citus.shard_replication_factor`` to 2
```
begriffs: In one way of reading it, sounds like we're saying Citus Cloud is not a production environment. Maybe we could say, "in a self-hosted production environment" or "in a production environment where streaming replication is unavailable"

begriffs:
I tested the commands in the "Data Model" section in Docker and on Cloud and verified that the example dashboard query is able to run along with the ingestion script without blocking. I just added a note about one of sentences in this section and then looks good to me!

begriffs:
@sumedhpathak if you like I can adjust the sentence I pointed out and then merge this for you.

sumedhpathak:
@begriffs That would be great. Thanks!

PR #397, opened 2017-06-29
Update multi_tenant.rst

ozgune:
I'm proposing a few edits to the multi-tenant reference architecture document.

Probably, the biggest proposal is in the introduction section. The reason I'm proposing this change is that I read several new terms in the first two paragraphs and I was a bit confused by them (companies, end users, businesses, customers, clients, instance, tenants, and users). I wanted to reduce the number of terms.


begriffs:
Agree that this makes it simpler.

PR #400, opened 2017-07-05
Use case guide edits

### LINKED ISSUE #399: Improvements to MT use case guide
I'm merging the use case guide into master (https://github.com/citusdata/citus_docs/pull/365), but I wanted to track some minor feedback. @begriffs : We can look to fix it as a follow-up PR early next week.

1) We say "For SaaS applications, each tenant’s data can be isolated and kept invisible from other tenants. This is efficient in three ways. First application improvements apply to all clients. Second, sharing a database between tenants uses hardware efficiently. Last, it is much simpler to manage a single database for all tenants than a different database server for each tenant."

The benefits are of running a single instance of database to store all tenants data.

Can we make it. "For SaaS applications, each tenant’s data can be isolated and kept invisible from other tenants. However, running a single database instance to handle all your tenants is efficient in three ways. ....."

2) In preparing tables and ingesting data:
"Citus requires that primary and foreign key constraints include the distribution column. This requirement makes enforcing these constraints much more efficient in a distributed environment as only a single node has to be checked to guarantee them."

3) Highlight all COPY commands in the Try it yourself section.

4) In when data differs across tenants: "The database administrator can even create a partial index to improve speed for an individual tenant’s query patterns. Here is one to improve company 5’s filters for browser information:"

The query creates an index on the ismobile dimension and not on browser info. Can we fix the text there?

5) Now that we have the scaling on Citus Cloud section, could we update the "Scaling Hardware Resources" section to become shorter?

begriffs:
Fixes #399

@samay-sharma, my only reservation is in cutting out some of the scaling section. The flow feels damaged, because it vaguely tells the user to consult another section (in a way that seems easy to overlook), but then must resume the instructions for rebalancing because it is specific to the `companies` table. Maybe you have ideas to repair it.


use_case_guide/multi_tenant.rst
```diff
@@ -230,7 +230,7 @@ The next step is loading sample data into the cluster from the command line.
 
 Being an extension of PostgreSQL, Citus supports bulk loading with the COPY command. Use it to ingest the data you downloaded, and make sure that you specify the correct file path if you downloaded the file to some other location. Back inside psql run this:
 
-.. code-block:: psql
+.. code-block:: postgres
```
begriffs: This change is unfortunate because it removes highlighting from all lines. However I could not figure out how to make the psql lexer (aka [PostgresConsoleLexer](https://www.complang.tuwien.ac.at/doc/python-pygments/lexers.html#lexers-for-various-sql-dialects-and-related-interactive-sessions)) color all the `\copy` commands. The only way I could discover was to break the block into multiple little blocks, each with a single copy in them, but that looked bad.

use_case_guide/multi_tenant.rst
```diff
@@ -438,19 +438,11 @@ First log in to the `Citus Console <https://console.citusdata.com/>`_ and open t
...
-
-Drag the slider to increase node count by one, and click "Resize Formation." While the node is added the Cloud Console will display a message at the top of the screen:
-
-.. image:: ../images/cloud-change-progress.png
-
-.. note::
+To add nodes, click "Change node count and size." A slider will appear for both the count and size. For this section we'll be changing only the count.
 
-  Don't forget that even when this process finishes there is more to do! The new node will be available in the system, but at this point no tenants are stored on it so **Citus will not yet run any queries there**. Below we'll explore how to move existing data to the new nodes.
+Follow the instructions in :ref:`Cloud Scaling <cloud_scaling>` to add one more node to your cluster.
```
begriffs: I made this its own paragraph to cast it into the spotlight, but we may want to do more.

use_case_guide/multi_tenant.rst
```diff
@@ -230,7 +230,7 @@ The next step is loading sample data into the cluster from the command line.
 
 Being an extension of PostgreSQL, Citus supports bulk loading with the COPY command. Use it to ingest the data you downloaded, and make sure that you specify the correct file path if you downloaded the file to some other location. Back inside psql run this:
 
-.. code-block:: psql
+.. code-block:: text
```
begriffs: I could not figure out how to make the psql lexer (aka [PostgresConsoleLexer](https://www.complang.tuwien.ac.at/doc/python-pygments/lexers.html#lexers-for-various-sql-dialects-and-related-interactive-sessions)) color all the `\copy` commands. The only way I could discover was to break the block into multiple little blocks, each with a single copy in them, but that looked bad.

So giving up and removing all highlighting from this block.

PR #401, opened 2017-07-07
Remove note from the DDL section

ozgune:
We had a note in this section warning the user about setting their environment PATHs.

Is this note still relevant for the DDL section? Or should we move this warning to a different section?

### DIFF
--- a/dist_tables/ddl.rst
+++ b/dist_tables/ddl.rst
@@ -3,13 +3,6 @@
 Creating and Modifying Distributed Tables (DDL)
 ###############################################
 
-.. note::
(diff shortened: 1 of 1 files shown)


begriffs:
Those instructions do seem to be from a time when that content was in a more interactive section. Now that it's a reference section I agree that the PATH stuff is out of place.

Also the first code block includes `psql -h localhost -d postgres` which should probably be removed as well.

ozgune:
Good point. I removed `psql -h localhost -d postgres` and just updated the pull request.

@begriffs -- Do the changes look good to you?

begriffs:
Yep, looks good to me.

Issue #375, opened 2017-05-18
Django integration


begriffs:
## Why are we implementing it? (sales eng)

Many people have expressed interest in Citus for use as the backend for multi-tenant Django apps.

## Communication goals (e.g. detailed howto vs orientation)

Either documentation for @saicitus' python library, or if that is not ready then a cleaned-up version of his Django workaround notes. The goal is a howto that allows users to modify their Django apps.

samay-sharma:
We had a meeting today to discuss how to go about this. There are 2 pieces of content we need to draft:-

(a) Schema migration
(b) App / Query migration.

For (a), @saicitus would share a document on this issue which lists the changes we need to make based on customer experiences. @begriffs can then take that and turn it into a PR.

For (b), the idea is to document the library. Before we get started on that, @saicitus would write some tests for it and we'll do some internal testing. Based on the feedback from that, we can start drafting content for it.

saicitus:
@begriffs I have shared the schema-migration document with you. I have reviewed it recently and did some changes. You could take a look at it, proof-read and make suitable changes before publishing it in our docs.

PR #411, opened 2017-07-14
Django migration guide

begriffs:
Opening this PR early, just to make it visible.

### DIFF
--- a/migration/transitioning.rst
+++ b/migration/transitioning.rst
@@ -554,7 +554,7 @@ Finally, we'll establish a composite foreign key. :code:`./manage.py makemigrati
       ),
     ]
 
-Apply the migrations by running :code:`./`manage.py migrate`.
+Apply the migrations by running :code:`./manage.py migrate`.
 
 At this point the Django application models are ready to work with a Citus backend. You can continue by importing data to the new system and modifying controllers as necessary to deal with the model changes.
 
--- a/migration/transitioning.rst
+++ b/migration/transitioning.rst
@@ -513,6 +513,22 @@ Now make a custom migration for us to use to do low-level sql: :code:`./manage.p
 
 Make another custom migration to distribute the tables in Citus. This will not be undoable.
 
+.. code-block:: python
+
+  # as before, keep the structure of the migration the same, but use
+  # these operations
+  operations = [
+      migrations.RunSQL(
+        "SELECT create_distributed_table('appname_store','id')"
+      ),
+      migrations.RunSQL(
+        "SELECT create_distributed_table('appname_product','store_id')"
+      ),
+      migrations.RunSQL(
+        "SELECT create_distributed_table('appname_purchase','store_id')"
+      )
+  ]
+
 
 Real-Time Analytics Data Model
 ==============================
--- a/migration/transitioning.rst
+++ b/migration/transitioning.rst
@@ -225,7 +225,7 @@ the future:
 **1. Introduce a column for the tenant\_id on every record that belongs
 to a tenant**
 
-In order to scale out a multi-tenant model, its essential you can locate
+In order to scale out a multi-tenant model, it's essential you can locate
 all records that belong to a tenant quickly. The easiest way to achieve
 this is to simply add a ``tenant_id`` column (or “customer\_id” column,
 etc) on every object that belongs to a tenant, and backfilling your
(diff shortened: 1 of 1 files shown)


begriffs:
Should also note: fixes #375

migration/transitioning.rst
```diff
@@ -384,9 +387,225 @@ currently logged in customer - even though this is complex analytical
...
+      unique_together = ["id", "store"]  #
+
+  class Purchase(models.Model):
+    ordered_at = models.DateTimeField(default=timezone.now)
+    billing_address = models.TextField()
+    shipping_address = models.TextField()
+
+    product = models.ForeignKey(
+      Product,
+      db_constraint=False                # added
```
begriffs: Later we have a custom migration:

```sql
ALTER TABLE mtdjango_purchase
ADD CONSTRAINT mtdjango_purchase_product_fk
FOREIGN KEY (store_id, product_id)
REFERENCES mtdjango_product (store_id, id)
ON DELETE CASCADE;
```

I believe that declaring a models.ForeignKey relation in django between product and purchase allows succinct requests for a product's purchases in the ORM. I don't think that merely adding another uniqueness constraint would do that. (When the user uses your multi tenant library the query will include the tenant id as well.)

begriffs:
@saicitus this is ready for another look

sumedhpathak:
@saicitus Ping on this one.

migration/transitioning.rst
```diff
@@ -384,9 +391,233 @@ currently logged in customer - even though this is complex analytical
...
+    dependencies = [
+      ('appname', '<name of latest migration>')
+    ]
+
+    operations = [
+      # Django considers "id" the primary key of these tables, but
+      # the database mustn't, because the primary key will be composite
+      migrations.RunSQL(
+        "ALTER TABLE mtdjango_product DROP CONSTRAINT mtdjango_product_pkey;",
+        "ALTER TABLE mtdjango_product ADD CONSTRAINT mtdjango_product_pkey PRIMARY KEY (id)"
```
saicitus: I think it should be
```
"ALTER TABLE mtdjango_product ADD CONSTRAINT mtdjango_product_pkey PRIMARY KEY (store_id,id)"
```

migration/transitioning.rst
```diff
@@ -384,9 +391,233 @@ currently logged in customer - even though this is complex analytical
...
+      ),
+      migrations.RunSQL(
+        "ALTER TABLE mtdjango_purchase DROP CONSTRAINT mtdjango_purchase_pkey;",
+        "ALTER TABLE mtdjango_purchase ADD CONSTRAINT mtdjango_purchase_pkey PRIMARY KEY (id)"
```
saicitus: I think it should be:
```
"ALTER TABLE mtdjango_purchase ADD CONSTRAINT mtdjango_purchase_pkey PRIMARY KEY (store_id,id)"
```

migration/transitioning.rst
```diff
@@ -384,9 +391,233 @@ currently logged in customer - even though this is complex analytical
...
+    ]
+
+Apply the migrations by running :code:`./manage.py migrate`.
+
+At this point the Django application models are ready to work with a Citus backend. You can continue by importing data to the new system and modifying controllers as necessary to deal with the model changes.
+
+Updating the Django Application
+*******************************
+
+To simplify queries in the Django application, Citus has developed a Python library called `django-multitenant <https://github.com/citusdata/django-multitenant>`_. Include :code:`django-multitenant` in the :code:`requirements.txt` package file for your project, and then modify your models.
```
saicitus: Just mention that this library is in **Beta** because it has to be still tested well.

migration/transitioning.rst
saicitus: I think it should be:

"ALTER TABLE mtdjango_product ADD CONSTRAINT mtdjango_product_pkey PRIMARY KEY (store_id,id)"
(OR)
It could be removed(considering we have NOT NULL constraint for both store_id,id) because we already have unique_together = ["id", "store"]

migration/transitioning.rst
saicitus: I think it should be:

"ALTER TABLE mtdjango_purchase ADD CONSTRAINT mtdjango_purchase_pkey PRIMARY KEY (store_id,id)"
(OR)
It could be removed(considering we have NOT NULL constraint for both store_id,id) because we already have unique_together = ["id", "store"]

migration/transitioning.rst
begriffs: D'oh, good catch.

PR #414, opened 2017-07-19
Fix "Tutorial", and rename "Distributed Tables" section

### LINKED ISSUE #406: Rename "Distributed Tables" section
No other docs have something generically titled as "Distributed Tables".
The term "SQL Reference" seems commonly used.

sumedhpathak:
Renamed Distributed Tables to SQL Reference. I think that's a reasonable enough name, and couldn't come up with anything more descriptive. Also makes us think about adding information on SQL support.

Also fixed "Tutorial".

Fixes #406

### DIFF
--- a/index.rst
+++ b/index.rst
@@ -75,7 +75,7 @@ Welcome to the documentation for Citus 7.0! Citus horizontally scales PostgreSQL
    installation/production.rst
 
 .. toctree::
-   :caption: Tutorial
+   :caption: Tutorials
    :hidden:
 
    tutorials/multi-tenant-tutorial.rst
@@ -90,7 +90,7 @@ Welcome to the documentation for Citus 7.0! Citus horizontally scales PostgreSQL
 
 .. toctree::
    :maxdepth: 1
-   :caption: Distributed Tables
+   :caption: SQL Reference
    :hidden:
 
    dist_tables/ddl.rst
--- a/performance/scaling_data_ingestion.rst
+++ b/performance/scaling_data_ingestion.rst
@@ -92,7 +92,7 @@ The benefit of running INSERT or UPDATE commands, compared to issuing bulk COPY 
 .. _bulk_copy:
 
 Bulk Copy (250K - 2M/s)
-----------------------
+-----------------------
 
 Distributed tables support `COPY <http://www.postgresql.org/docs/current/static/sql-copy.html>`_ from the Citus coordinator for bulk ingestion, which can achieve much higher ingestion rates than INSERT statements.
 


begriffs:
"SQL Reference" is slightly confusing in that we have an existing "Reference" section that is full of SQL (UDFs etc).

Since we have a "Distributed Data Modeling" section, what about going along with that theme and making this one "Distributed Data Manipulation" or maybe "Distributed Queries"?

sumedhpathak:
Hmm. True. How about removing the existing SQL Reference section from the "reference" section, and possibly moving it into the current SQL section?

begriffs:
Yeah that would work, as long as you want those reference sections moved up that high in the docs.

samay-sharma:
@sumedhpathak : What about something like "Working with Citus" , "Working with Tables"? I think SQL reference it too generic.

samay-sharma:
@sumedhpathak , @begriffs : Any thoughts on above recommendation? I think we can get this PR merged soon once we agree on the naming.

sumedhpathak:
@samay-sharma Thanks for reminding me about this one. I don't like those, and in-fact like SQL reference. I think SQL reference is more specific right? To me it clearly signifies what to expect. I think the complication is that we have another 'reference' section. 

We previously had the "working with.." title, and I just think its very vague. I haven't seen any examples like that on other docs. 

My proposal is to move the SQL reference (and also the workarounds) from our current reference to this section. I think it makes sense for it to be here.

The current reference section is a mix of "Information schema" and "Functions and Operators" or some such. 

I'd much rather make a change like that. For now, what do you think about just merging the "Tutorial" change in a different commit?

samay-sharma:
@sumedhpathak : Hmm. One question I have there is if we want to move stuff like workarounds, or SQL language reference so high in the docs or move these creating distributed tables, DDLs, insert / copy etc to lower in the docs. I have a preference for the latter as I'd like to have other stuff featured more prominently than these. Also, if people follow the tutorials / use case guides, they learn enough about these functions to understand the basics. What do you think?

samay-sharma:
Also, while we're on the topic of renaming headers (:)), I noticed that we call the migration section "Transitioning to Citus". What do you think about putting the word "Migration" in there? We could call it just Migration or Migrating to Citus.

We'll be adding more content there in the upcoming weeks (Django and possible other stuff) and it would be good to call it Migration as opposed to Transitioning.

begriffs:
I rebased this so it's clean to merge.

What if we merge these changes and then start a google doc to discuss more ambitious menu reorganization? I know other sections like Cloud are starting to feel unfocused too.

sumedhpathak:
That works for me. Could you go ahead and merge? (And cherry-pick). Thanks!

PR #416, opened 2017-07-19
Independent menu scrolling

### LINKED ISSUE #405: Allow the top-level menu sections to be collapsible, so we can navigate the index easier.
- Make the Left Nav bar independently scrollable
- Allow the top-level sections to be collapsable, so we can navigate the index easier.

begriffs:
Fixes #405


sumedhpathak:
Hey @begriffs I presume this PR is still WIP?

begriffs:
Ha, very much so. 😄

begriffs:
@sumedhpathak it's tricky to get the sidebar menu to stay in position without that persistent gap appearing above it on scroll. That's because it's `position: fixed` and starts out offset to make room for our top banner. I experimented with `position: sticky` which is a newer feature of CSS that is supposed to do what we want. Having trouble there because it causes the content next to it to "clear" below it.

One observation is that part of our difficulty is our custom theme where we have the Citus logo and horizontal menu above everything. The sidebar scrolls independently on read the docs sites that use the default sphinx theme. We might consider going back to the default theme and just putting the citus logo on our index page. In fact the horizontal menu style in the docs no longer matches our web site ever since the site redesign.

begriffs:
OK, check out the preview. I simplified the theme to fix the menu scrolling. Also removed the redundant TOC from the index page and put four sections as an idea of how we might guide people. If you like the content there I can turn it into a nice looking grid.

One thing I can also look into is showing a scrollbar on the menu maybe. See if you feel it provides enough visual cue that the menu can scroll.

lfittl:
Since there were issues with making the scrolling fix work with the header as it is, here is two quick variations that we could do, which look nicer (in my opinion) and we can make the scrolling work there:

screen shot 2017-08-03 at 2 32 20 pm
screen shot 2017-08-03 at 2 37 35 pm

sumedhpathak:
@lfittl Do we have these deployed somewhere as a preview? We could share with Claire/Craig to get their feedback from a branding perspective as well.

begriffs:
They're not deployed -- we adjusted the properties live in the DOM in Chrome and took screenshots of the results. No actual code was changed or saved.

lfittl:
Yeah, this was just a quick hack - no easy way to persist that. I would recommend sharing the screenshots as discussion starters.

PR #418, opened 2017-07-25
Using pg_dist_placement rather than pg_dist_shard_placement

### LINKED ISSUE #382: Document breaking changes for Citus 7.0
## How does this work? (devs)

TBD?

begriffs:
Fixes #382

### DIFF
--- a/admin_guide/cluster_management.rst
+++ b/admin_guide/cluster_management.rst
@@ -133,8 +133,10 @@ The new shard(s) are created on the same node as the shard(s) from which the ten
 
   -- find the node currently holding the new shard
   SELECT nodename, nodeport
-    FROM pg_dist_shard_placement
-   WHERE shardid = 102240;
+    FROM pg_dist_placement AS placement,
+         pg_dist_node AS node
+   WHERE placement.groupid = node.groupid
+     AND shardid = 102240;
 
   -- list the available worker nodes that could hold the shard
   SELECT * FROM master_get_active_worker_nodes();
@@ -205,10 +207,12 @@ To find the worker node holding the data for store id=4, ask for the placement o
 .. code-block:: postgresql
 
   SELECT *
-    FROM pg_dist_shard_placement
-   WHERE shardid = (
-     SELECT get_shard_id_for_distribution_column('stores', 4)
-   );
+    FROM pg_dist_placement AS placement,
+         pg_dist_node AS node
+   WHERE placement.groupid = node.groupid
+     AND shardid = (
+       SELECT get_shard_id_for_distribution_column('stores', 4)
+     );
 
 The output contains the host and port of the worker database.
 
--- a/admin_guide/cluster_management.rst
+++ b/admin_guide/cluster_management.rst
@@ -136,6 +136,7 @@ The new shard(s) are created on the same node as the shard(s) from which the ten
     FROM pg_dist_placement AS placement,
          pg_dist_node AS node
    WHERE placement.groupid = node.groupid
+     AND node.noderole = 'primary'
      AND shardid = 102240;
 
   -- list the available worker nodes that could hold the shard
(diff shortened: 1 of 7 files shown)


admin_guide/cluster_management.rst
```diff
@@ -133,8 +133,10 @@ The new shard(s) are created on the same node as the shard(s) from which the ten
 
   -- find the node currently holding the new shard
   SELECT nodename, nodeport
-    FROM pg_dist_shard_placement
-   WHERE shardid = 102240;
+    FROM pg_dist_placement AS placement,
+         pg_dist_node AS node
+   WHERE placement.groupid = node.groupid
+     AND shardid = 102240;
```
begriffs: ✅ Tested (but with a different shardid)

reference/metadata_tables.rst
```diff
@@ -111,33 +111,46 @@ The shardstorage column in pg_dist_shard indicates the type of storage used for
...
+  SELECT * from pg_dist_placement;
+    shardid | shardstate | shardlength | placementid | groupid
+   ---------+------------+-------------+-------------+---------
+     102008 |          1 |           0 |           1 |       1
+     102008 |          1 |           0 |           2 |       2
+     102009 |          1 |           0 |           3 |       2
+     102009 |          1 |           0 |           4 |       3
+     102010 |          1 |           0 |           5 |       3
+     102010 |          1 |           0 |           6 |       4
+     102011 |          1 |           0 |           7 |       4
```
begriffs: ✅ Verified columns are correct (modulo column order)

begriffs:
I tested it except the upgrading guide and the real-time use case.

The output looked good, although the order of the columns in some of those tables differed from what's in the docs, but all the columns are present.

admin_guide/cluster_management.rst
```diff
@@ -133,8 +133,10 @@ The new shard(s) are created on the same node as the shard(s) from which the ten
 
   -- find the node currently holding the new shard
   SELECT nodename, nodeport
-    FROM pg_dist_shard_placement
-   WHERE shardid = 102240;
+    FROM pg_dist_placement AS placement,
+         pg_dist_node AS node
+   WHERE placement.groupid = node.groupid
```
sumedhpathak: Should we also maybe query for where noderole is 'primary'? In a cluster with followers, this will return multiple rows.

use_case_guide/multi_tenant.rst
```diff
@@ -494,8 +494,10 @@ Next we move the data across the network to a new dedicated node. Create a new n
   -- find the node currently holding the new shard
 
   SELECT nodename, nodeport
-    FROM pg_dist_shard_placement
-   WHERE shardid = 102240;
+    FROM pg_dist_placement AS placement,
+         pg_dist_node AS node
+   WHERE placement.groupid = node.groupid
```
sumedhpathak: Same comment here (to also add a filter where noderole is 'primary'.

admin_guide/upgrading_citus.rst
```diff
@@ -123,7 +123,7 @@ On the Coordinator Node
 
     CREATE TABLE public.pg_dist_partition AS SELECT * FROM pg_catalog.pg_dist_partition;
     CREATE TABLE public.pg_dist_shard AS SELECT * FROM pg_catalog.pg_dist_shard;
-    CREATE TABLE public.pg_dist_shard_placement AS SELECT * FROM pg_catalog.pg_dist_shard_placement;
+    CREATE TABLE public.pg_dist_placement AS SELECT * FROM pg_catalog.pg_dist_placement;
```
jasonmp85: Just dropping a note here that during the upgrade process this was the only notable issue I ran into. Basically I copy-pasted all the commands and just removed occurrences of the string `shard_` from them and things worked thereafter. This was as a result of upgrading from Citus 6.2/PostgreSQL 9.6 to Citus 7.0/PostgreSQL 9.6 and then finally Citus 7.0/PostgreSQL 10. I repeated the steps beginning with Citus 6.0 instead of 6.2 and the need still exists there, of course.

PR #419, opened 2017-07-26
Update supported pg versions to mention pg 10 (not much to update)

### LINKED ISSUE #378: Support for PostgreSQL 10

begriffs:
@sumedhpathak actually our docs don't talk much about which versions of postgres work with citus. The upgrading section does have a section, "Upgrading PostgreSQL version from 9.5 to 9.6." I could change it to "Upgrading PostgreSQL version from 9.5 to 10" and test the instructions on the pg beta. Alternately I could add another section, "Upgrading PostgreSQL version from 9.6 to 10."

I did update all the links to pg docs to unpin them from 9.6 in particular and onto `current` instead. The only risk there is if pg removes some of the sections in future versions. However I think the sections we link to will probably stick around.

Also our installation instructions use packages specific to 9.6, but I'm assuming that we won't change that until pg 10 is out of beta.

Our [site FAQ](https://www.citusdata.com/faq) does have a table of "Which PostgreSQL versions are compatible with Citus?" and I opened a [PR](https://github.com/citusdata/citusdata.com/pull/135) there to include pg 10 for Citus 6.x (should we add Citus 7 in there too?).

Fixes #378


jasonmp85:
@sumedhpathak I left a comment on the one place in the documentation that I needed to change something; as anticipated it was just around the change from `pg_dist_shard_placement` to `pg_dist_placement`.

The logic to re-add triggers and dependencies appeared to work correctly but I'm not 100% sure it's exhaustive (at any rate, `TRUNCATE` against distributed tables worked fine after the upgrade). I'll do a quick check in the morning to see whether we've added any trigger logic since `6.2`, since inspecting the diff is likely easier than exhaustively checking for edge case bugs.

The changes @begriffs made for the `pg_dist_placement` table look correct to me.

PR #430, opened 2017-08-09
Document support for different clusters

### LINKED ISSUE #420: Document support for different clusters
We are adding support for "followers", or read-only clusters.

We are adding a few different GUC's:
Add the `citus.read_from_secondaries` and `citus.current_cluster` GUCs
Add the `nodecluster` column, including as parameters to `master_add{,_inactive}_node`

See: https://github.com/citusdata/citus/pull/1518/files. 

This is still WIP, so contents of the PR might change. Primarily there are some metadata changes and possibly UDF changes/additions we might want to document.

begriffs:
Fixes #420 

This PR changes the reference section. It doesn't cover motivation or use cases.

### DIFF
--- a/reference/user_defined_functions.rst
+++ b/reference/user_defined_functions.rst
@@ -330,6 +330,12 @@ Arguments
 
 **node_port:** The port on which PostgreSQL is listening on the worker node.
 
+**group_id:** Default 0
+
+**node_role:** Default 'primary'
+
+**node_cluster:** Default 'default'
+
 Return Value
 ******************************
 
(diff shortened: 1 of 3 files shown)


reference/user_defined_functions.rst
```diff
@@ -365,22 +371,26 @@ Arguments
 
 **node_port:** The port on which PostgreSQL is listening on the worker node.
 
+**group_id:** Default 0
```
begriffs: What is this, precisely? Is it related the shard placement groups?

reference/user_defined_functions.rst
```diff
@@ -365,22 +371,26 @@ Arguments
...
+
+**node_role:** Default 'primary'
```
begriffs: Options are primary/secondary? Any others?

begriffs:
This is what I pieced together from the PRs. Hopefully it's mostly correct and complete...

begriffs:
This is ready for another look.

reference/user_defined_functions.rst
```diff
@@ -330,6 +330,13 @@ Arguments
 
 **node_port:** The port on which PostgreSQL is listening on the worker node.
 
+**group_id:** A group of one primary server and zero or more secondary
+servers, relevant only for streaming replication.  Default 0
+
+**node_role:** Whether it is 'primary' or 'secondary'. Default 'primary'
+
+**node_cluster:** The sub-64-character cluster name. Default 'default'
```
sumedhpathak: Why did you qualify it as sub-64-character?

begriffs: I didn't check the code, but the character limit was noted in the google doc.

sumedhpathak: I think we can leave it out, as we don't specify it for other text types. The code should be robust to error meaningfully in case someone exceeds such boundaries.

PR #431, opened 2017-08-09
Use a multi-row insert to illustrate that it is possible

### LINKED ISSUE #379: Multi-row insert statements

https://github.com/citusdata/citus/issues/455

begriffs:
Fixes #379 

I didn't see anywhere that explicitly warned against multi-row inserts, but I did find a spot to show one in action. I also grepped the files for "multi.?row", "multi.*insert", and "INSERT INTO" and didn't see anything bad.

### DIFF
--- a/dist_tables/dml.rst
+++ b/dist_tables/dml.rst
@@ -8,23 +8,25 @@ The following code snippets use the Github events example, see :ref:`ddl`.
 Inserting Data
 --------------
 
-Single row inserts
-$$$$$$$$$$$$$$$$$$
+Inserting Row Literals
+$$$$$$$$$$$$$$$$$$$$$$
 
 To insert data into distributed tables, you can use the standard PostgreSQL `INSERT <http://www.postgresql.org/docs/9.6/static/sql-insert.html>`_ commands. As an example, we pick two rows randomly from the Github Archive dataset.
 
 ::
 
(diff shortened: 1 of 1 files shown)


dist_tables/dml.rst
```diff
@@ -8,23 +8,25 @@ The following code snippets use the Github events example, see :ref:`ddl`.
 Inserting Data
 --------------
 
-Single row inserts
-$$$$$$$$$$$$$$$$$$
+Inserting Row Literals
```
begriffs: It's applicable to more than a single row, so just called it a row literal because you're inserting literal values directly.

ozgune: @begriffs -- Instead of editing the example in this section, what do you think about providing another example?

I'd consider multi-row inserts an advanced feature. Most users would start by running single row insert statements. For example, I ran my first multi-row insert statement four years into my Postgres journey.

sumedhpathak: Echoing Ozgun's comment, we can probably leave our basic INSERT example as-is. We could add a separate section (multi-row inserts), and expand a bit on why you may want to do this along with examples.

sumedhpathak:
@begriffs This doesn't seem to have auto-deployed. Could you deploy to Heroku? Thanks!

PR #432, opened 2017-08-10
Front page UI

### LINKED ISSUE #404: Create a navigation guide on main page
Exact structure, content, and display TBD. We want a condensed guide for users for our docs.
At a very minimum, we could provide a 1-2 line description of some of the important top-level sections.

As part of this, we also want to add a "Featured Articles" section on this page, which may rotate to various new pieces of Docs content.

begriffs:
Fixes #404 

At the time of opening, this PR is just a preview.


begriffs:
OK, the rough idea is there, see how you like the styling. Note it looks bad on mobile sizes because I didn't make the grid collapse properly with media breakpoints, but I can invest more time into that if we agree the overall design is sound.

sumedhpathak:
Ok, this looks kinda cool, but aren't the buttons clickable themselves? Do we have to click read-more to get further?

begriffs:
Just the Read More links, but I can add a mouse hover state to the buttons to add an outline or change the background color, and make it so you can click the whole thing.

begriffs:
@sumedhpathak see how you like it now

sumedhpathak:
This looks pretty good! Although I'm wondering if we can make these boxes a bit squarer, rather than the long rectangular ones we have right now. Maybe we can also reduce the font size?
The highlighting also seems a bit slow/laggy. Overall, its promising though, so maybe we chat this week to discuss refining the UI part a bit, and then progress to content?

begriffs:
I actually added an intentional fade-in on hover but can remove that if it feels laggy to you.

The length of the rectangles can of course be adjusted by putting less text in the buttons. I could also use small icons and pull them to the left of the text.

sumedhpathak:
This looks quite good!

One section we're missing is "choosing your partition column".  I had it marked in the "migration" section in the [doc here.](https://docs.google.com/document/d/1gOwoAaH9f2KEwrwqUpcDR3ppE2XbNHtrix8Hp3NxuAc/edit)
However, its a critical enough section that maybe we also add it do Use-cases? (as a next-step?)

sumedhpathak:
Could you also add it to the migration section? (Or if you think it more logical, just in the migration section). Choosing the column is definitely part of the migration section. Other than that its :ship. We should cherry-pick to 7.0.

PR #436, opened 2017-08-16
Fix trigger example in query propagation section

### LINKED ISSUE #434: Update Trigger example which breaks in 7.0
Not sure where to open this issue, but the reason is a regression on Citus, so opening here.

shard_name() now returns quoted schema-qualified identifier instead of just the table name. This breaks our trigger example on docs.
https://docs.citusdata.com/en/v6.2/extra/query_propagation.html#running-on-all-placements

See related https://github.com/citusdata/citus/issues/1568. Marco has a suggestion on how to fix this there.

begriffs:
Fixes #434

### DIFF
--- a/extra/query_propagation.rst
+++ b/extra/query_propagation.rst
@@ -40,14 +40,15 @@ The :code:`run_command_on_shards` function applies a SQL command to each shard, 
   -- Get the estimated row count for a distributed table by summing the
   -- estimated counts of rows for each shard.
   SELECT sum(result::bigint) AS estimated_count
-    FROM run_command_on_shards(
-      'my_distributed_table',
-      $cmd$
-        SELECT reltuples
-          FROM pg_class
-         WHERE relname = '%I';
-      $cmd$
-    );
+  FROM run_command_on_shards(
+    'my_distributed_table',
+    $cmd$
+      SELECT reltuples
+        FROM pg_class c
+        JOIN pg_catalog.pg_namespace n on n.oid=c.relnamespace
+       WHERE n.nspname||'.'||relname = '%s';
+    $cmd$
+  );
 
 Running on all Placements
 -------------------------
@@ -65,7 +66,7 @@ For example, suppose a distributed table has an :code:`updated_at` field, and we
   SELECT run_command_on_placements(
     'my_distributed_table',
     $cmd$
-      UPDATE %I SET updated_at = '2017-01-01';
+      UPDATE %s SET updated_at = '2017-01-01';
     $cmd$
   );
 
--- a/extra/query_propagation.rst
+++ b/extra/query_propagation.rst
@@ -82,14 +82,17 @@ A useful companion to :code:`run_command_on_placements` is :code:`run_command_on
   -- We want to synchronise them so that every time little_vals
   -- are created, big_vals appear with double the value
   --
-  -- First we make a trigger function for each placement
-  SELECT run_command_on_placements('big_vals', $cmd$
-    CREATE OR REPLACE FUNCTION embiggen_%1$I() RETURNS TRIGGER AS $$
+  -- First we make a trigger function on each worker, which will
+  -- take the destination table placement as an argument
+  SELECT run_command_on_workers($cmd$
+    CREATE OR REPLACE FUNCTION embiggen() RETURNS TRIGGER AS $$
       BEGIN
         IF (TG_OP = 'INSERT') THEN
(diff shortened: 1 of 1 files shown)


extra/query_propagation.rst
```diff
@@ -80,14 +80,17 @@ A useful companion to :code:`run_command_on_placements` is :code:`run_command_on
...
-    CREATE OR REPLACE FUNCTION embiggen_%1$I() RETURNS TRIGGER AS $$
+  -- First we make a trigger function on each worker, which will
+  -- take the destination table placement as an argument
+  SELECT run_command_on_workers($cmd$
+    CREATE OR REPLACE FUNCTION embiggen() RETURNS TRIGGER AS $$
       BEGIN
         IF (TG_OP = 'INSERT') THEN
-          INSERT INTO %1$I (key, val) VALUES (NEW.key, NEW.val*2);
+          EXECUTE format(
+            'INSERT INTO %s (key, val) SELECT ($1).key, ($1).val*2;',
```
begriffs: Note that `%s` rather than `%I` was indeed necessary here because postgres considers `"public.foo"` different from `"public"."foo"`.

begriffs: Also `$1` is substituted *after* format() finishes, by the USING construct.

extra/query_propagation.rst
```diff
@@ -80,14 +80,17 @@ A useful companion to :code:`run_command_on_placements` is :code:`run_command_on
...
+            TG_ARGV[0]
```
begriffs: Postgres arrays are usually 1-based, but TG_ARGV is 0-based...

extra/query_propagation.rst
```diff
@@ -80,14 +80,17 @@ A useful companion to :code:`run_command_on_placements` is :code:`run_command_on
   -- We want to synchronise them so that every time little_vals
   -- are created, big_vals appear with double the value
   --
-  -- First we make a trigger function for each placement
-  SELECT run_command_on_placements('big_vals', $cmd$
-    CREATE OR REPLACE FUNCTION embiggen_%1$I() RETURNS TRIGGER AS $$
+  -- First we make a trigger function on each worker, which will
+  -- take the destination table placement as an argument
+  SELECT run_command_on_workers($cmd$
+    CREATE OR REPLACE FUNCTION embiggen() RETURNS TRIGGER AS $$
```
begriffs: Trigger functions are not supposed to declare their arguments, or else the type won't match up. That's why we leave the args list empty and consult TG_ARGV.

begriffs:
Tested this in my local citus 7.0-8 installation.

samay-sharma:
hey @begriffs : I tried out these changes and still get a few failing commands.

(a) the estimated row count query returns an error.
(b) the update query should be changed to use %s.
(c) The trigger propagation (run_command_on_colocated_placements) command fails with an error saying:
`ERROR:  permission denied for schema citus`

Could you look to modify them as well?

begriffs:
@samay-sharma I fixed (a) and (b). Noticed that (c) works fine on my local dev cluster, but fails on Cloud. @lfittl suspects it may be an underlying Citus Enterprise issue.

begriffs:
One other question - #471 means that example (b) about cross-shard updates will be unnecessary. Are there other use cases for the `run_command_on_placements` function?

sumedhpathak:
I am also testing this, and the trigger example crashes my coordinator node. It actually runs successfully on the worker nodes though. @begriffs Did you also experience something like this?

begriffs:
@sumedhpathak I had originally tested it locally and it worked (still does, on pg 9.6.5, citus 7.0-15), but looks like this is a Cloud issue. The current version of cloud with pg 9.6.4 and citus 7.0-15 acts like this:

```sql
SELECT run_command_on_colocated_placements(
  'little_vals',
  'big_vals',
  $cmd$
    CREATE TRIGGER after_insert AFTER INSERT ON %s
      FOR EACH ROW EXECUTE PROCEDURE embiggen(%I)
  $cmd$
);

ERROR:  42501: permission denied for schema citus
LINE 8:  table1_placements citus.colocation_placement_type[];
                           ^
```

@lfittl mentioned in chat that there may be insufficient permissions granted in the Citus Enterprise migration scripts.

sumedhpathak:
@begriffs I am OK to merge in these changes, given now we have working examples. The issue on Cloud has been captured here: https://github.com/citusdata/citus/issues/1713.

PR #443, opened 2017-08-17
Simple query fixes

furkansahin:
During the documentation tests for Citus 7.0. I encountered a couple of simple problems in the queries. I fixed some of them on this branch and created issues for some others.

### DIFF
--- a/reference/sql_workarounds.rst
+++ b/reference/sql_workarounds.rst
@@ -57,7 +57,7 @@ In this technique we use PL/pgSQL to construct and execute one statement based o
 
   -- create temporary table with results
   do language plpgsql $$
-    declare user_ids integer[];
+    declare user_ids bigint[];
   begin
     execute
       'select user_id'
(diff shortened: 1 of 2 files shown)


use_case_guide/real_time_dashboards.rst
```diff
@@ -195,7 +195,7 @@ its own function to figure that out:
       SELECT
         a.logicalrelid::regclass||'_'||a.shardid,
         b.logicalrelid::regclass||'_'||b.shardid,
-        nodename, nodeport
+        nodename, nodeport::BIGINT
```
sumedhpathak: This change has already been fixed as part of #445 .

reference/sql_workarounds.rst
```diff
@@ -68,7 +68,7 @@ In this technique we use PL/pgSQL to construct and execute one statement based o
       'create temp table results_temp as '
       'select user_id'
       '  from events'
-      ' where user_id = any(array[%s])'
+      ' where user_id = any(array[%s]::bigint[])'
```
begriffs: `user_ids` is `integer[]`, and this line is `bigint[]`, should they match as one of those types or the other?

PR #448, opened 2017-08-23
Clarify Local vs. Distributed Table steps for hll

### LINKED ISSUE #433: HLL aggreagate steps are missing for distributed tables in Realtime Dashboards tutorial
The Real Time Dashboards tutorial is aimed towards mostly single node Postgres. This causes confusion when users try these queries with distributed tables and see them fail.

Specifically the following HLL query won't work on distributed tables:
```
SELECT
  site_id, date_trunc('minute', ingest_time) as minute,
  hll_add_agg(hll_hash_text(ip_address)) AS distinct_ip_addresses
FROM http_request
WHERE date_trunc('minute', ingest_time) = date_trunc('minute', now())
GROUP BY site_id, minute;
```

We may consider distinguishing between what is distributed and what is non distributed.

aamederen:
Fixes #433

### DIFF
--- a/use_case_guide/real_time_dashboards.rst
+++ b/use_case_guide/real_time_dashboards.rst
@@ -88,7 +88,7 @@ queries such as:
   FROM http_request
   WHERE site_id = 1 AND date_trunc('minute', ingest_time) > now() - interval '5 minutes'
   GROUP BY minute;
- 
+
 The setup described above works, but has two drawbacks:
 
 * Your HTTP analytics dashboard must go over each row every time it needs to generate a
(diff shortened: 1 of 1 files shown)


use_case_guide/real_time_dashboards.rst
```diff
@@ -314,9 +314,41 @@ like this:
 
 .. code-block:: sql
 
+  -- let's first use hll extension on a regular postgres table
+  CREATE TABLE http_request_local AS SELECT * FROM http_request;
```
sumedhpathak: Thanks for opening up this PR!! 

Should we just remove this part here, and focus on the distributed table itself? This is a bit misleading, since it suggests pulling all the data to the master. 

Alternately, we could just clarify this query with a comment saying that these queries are intended to run on the worker shards, and roll-up into an aggregate shard, similar to the examples above.

aamederen: Removing the local part makes sense but I would be happy if we had some way to show that we would use `hll_add_agg` in the non-citus case.

Issue #435, opened 2017-08-16
7.0 Release General Checklist


begriffs:
- [x] s/6\\.\d/7.0/g
- [x] Verify / Update major-version upgrade steps
    - Check that it says 6->7 rather than 5->6
- [x] Check / Document new UDF's added in 7
- [x] Check / Document new config options added in 7
- [x] Update links from Readme and Citus web site to newer version of docs if required
   - (remove this item if they all point at stable now)
- [x] Update all installation steps to use new package name (except CloudFormation)
- [x] Update links to new CloudFormation template (Burak)
- [x] Verify Installation steps with new packages
    - [x] Ubuntu 16.04 (Joe)
    - [x] Ubuntu 14.04
    - [x] Debian Jessie
    - [x] Debian Wheezy
    - [x] Amazon Linux 2017.03.0
    - [x] CentOS 7.3
    - [x] CentOS 6.9
    - [x] RHEL 7.2 
    - [x] RHEL 6.5 
    - [x] Fedora 26
    - [x] Fedora 25
- [x] Test Multi-tenant tutorial and use-case guide (on just one version) (Sumedh)
- [x] Test Real-time analytics tutorial and use-case guide (on just one version) (Joe)
- [x] Test Docker installation. 
- [x] Update citusdata.com/download links.
- [x] Create new 7.0 release branch and set as default ver in RTD (last step)

begriffs:
Verified that Citus installs on the Debian family (it identifies as 7.0-15). Also checked Ubuntu 16.04 *32-bit* and that failed, but I suspect that we're intentionally not providing 32-bit Citus packages.

sumedhpathak:
I've verified RHEL, but no package was found for Fedora 25. I opened an issue in the packaging repo. https://github.com/citusdata/packaging/issues/71. 

Continuing on to tutorial and use-case guide.

sumedhpathak:
Testing Docker, and https://raw.githubusercontent.com/citusdata/docker/master/docker-compose.yml still pull 6.2.3. @jasonmp85 Where do we update this file?

begriffs:
Noting that the real-time use case guide does not work for me. The script to load fake http requests keeps inserting them but the ingest_time is always the exact same. When I kill the loop and start it again it starts with a new time and keeps repeating that new one.

For instance, try running the ingestion loop a few times, then check:

```
select count(*), ingest_time from http_request where site_id = 1 group by ingest_time;
```
There will be only as many entries as the number of times you run the loop.

PR #449, opened 2017-08-23
Resolve the 7.0 upgrade checklist

begriffs:
A running PR to add commits to as we go through the checklist

### DIFF
--- a/cloud/mx.rst
+++ b/cloud/mx.rst
@@ -107,7 +107,7 @@ An important aspect to consider is that horizontally scaling out your processing
 Limitations Compared to Citus
 -----------------------------
 
-All Citus 6.2 features are supported in Citus MX with the following exceptions:
+All Citus 7.0 features are supported in Citus MX with the following exceptions:
 
 Append-distributed tables currently cannot be made available from workers. They can still be used in the traditional way, with queries going through the coordinator. However, append-distributed tables already allowed you to :ref:`bulk_copy`.
 
--- a/installation/requirements.rst
+++ b/installation/requirements.rst
@@ -3,7 +3,7 @@
 Requirements
 ############
 
-Citus works with modern 64-bit Linux and most Unix based operating systems. Citus 6.0 requires PostgreSQL 9.5 or later versions.
+Citus works with modern 64-bit Linux and most Unix based operating systems. Citus 6.x requires PostgreSQL 9.5 or later versions, and Citus 7.0 requires PostgreSQL 9.6 or later.
 
 Before setting up a Citus cluster, you should ensure that the network and firewall settings are configured to allow:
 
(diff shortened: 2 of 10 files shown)


begriffs:
Fixes #435

begriffs:
See commit messages for details about UDFs and GUCs omitted.

installation/requirements.rst
```diff
@@ -3,7 +3,7 @@
 Requirements
 ############
 
-Citus works with modern 64-bit Linux and most Unix based operating systems. Citus 6.0 requires PostgreSQL 9.5 or later versions.
+Citus works with modern 64-bit Linux and most Unix based operating systems. Citus 6.x requires PostgreSQL 9.5 or later versions, and Citus 7.0 requires PostgreSQL 9.6 or later.
```
sumedhpathak: Can we just update this to Citus 7? (Instead of adding onto version 6).

reference/configuration.rst
```diff
@@ -49,6 +49,16 @@ This value defaults to true, and is effective on the coordinator. In rare cases,
 complex upgrade processes may require setting this parameter to false, thus
 disabling the check.
 
+citus.log_distributed_deadlock_detection (boolean)
```
sumedhpathak: These look good, but I might still add `max_task_string_size`. We usually talk about the type of the param, min/max values etc, which might be useful.

reference/user_defined_functions.rst
```diff
@@ -719,6 +718,35 @@ Example
   --------------
   73 MB
 
+citus.find_groupid_for_node
```
sumedhpathak: I don't think this is meant to be external. It also seems low-level. `master_add_secondary_node` probably should be added though. The function comment is here: https://github.com/citusdata/citus/blob/f1b51d7bbe4178e2cd5fb08959e7a4cd87bd6a3b/src/backend/distributed/utils/node_metadata.c#L174. The arguments are different from this and master_add_node. In this function, we also specify which is the primary node for this one.

cloud/mx.rst
```diff
@@ -107,7 +107,7 @@ An important aspect to consider is that horizontally scaling out your processing
 Limitations Compared to Citus
 -----------------------------
 
-All Citus 6.2 features are supported in Citus MX with the following exceptions:
+All Citus 7.0 features are supported in Citus MX with the following exceptions:
```
sumedhpathak: This probably needs updating, but let's skip that for now.

conf.py
```diff
@@ -57,9 +57,9 @@
 # built documents.
 #
 # The short X.Y version.
-version = '6.2'
+version = '7.0'
```
sumedhpathak: This was merged as part of Burak's PR. Could you revert, or rebase with master?

installation/requirements.rst
sumedhpathak: Ping on this one.

Issue #426, opened 2017-08-09
FAQ - Incorrect answer


furkansahin:
On our [FAQ](https://docs.citusdata.com/en/v6.2/faq/faq.html) page we have two questions as 
* [How do I change the shard count for a hash partitioned table?](https://docs.citusdata.com/en/v6.2/faq/faq.html#how-do-i-change-the-shard-count-for-a-hash-partitioned-table)
* [How do I choose the shard count when I hash-partition my data?](https://docs.citusdata.com/en/v6.2/faq/faq.html#how-do-i-choose-the-shard-count-when-i-hash-partition-my-data)
and the answer is the same for both questions;
> Optimal shard count is related to the total number of cores on the workers. Citus partitions an incoming query into its fragment queries which run on individual worker shards. Hence the degree of parallelism for each query is governed by the number of shards the query hits. To ensure maximum parallelism, you should create enough shards on each node such that there is at least one shard per CPU core.
We typically recommend creating a high number of initial shards, e.g. 2x or 4x the number of current CPU cores. This allows for future scaling if you add more workers and CPU cores.

This is a right answer to the question _"How do I choose the shard count when I hash-partition my data?"_. However, this is not answering the question _"How do I change the shard count for a hash partitioned table?"_

sumedhpathak:
@furkansahin Do you have a suggestion here? Thanks!

begriffs:
We're tracking this as part of #454, so closing this one.

PR #450, opened 2017-08-25
The answer to the question 'How do I change the shard count for a hash partitioned table?' is fixed.

furkansahin:
This fixes [#426](https://github.com/citusdata/citus_docs/issues/426)

### DIFF
--- a/faq/faq.rst
+++ b/faq/faq.rst
@@ -15,9 +15,11 @@ You can add nodes to a Citus cluster by calling the master_add_node UDF with the
 How do I change the shard count for a hash partitioned table?
 -------------------------------------------------------------
 
-Optimal shard count is related to the total number of cores on the workers. Citus partitions an incoming query into its fragment queries which run on individual worker shards. Hence the degree of parallelism for each query is governed by the number of shards the query hits. To ensure maximum parallelism, you should create enough shards on each node such that there is at least one shard per CPU core.
+Shard count is a variable in citus and you can always set it by `SET citus.shard_count=[Shard Count];`. After setting the shard count, your new tables are going to be distributed according to the new value.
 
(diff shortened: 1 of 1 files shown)


faq/faq.rst
```diff
@@ -15,9 +15,11 @@ You can add nodes to a Citus cluster by calling the master_add_node UDF with the
...
 -------------------------------------------------------------
 
-Optimal shard count is related to the total number of cores on the workers. Citus partitions an incoming query into its fragment queries which run on individual worker shards. Hence the degree of parallelism for each query is governed by the number of shards the query hits. To ensure maximum parallelism, you should create enough shards on each node such that there is at least one shard per CPU core.
+Shard count is a variable in citus and you can always set it by `SET citus.shard_count=[Shard Count];`. After setting the shard count, your new tables are going to be distributed according to the new value.
 
-We typically recommend creating a high number of initial shards, e.g. 2x or 4x the number of current CPU cores. This allows for future scaling if you add more workers and CPU cores.
+Note that it is not allowed to change the shard count of an already distributed table. However, if it is really necessary, you can create a new table with the same schema, set the shard_count and `INSERT INTO new_table SELECT * FROM old_table;` but this process will take a large amount of time depending on the size of your table.
```
marcocitus: I don't think we should suggest `INSERT...SELECT` since this will pull the entire distributed table to the coordinator. Maybe recommend contacting Citus Data.

(we can then point them to https://gist.github.com/marcocitus/a2fa5d42053416e4c7dfa37fe91ddefa )

sumedhpathak: Agreed on this one. Let's not recommend selecting the entire table. We can say it would involve re-creating the table with the right shards, and thus would involve data copy. But like Marco suggested, we can ask them to contact us (with the website contact-us link).

furkansahin: Made the change accordingly.

begriffs:
Thanks for the suggestion here, I incorporated it into #461

PR #458, opened 2017-08-31
Document workaround for cross-shard view agg

### LINKED ISSUE #227: How to create view containing distributed query

begriffs:
Fixes #227


sumedhpathak:
@begriffs This review primarily talks about a workaround for an unsupported view. However, #227 is about documenting our existing view support. This PR doesn't address all the use-cases Murat listed in #227. Is there an issue which talks about adding this use-case as a specific workaround? (i.e. did it come from SE?). 

If not, I'd rather close this PR, and then document our existing view support based on Murat's examples.

PR #459, opened 2017-08-31
Remove tutorial links from multi-node install docs

### LINKED ISSUE #293: Remove inconsistency around how we point users to tutorials
We provide single node and multiple node installation steps in our docs. We currently point users to our tutorials after a production multiple node Debian or RHEL install, but don't reference the tutorial from anywhere else.

Our tutorials are designed for single node quick installations. We're also planning to focus on Cloud / Docker in our tutorials' first version.

We should probably (?) remove the reference to tutorials from our production multiple node Debian and RHEL install steps -- unless we had a particular motivation for adding them. We could also add a link to Docker / Cloud steps to point users to our tutorial.

begriffs:
Tutorials presuppose docker or cloud

Fixes #293 

I didn't need to add a link to tutorials from Docker install since they already exist. Adding a link from the Cloud section seemed out of place, but we can do it desired.

### DIFF
--- a/installation/production_deb.rst
+++ b/installation/production_deb.rst
@@ -106,9 +106,7 @@ shell should output the worker nodes we added to the pg_dist_node table above.
 
 **Ready to use Citus**
 
-At this step, you have completed the installation process and are ready to use your Citus cluster. To help you get started, we have a :ref:`tutorial<multi_tenant_tutorial>` which has instructions on setting up a Citus cluster with sample data in minutes.
-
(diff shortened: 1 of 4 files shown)


sumedhpathak:
I think part of this change was also to point them from the single-node installation instructions. We still want to point people to the tutorial, but the multi-node production setup probably wasn't the right location.

begriffs:
I'd left out the link from the single node linux installs because the tutorials were written for Docker users, but native single node installs would pretty easily too. Just added the links there.

PR #460, opened 2017-08-31
Hold sphinx version back to fix search in our custom theme

begriffs:
We'll see in the preview deployment whether this fixes search.


begriffs:
Confirmed that search works on this preview, and does not work on a preview of another of the open pull requests.

Issue #453, opened 2017-08-29
Sizing a citus cluster


saicitus:
## Why are we implementing it? (sales eng)
To help users/prospects/customers come up with the right cluster size.

### What are the typical use cases?
Multi-tenant use cases.
Real-time use cases

### Good locations for content in docs structure
May be a small section which talks about it.

## How does this work? (devs)
**Multitenant SAAS usecases (sharding on tenant)id)**
What we generally recommend to customers/prospects is to choose a cluster such that the hardware(total number of cores and total RAM) on workers is same as that of existing postgres (or other single node db) instance. In such scenarios we have seen 2-3x performance improvements because of sharding and improvement resource utilization.
In case of coordinator size, choose a compute optimized box because coordinator is mostly used for routing queries to the right workers. Depending on the existing workload(write/read throughput) you can go with suitable number of cores.
In Citus cloud by default workers are of the type r4s and coordinator is c4s.

**Realtime Analytics usecases (shard on entity_id)**
**Total Cores:** Considering working data in memory(RAM), ideally you can expect linear performance improvement with the number of cores. Suppose you are on single node db (ex postgres). Based on i) what performance you are getting with the current database and ii) what is the expected requirement with citus, you can decide what is the number of cores. For ex you are currently getting on avg 10s latency on postgres and are looking at 1s latencies, you might choose total 10 cores on workers. 
**RAM:**
In terms of RAM on workers, ideal/best case would be go with RAM such that the majority of the working set fits memory. Working set depends on the queries (filters on the say time)

sumedhpathak:
@begriffs This can be closed I presume?

PR #461, opened 2017-09-12
Recommendations for hardware size and shard count

### LINKED ISSUE #454: Choosing shard count
Currently we have https://docs.citusdata.com/en/v6.2/faq/faq.html#how-do-i-change-the-shard-count-for-a-hash-partitioned-table . This is more relevant to real-time analytics (shard on entity_id) workloads.
But for Multi-tenant SAAS workloads, where you shard on tenant_id it changes. We recommend anywhere from 32 to 128 shards. For smaller workloads say <100GB, you could start with 32 shards and for larger workloads you could choose 64 or 128. This means that you have the leeway to scale from 32 to 128 worker machines.

**Where to put on docs:**
A sub section in sizing (or)
Update the existing FAQ.

begriffs:
Fixes #454

### DIFF
--- a/admin_guide/production_sizing.rst
+++ b/admin_guide/production_sizing.rst
@@ -28,7 +28,7 @@ Multi-Tenant SaaS Use-Case
 
 For those migrating to Citus from an existing single-node database instance, we recommend choosing a cluster where the number of worker cores and RAM in total equals that of the original instance. In such scenarios we have seen 2-3x performance improvements because sharding improves resource utilization, allowing smaller indices etc.
 
-For the coordinator node, choose a compute-optimized machine. The coordinator is responsible for the computationally intensive job of routing queries to the right workers. The number of cores needed on the coordinator depends on your existing workload (write/read throughput). By default in Citus Cloud the workers use Amazon EC2 instance type R4S, and the coordinator uses C4S.
+The coordinator node needs less memory than workers, so you can choose a compute-optimized machine for running the coordinator. The number of cores required depends on your existing workload (write/read throughput). By default in Citus Cloud the workers use Amazon EC2 instance type R4S, and the coordinator uses C4S.
 
 Real-Time Analytics Use-Case
 ----------------------------
--- a/admin_guide/production_sizing.rst
+++ b/admin_guide/production_sizing.rst
@@ -10,7 +10,7 @@ This section explores configuration settings for running a cluster in production
 Shard Count
 ===========
 
-The number of nodes in a cluster is easy to change (see :ref:`scaling_out_cluster`), but the number of shards to distribute among those nodes is more difficult to change after cluster creation. Choosing the shard count is a balance between the flexibility of having more shards, and the overhead for query planning and execution across them.
+The number of nodes in a cluster is easy to change (see :ref:`scaling_out_cluster`), but the number of shards to distribute among those nodes is more difficult to change after cluster creation. Choosing the shard count for each distributed table is a balance between the flexibility of having more shards, and the overhead for query planning and execution across them.
 
 The optimal choice varies depending on your access patterns for the data. For instance, in the :ref:`mt_blurb` use-case we recommend choosing between **32 - 128 shards**.  For smaller workloads say <100GB, you could start with 32 shards and for larger workloads you could choose 64 or 128. This means that you have the leeway to scale from 32 to 128 worker machines.
 
--- a/admin_guide/production_sizing.rst
+++ b/admin_guide/production_sizing.rst
@@ -5,6 +5,8 @@ $$$$$$$$$$$$$$$$$
 
 This section explores configuration settings for running a cluster in production.
 
+.. _prod_shard_count:
+
 Shard Count
 ===========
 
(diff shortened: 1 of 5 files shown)


begriffs:
Included a commit which fixes #453 since this PR already added a sizing section.

begriffs:
This PR updates the FAQ entry about changing shard count. It essentially leaves the question unanswered and talks about good shard counts for new tables. I could also incorporate content from https://github.com/citusdata/citus_docs/pull/450 about `citus.shard_count`.

sumedhpathak:
Do we modify this FAQ in any way?: https://docs.citusdata.com/en/v7.0/faq/faq.html#how-do-i-choose-the-shard-count-when-i-hash-partition-my-data? Maybe we add a brief blurb and point it to this new section?

sumedhpathak:
Also, let's incorporate the change from #450 here, as it probably makes sense to give a more pointed answer there.

admin_guide/production_sizing.rst
```diff
@@ -0,0 +1,38 @@
+.. _production_sizing:
+
+Production Sizing
```
sumedhpathak: Now that we have our fancy new front-page, could we add this to the migration button page?

admin_guide/production_sizing.rst
```diff
@@ -0,0 +1,38 @@
...
+$$$$$$$$$$$$$$$$$
+
+This section explores configuration settings for running a cluster in production.
+
+.. _prod_shard_count:
+
+Shard Count
+===========
+
+The number of nodes in a cluster is easy to change (see :ref:`scaling_out_cluster`), but the number of shards to distribute among those nodes is more difficult to change after cluster creation. Choosing the shard count is a balance between the flexibility of having more shards, and the overhead for query planning and execution across them.
```
sumedhpathak: Could we clarify that number of shards is for an existing distributed table? (not as a whole cluster-wide)?

admin_guide/production_sizing.rst
```diff
@@ -0,0 +1,38 @@
...
+
+The optimal choice varies depending on your access patterns for the data. For instance, in the :ref:`mt_blurb` use-case we recommend choosing between **32 - 128 shards**.  For smaller workloads say <100GB, you could start with 32 shards and for larger workloads you could choose 64 or 128. This means that you have the leeway to scale from 32 to 128 worker machines.
```
sumedhpathak: The real-time analytics answer makes sense to me, but I don't have a clear understanding for this one. We say "you could start with 32 shards", which implies this can be changed later. So we should change that.
Also, I don't think anyone would not choose 64 or 128 based on this recommendation? We don't really say why you would want to choose 32 or 64 or 128. @saicitus/ @samay-sharma  thoughts?

admin_guide/production_sizing.rst
```diff
@@ -0,0 +1,38 @@
...
+--------------------------
+
+For those migrating to Citus from an existing single-node database instance, we recommend choosing a cluster where the number of worker cores and RAM in total equals that of the original instance. In such scenarios we have seen 2-3x performance improvements because sharding improves resource utilization, allowing smaller indices etc.
+
+For the coordinator node, choose a compute-optimized machine. The coordinator is responsible for the computationally intensive job of routing queries to the right workers. The number of cores needed on the coordinator depends on your existing workload (write/read throughput). By default in Citus Cloud the workers use Amazon EC2 instance type R4S, and the coordinator uses C4S.
+
+Real-Time Analytics Use-Case
+----------------------------
+
+**Total cores:** when working data fits in RAM, you can expect a linear performance improvement on Citus proportional to the number of worker cores. To determine the right number of cores for your needs, consider the current latency for queries in your single-node database and the required latency in Citus. Divide current latency by desired latency, and round the result.
```
sumedhpathak: In practice this isn't as clear-cut, as postgreSQL contention often limits us. @saicitus @samay-sharma Are we OK making such a recommendation?

admin_guide/production_sizing.rst
```diff
@@ -0,0 +1,38 @@
...
+=====================
+
+The size of a cluster, in terms of number of nodes and their hardware capacity, is easy to change. (:ref:`Scaling <cloud_scaling>` on Citus Cloud is especially easy.) However you still need to choose an initial size for a new cluster. Here are some tips for a reasonable initial cluster size.
+
+Multi-Tenant SaaS Use-Case
+--------------------------
+
+For those migrating to Citus from an existing single-node database instance, we recommend choosing a cluster where the number of worker cores and RAM in total equals that of the original instance. In such scenarios we have seen 2-3x performance improvements because sharding improves resource utilization, allowing smaller indices etc.
+
+For the coordinator node, choose a compute-optimized machine. The coordinator is responsible for the computationally intensive job of routing queries to the right workers. The number of cores needed on the coordinator depends on your existing workload (write/read throughput). By default in Citus Cloud the workers use Amazon EC2 instance type R4S, and the coordinator uses C4S.
```
sumedhpathak: "The coordinator is responsible for the computationally intensive job of routing queries" --> We typically say the co-ordinator doesn't do as-much work, and we don't deem query routing as computationally intensive (from a communicatino standpoint). Can we reword this to say something like the co-ordinator doesn't hold any data (and only routes and does final aggregation), and so we can use compute-optimized machines, rather than memory-optimized?

admin_guide/production_sizing.rst
begriffs: (Added)

begriffs:
Addressed the comments so far, and awaiting feedback from @samay-sharma or @saicitus

faq/faq.rst
```diff
@@ -14,13 +14,6 @@ How do I add nodes to an existing Citus cluster?
 
 You can add nodes to a Citus cluster by calling the master_add_node UDF with the hostname (or IP address) and port number of the new node. After adding a node to an existing cluster, it will not contain any data (shards). Citus will start assigning any newly created shards to this node. To rebalance existing shards from the older nodes to the new node, the Citus Enterprise edition provides a shard rebalancer utility. You can find more information about shard rebalancing in the :ref:`cluster_management` section.
 
-How do I change the shard count for a hash partitioned table?
```
sumedhpathak: I'd prefer retaining this question, as it did come up somewhat frequently, rather than adding a note to the end of the other question.

admin_guide/production_sizing.rst
```diff
@@ -0,0 +1,38 @@
...
+This section explores configuration settings for running a cluster in production.
+
+.. _prod_shard_count:
+
+Shard Count
+===========
+
+The number of nodes in a cluster is easy to change (see :ref:`scaling_out_cluster`), but the number of shards to distribute among those nodes is more difficult to change after cluster creation. Choosing the shard count for each distributed table is a balance between the flexibility of having more shards, and the overhead for query planning and execution across them.
+
+The optimal choice varies depending on your access patterns for the data. For instance, in the :ref:`mt_blurb` use-case we recommend choosing between **32 - 128 shards**.  For smaller workloads say <100GB, you could start with 32 shards and for larger workloads you could choose 64 or 128. This means that you have the leeway to scale from 32 to 128 worker machines.
```
sumedhpathak: Should we create a separate sub-section for multi-tenant vs real-time like we did below for hardware sizing? (For consistency). Or do you think its too sparse for that?

faq/faq.rst
```diff
@@ -59,13 +52,29 @@ Other queries which, by contrast, combine data from multiple nodes, do not suppo
...
+One of the choices when first distributing a table is its shard count. This setting can be set differently for each table, and the optimal value depends on use-case. It is possible, but difficult, to change the count after cluster creation, so use these guidelines to choose the right size.
+
+In the :ref:`mt_blurb` use-case we recommend choosing between 32 - 128 shards.  For smaller workloads say <100GB, you could start with 32 shards and for larger workloads you could choose 64 or 128. This means that you have the leeway to scale from 32 to 128 worker machines.
+
+In the :ref:`rt_blurb` use-case, shard count should be related to the total number of cores on the workers. To ensure maximum parallelism, you should create enough shards on each node such that there is at least one shard per CPU core. We typically recommend creating a high number of initial shards, e.g. 2x or 4x the number of current CPU cores. This allows for future scaling if you add more workers and CPU cores.
```
sumedhpathak: This info is good, but could you also link to the production sizing section (e.g. for more information see ...)?

faq/faq.rst
```diff
@@ -59,13 +52,29 @@ Other queries which, by contrast, combine data from multiple nodes, do not suppo
...
+
+  SET citus.shard_count = 64;
+  -- any tables distributed at this point will have
+  -- sixty-four shards
+
+How do I change the shard count for a hash partitioned table?
+-------------------------------------------------------------
 
-We typically recommend creating a high number of initial shards, e.g. 2x or 4x the number of current CPU cores. This allows for future scaling if you add more workers and CPU cores.
+Note that it is not straightforward to change the shard count of an already distributed table. If you need to do so, please `Contact Us <https://www.citusdata.com/about/contact_us>`_. It's good to think about shard count carefully at distribution time, see :ref:`faq_choose_shard_count`.
```
sumedhpathak: I'm good to retain this note, or link to the other FAQ item, but I'd prefer to keep the other FAQ as well.

faq/faq.rst
begriffs: The github diff layout doesn't make it very apparent, but this question still exists in the FAQ. It's just moved below "How do I choose the shard count." Still a top-level entry in the document, though its answer is pretty brief.

faq/faq.rst
begriffs: It is actually the same question, but moved lower. Are you saying you prefer the old answer to the question over this new text?

faq/faq.rst
begriffs: Good idea. Done.

admin_guide/production_sizing.rst
begriffs: Sure, we do it that way in the next section so it's more consistent to do it here too.

PR #462, opened 2017-09-12
Clean up font style

begriffs:
Remove uses of "Roboto Slab" for headers, and stick with Lato like in the text. Also changes the base font size.

The font size ratio of h1, h2, ... to paragraph text is handled by the Wyrm library. http://wyrmsass.org/section-4.html  It would be harder and hackier to, say, make headers smaller while leaving text size unchanged. I'd have to fight the library. As it is I shrunk the general base font size, making everything smaller.


_themes/citus/static/css/badge_only.css
```diff
@@ -1,2 +1,226 @@
...
+@charset "UTF-8";
```
sumedhpathak: Is this part just formatting? Or is anything changing here?

begriffs:
@sumedhpathak good observation about the big file diffs. They were caused because during development I compiled the styles into their final form by running `grunt sass:dev`, which preserves human-friendly whitespace. However just now I ran `grunt sass:build` instead which removes unnecessary whitespace to make the download slightly faster. That's how the file is normally prepared, so this made the diff smaller in this PR.

PR #479, opened 2017-09-15
Update docker-compose instructions

jasonmp85:
Figured we should stop directing users to an old `docker-compose` binary.

In addition, this switches from the `-p` flag to `COMPOSE_PROJECT_NAME` to ensure that environment variable is set. `-p` behaves the same but does not set the variable. See [here]()https://github.com/docker/compose/issues/3431#issuecomment-221291167) for more.

Finally, I added the `-v` flag to `docker-compose` down, which ensures any automatic volumes are removed when the cluster is stopped.

### DIFF
--- a/installation/single_machine_docker.rst
+++ b/installation/single_machine_docker.rst
@@ -20,7 +20,7 @@ This section describes setting up a Citus cluster on a single machine using dock
   sudo usermod -aG docker $USER && exec sg docker newgrp `id -gn`
   sudo systemctl start docker
 
-  sudo curl -sSL https://github.com/docker/compose/releases/download/1.11.2/docker-compose-`uname -s`-`uname -m` -o /usr/local/bin/docker-compose
+  sudo curl -sSL https://github.com/docker/compose/releases/download/1.16.1/docker-compose-`uname -s`-`uname -m` -o /usr/local/bin/docker-compose
   sudo chmod +x /usr/local/bin/docker-compose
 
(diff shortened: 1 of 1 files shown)


begriffs:
Glad you updated the version, was only a matter of time before it got out of date.

For the other change I must admit the env var looks more ungainly than `-p`, and the comment you linked didn't exactly motivate the change for me:
> if you need to set the project name to prepend the name of your container AND you need to access the project name inside the docker-compose file, via ${COMPOSE_PROJECT_NAME} then you should use export the ENV variable COMPOSE_PROJECT_NAME.

Do we need access to the project name inside the compose file?

jasonmp85:
> Do we need access to the project name inside the compose file?

Yes.

begriffs:
I tried these instructions on my mac (including installing the newer docker-compose) and it worked fine.

If you want to document `MASTER_EXTERNAL_PORT` that would be a positive as well, but other than that it looks good to me.

sumedhpathak:
@begriffs Are these changes approved? If so, could you go ahead and mark the review as such? We should also cherry-pick this to 7.0.

begriffs:
( @jasonmp85 between Sumedh and I we usually go with the convention that the person who opens the PR merges it after review. Want to do the honors? )

jasonmp85:
Oh, didn't know that :-P I review enough people on the other side of the world that I've gotten into habits of merging things myself rather than waiting for another day… Sure.

PR #486, opened 2017-09-25
Reflect that 2PC is now the default

### LINKED ISSUE #472: 2PC is now enabled by default
We have notices in docs telling people to enable 2pc, those notices can be removed now.

begriffs:
Fixes #472


sumedhpathak:
It's possible this won't make it into 7.1. I'll chat with Ozgun on deciding to deprioritize this issue.

sumedhpathak:
Confirmed this won't make 7.1. We can probably just close this PR instead of having it hanging around.

sumedhpathak:
ok. :) So we made other 2PC changes for 7.2 I think? Can we just reopen this one?

PR #493, opened 2017-10-10
Describe PITR on Cloud

### LINKED ISSUE #484: Update Cloud Documentation to describe PITR
## Why are we implementing it? (sales eng) /  What are the typical use cases?

Similar to fork in that it creates a copy of your formation, but this allows customers to go back in time to a previous state of their formation.

The main use case here is when data gets lost, e.g. customer deletes a record and you want to recover it, or an engineer drops a table by accident.

Detailed, users will want to know how long it takes (depends on data size), whether it has any impact on the source (it does not)

We should also mention that the earliest recovery point is about 10 days back, as thats how long we keep base backups and WAL around. The user interface will prevent you from going further than that, and also show you the earliest possible recovery time.

Cloud

### Example

screen shot 2017-09-20 at 12 29 27 pm

### Corner cases, gotchas

Nodes have different base backup timings and sizes, so it can happen that one node restores faster than another.

Not yet, but there will be

begriffs:
Fixes #484

I also combined the Cloud submenu into a better hierarchy since it was getting too expansive. Speaking for myself I like the new organization although my only caution is that if outside web pages link directly to any of those inner cloud pages then some of the URLs have changed and will break. We could investigate on google webmaster tools to verify whether there are external links to those pages.


begriffs:
Fixes #10 too (noticed it while reviewing old issues)

cloud/availability.rst
```diff
@@ -163,6 +166,66 @@ Such crashes are a nightmare. But crash recovery is a way to make sure
 you sleep well at night by making sure none of your or your customers
 data is lost and your downtime is minimal.
 
+.. _cloud_forking:
+
+Forking
+=======
+
+Forking a Citus Cloud formation makes a copy of the cluster's data at an instant in time and produces a new formation in precisely that state. It allows you to change, query, or generally experiment with production data in a separate protected environment. Fork creation runs quickly, and you can do it as often as you want without causing any extra load on the original cluster. This is because forking doesn't query the cluster, rather it taps into the write-ahead logs for each database in the formation.
```
lfittl: Maybe say "at the current point in time" instead of "instant in time" (which a non-native speaker might read as "instantly")

cloud/availability.rst
```diff
@@ -163,6 +166,66 @@ Such crashes are a nightmare. But crash recovery is a way to make sure
...
+Citus Cloud makes forking easy. The control panel for each formation has a "Fork" tab. Go there and enter the name, region, and node sizing information for the destination cluster.
+
+.. image:: ../images/cloud-fork.png
+
+Shortly after you click "Fork Formation," the new formation will appear in the Cloud console. It runs on separate hardware and your database can connect to it in the :ref:`usual way <connection>`.
+
+When is it Useful
+-----------------
+
+A fork is a great place to do experiments. Do you think that denormalizing a table might speed things up? What about creating a roll-up table for a dashboard? How can you persuade your manager that you need more RAM in the coordinator node rather than in the workers? You could prove yourself if only you could try your idea on the production data.
```
lfittl: Maybe say "persuade your colleagues" rather than "persuade your manager" (which implies a certain person reading this)

cloud/availability.rst
```diff
@@ -163,6 +166,66 @@ Such crashes are a nightmare. But crash recovery is a way to make sure
...
+
+.. image:: ../images/cloud-pitr.png
+
+Similar to forking, PITR creates a copy of your formation and leaves the original unaffected. It uses the same operations internally as forking does: base backups and WAL shipping. This means that performing PITR causes no extra load on the original database.
+
+Citus Cloud stores base backups and WAL records for up to ten days, which limits the recovery period to that amount of time. The user interface will prevent you from going back further, and will show the earliest possible recovery time.
+
+Note that Citus worker nodes have different base backup timings and sizes, so it can happen that one node restores faster than another. In general the recovery process begins with restoring base backups at roughly 30MB/second. When that is complete the Cloud Console will provide a progress meter as it replays WAL events. The length of time for this stage varies by the number of events that must be replayed.
```
lfittl: Maybe say "as it replays transactions from WAL" and "varies by the number of transactions that must be replayed"

begriffs:
@lfittl great editing suggestions, thanks.

PR #494, opened 2017-10-11
updating details of cloud support

craigkerstiens:


begriffs:
@craigkerstiens I rebased this on the cloud TOC reorganization. Do you think it's OK to have this inside another top level section, or would you rather it existed as its own top level item in the cloud menu?

PR #497, opened 2017-10-16
Privacy section

### LINKED ISSUE #478: Reveal upcoming anonymized statistics gathering

To better understand the growth and usage of Citus across various platforms, we will be adding a "call home" feature. It collects non-identifying information and users can opt out of it.

We need to be very transparent that this is happening, explain what is collected, and how a user can opt out if desired. Privacy is always important, and especially in a data product that users trust to hold their own and their clients' information.

A prominent "Privacy" section, perhaps even in the "About Citus" area of the TOC. Also link to it from https://docs.citusdata.com/en/v7.0/portals/getting_started.html .

begriffs:
Fixes #478 

@pykello can you confirm that the information here is accurate?


sumedhpathak:
@ozgune Could you go over this change to ensure we're making the right statements from a product perspective?

aboutcitus/privacy.rst
```diff
@@ -0,0 +1,34 @@
...
+Reporting happens on installation and every twenty-four hours thereafter. Here is what happens:
+
+* Citus checks if there is a newer version of itself, and if so emits a notice to the database logs.
+* Citus collects and sends these statistics about your cluster:
+   * Randomly generated cluster identifier
+   * Number of servers
+   * OS and version (output of ``uname -psr`` command)
+   * Number of tables, rounded to a power of two
+   * Total size of shards, rounded to a power of two
+
```
pykello: Also, whether Citus is running inside Docker or not.

More accurately, we send everything in the `pg_dist_node_metadata` table to the server. This table contains only one row/one column which is a `jsonb` field. By default, it only contains the random `server_id`, which is the cluster identifier you mentioned earlier. In citusdata/docker#50, we add a `docker` field to this value, so we also know if the collected report is from inside docker image or not.

aboutcitus/privacy.rst
```diff
@@ -0,0 +1,34 @@
...
+.. code-block:: ini
+
+  citus.enable_statistics_collection = off
+
+If you have super-user SQL access you can also achieve this without needing to find and edit the configuration file. Just execute the following statement in psql:
+
+.. code-block:: postgresql
+
+  ALTER SYSTEM SET citus.enable_statistics_collection = 'off';
```
pykello: Since Docker users won't have the chance to edit this GUC variable before running the image and we send the first report when we run the image, we also added a Docker flag to disable reports at citusdata/docker#42.

aboutcitus/privacy.rst
```diff
@@ -0,0 +1,34 @@
...
+   * Total size of shards, rounded to a power of two
+
+How to Opt Out
+--------------
+
+If you wish to disable our anonymized cluster statistics gathering, set the following GUC in postgresql.conf on your coordinator node:
+
+.. code-block:: ini
+
+  citus.enable_statistics_collection = off
```
pykello: In the current release, this also disables update checks.

begriffs: (Noted)

aboutcitus/privacy.rst
begriffs: (Noted)

installation/single_machine_rhel.rst
```diff
@@ -75,6 +75,10 @@ Finally, the coordinator needs to know where it can find the workers. To tell it
   psql -p 9700 -c "SELECT * from master_add_node('localhost', 9701);"
   psql -p 9700 -c "SELECT * from master_add_node('localhost', 9702);"
 
+.. note::
```
sumedhpathak: Should these notes be at the end after the "verify installation has succeeded" step?

admin_guide/cluster_management.rst
```diff
@@ -193,6 +193,54 @@ The coordinator node needs to know roles' passwords in order to communicate with
...
+* Citus checks if there is a newer version of itself, and if so emits a notice to the database logs.
+* Citus collects and sends these statistics about your cluster:
+   * Randomly generated cluster identifier
+   * Number of workers
+   * OS version and hardware type (output of ``uname -psr`` command)
+   * Number of tables, rounded to a power of two
+   * Total size of shards, rounded to a power of two
+   * Whether Citus is running in Docker or natively
+
+Because Citus is an open-source PostgreSQL extension, the statistics reporting code is available for you to audit. See :code:`CitusMaintenanceDaemonMain` in `maintenanced.c <https://github.com/citusdata/citus/blob/master/src/backend/distributed/utils/maintenanced.c>`_ and the functions in `statistics_collection.c <https://github.com/citusdata/citus/blob/master/src/backend/distributed/utils/statistics_collection.c>`_.
```
sumedhpathak: Any reason to include the link to maintenance daemon? Since that seems mostly just basic background worker stuff. Also, if we want to be more specific, we can provide the exact link: https://github.com/citusdata/citus/blob/master/src/backend/distributed/utils/statistics_collection.c#L85. But the file is fine as well.

admin_guide/cluster_management.rst
```diff
@@ -193,6 +193,54 @@ The coordinator node needs to know roles' passwords in order to communicate with
...
+#########################################
+
+Unless you opt out, Citus checks if there is a newer version of itself during installation and every twenty-four hours thereafter. If a new version is available, Citus emits a notice to the database logs.
+
+During the check for updates, Citus also sends general information about the running cluster to Citus Data company servers. This helps us understand how Citus is commonly used and thereby improve the product. As explained below, the reporting is opt-out and does **not** contain personally identifying information about schemas, tables, queries, or data.
+
+What we Collect
+---------------
+
+* Citus checks if there is a newer version of itself, and if so emits a notice to the database logs.
```
sumedhpathak: What is the notice? Should we document this separately in Docs?

admin_guide/cluster_management.rst
begriffs: I included that link because I thought it would show when Citus is attempting to call home, but I can remove the link for simplicity.

Also I didn't link to a specific line number because it seemed likely to change and break the link.

installation/single_machine_rhel.rst
begriffs: Actually I might want to move the note even earlier, right about where someone is about to start the database. I want to avoid them starting it up, and then learning about the stats collection after it's too late and the software has already reported on them.

Admittedly there's not much for a new empty cluster to report, so it's not that big of a deal. I can move the notice wherever you think best.

admin_guide/cluster_management.rst
begriffs: The only other spots in the docs talking about logging are a section in Cloud, multi task log level in configuration, and in our imported articles.

I can add a little example of the log message right in this section.

begriffs: (added example log message in intro section above)

admin_guide/cluster_management.rst
```diff
@@ -193,6 +193,58 @@ The coordinator node needs to know roles' passwords in order to communicate with
...
+.. _phone_home:
+
+Checks For Updates and Cluster Statistics
+#########################################
+
+Unless you opt out, Citus checks if there is a newer version of itself during installation and every twenty-four hours thereafter. If a new version is available, Citus emits a notice to the database logs:
+
+.. code-block::
+
+  a new minor release of Citus (X.Y.Z) is available
```
samay-sharma: I don't see the code block including the notice in the [deployment](https://citus-doc-preview-pr-497.herokuapp.com/admin_guide/cluster_management.html#checks-for-updates-and-cluster-statistics).

admin_guide/cluster_management.rst
```diff
@@ -193,6 +193,58 @@ The coordinator node needs to know roles' passwords in order to communicate with
...
+
+During the check for updates, Citus also sends general information about the running cluster to Citus Data company servers. This helps us understand how Citus is commonly used and thereby improve the product. As explained below, the reporting is opt-out and does **not** contain personally identifying information about schemas, tables, queries, or data.
+
+What we Collect
+---------------
+
+* Citus checks if there is a newer version of itself, and if so emits a notice to the database logs.
+* Citus collects and sends these statistics about your cluster:
```
samay-sharma: The bolding of "Citus collects and sends these statistics about your cluster:" visually seems a bit odd to me.

installation/production_deb.rst
```diff
@@ -60,6 +60,10 @@ Before starting the database let's change its access permissions. By default the
 
 **4. Start database servers, create Citus extension**
 
+.. note::
+
+  Before starting the cluster, be aware that by default Citus reports anonymous usage information to Citus Data company servers. To learn more about this, including how to opt out, see :ref:`phone_home`.
```
samay-sharma: usage statistics makes me feel that we're collecting information on how the cluster is being used like how many queries or utilization or things like that, where as we're just collecting information about the cluster i.e. how many nodes, how many tables etc. What do you think about changing the notice to:

`Please note that Citus reports anonymous information about your cluster to the Citus Data company servers. To learn more about what information is collected and how to opt out of it, see phone home.`

admin_guide/cluster_management.rst
```diff
@@ -193,6 +193,58 @@ The coordinator node needs to know roles' passwords in order to communicate with
...
+Checks For Updates and Cluster Statistics
+#########################################
+
+Unless you opt out, Citus checks if there is a newer version of itself during installation and every twenty-four hours thereafter. If a new version is available, Citus emits a notice to the database logs:
+
+.. code-block::
+
+  a new minor release of Citus (X.Y.Z) is available
+
+During the check for updates, Citus also sends general information about the running cluster to Citus Data company servers. This helps us understand how Citus is commonly used and thereby improve the product. As explained below, the reporting is opt-out and does **not** contain personally identifying information about schemas, tables, queries, or data.
```
samay-sharma: During the check for updates, Citus also sends general *anonymized* information

installation/single_machine_deb.rst
```diff
@@ -52,6 +52,10 @@ Citus is a Postgres extension, to tell Postgres to use this extension you'll nee
 
 **3. Start the coordinator and workers**
 
+.. note::
```
samay-sharma: @begriffs : Could we move this note to the end?

admin_guide/cluster_management.rst
begriffs: Sphinx was interpreting the list item wrong, but I see how to fix it.

PR #501, opened 2017-10-19
Bump version in Cloudformation template to 7.0.3

byucesoy:


conf.py
```diff
@@ -59,7 +59,7 @@
 # The short X.Y version.
 version = '7.0'
 # The full version, including alpha/beta/rc tags.
-release = '7.0.2'
+release = '7.0.3'
```
begriffs: I don't really know why we track the patch version in docs like this, but we've been doing it that way so I guess we can continue doing so. Maybe at the next major version we can make `version` and `release` always match.

begriffs:
If you've tested deploying with the new CF template and verified in psql that `\dx` reports the Citus version you are expecting than this looks good to me!

PR #511, opened 2017-10-23
Cross-shard update/delete!

### LINKED ISSUE #471: Cross-shard updates/deletes now work seamlessly
We no longer need to call out a separate section https://docs.citusdata.com/en/v7.0/dist_tables/dml.html?highlight=master_modify_multiple_shards#cross-shard-updates-and-deletion and don't need master_modify_multiple_shards for this. The standard DML statements work in Citus 7.1.

begriffs:
Fixes #471

### DIFF
--- a/dist_tables/dml.rst
+++ b/dist_tables/dml.rst
@@ -150,7 +150,7 @@ The situation changes when dealing with late arriving data, or running the rollu
 Updates and Deletion
 --------------------
 
-You can update or delete rows from your distributed tables using the standard PostgreSQL `UPDATE <http://www.postgresql.org/docs/current/static/sql-update.html>`_ and `DELETE <http://www.postgresql.org/docs/current/static/sql-delete.html>`_ commands. These commands will seamlessly apply across shards if necessary.
+You can update or delete rows from your distributed tables using the standard PostgreSQL `UPDATE <http://www.postgresql.org/docs/current/static/sql-update.html>`_ and `DELETE <http://www.postgresql.org/docs/current/static/sql-delete.html>`_ commands.
 
 ::
 
@@ -160,6 +160,22 @@ You can update or delete rows from your distributed tables using the standard Po
     UPDATE github_events
     SET event_public = TRUE
     WHERE org = 'happy_devs';
+
+When updates/deletes affect multiple shards as in the above example, Citus defaults to using a one-phase commit protocol. For greater safety you can enable two-phase commits by setting
+
+.. code-block:: postgresql
+
+  SET citus.multi_shard_commit_protocol = '2pc';
+
+If an update or delete affects only a single shard then it runs within a single worker node. In this case enabling 2PC is unnecessary. This often happens when updates or deletes filter by a table's distribution column:
+
+.. code-block:: postgresql
+
+  -- supposing github_events is distributed by user_id,
+  -- this will execute in a single worker node
+
+  DELETE FROM github_events
+  WHERE user_id = 120316;
 
 Maximizing Write Performance
 ----------------------------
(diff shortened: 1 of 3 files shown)


dist_tables/dml.rst
```diff
@@ -147,29 +147,19 @@ The situation changes when dealing with late arriving data, or running the rollu
...
---------------------------------
-
-The most flexible way to modify or delete rows throughout a Citus cluster is the master_modify_multiple_shards command. It takes a regular SQL statement as argument and runs it on all workers:
-
-::
-
-  SELECT master_modify_multiple_shards(
-    'DELETE FROM github_events WHERE repo_id IN (24509048, 24509049)');
-
-This uses a two-phase commit to remove or update data safely everywhere. Unlike the standard UPDATE statement, Citus allows it to operate on more than one shard. To learn more about the function, its arguments and its usage, please visit the :ref:`user_defined_functions` section of our documentation.
```
sumedhpathak: I think its worth calling out single-shard and multi-shard deletes. It would be useful to given an example of each, and note that multi-shard updates/deletes can use 2PC based on setting the configuration parameter for more safety (but by default is 1PC).

begriffs:
I updated the append distribution section, too, but I'm not sure that multi-shard deletes actually work with the append distribution. Here is what happens with my local citus which identifies itself as `7.1-3`:

```sql
DELETE FROM github_events
WHERE created_at >= '2015-01-01 00:03:00';
```
```
ERROR:  0A000: cannot run DELETE command which targets multiple shards
HINT:  Consider using an equality filter on partition column "created_at" to target a single shard. If you'd like to run a multi-shard operation, use master_modify_multiple_shards().
LOCATION:  PlanRouterQuery, multi_router_planner.c:1573
```

Also for the real-time use case, how about I update in the future pull request where I will modernize the section in general.

sumedhpathak:
@begriffs Interesting. I presume master_modify_multiple_shards works? Could you check that for append distribution?

begriffs:
Yeah, master modify is fine:

```sql
SELECT master_modify_multiple_shards($cmd$
  DELETE FROM github_events
  WHERE created_at >= '2015-01-01 00:03:00'
$cmd$);
```
```
┌───────────────────────────────┐
│ master_modify_multiple_shards │
├───────────────────────────────┤
│                         39157 │
└───────────────────────────────┘
```

begriffs:
Actually my local build from 9a04b789 is probably too old. I see that https://github.com/citusdata/citus/pull/1690 was just merged six days ago. Let me recompile from master and try again.

begriffs:
OK, false alarm it works fine with version `7.1-4`.

sharding/data_modeling.rst
```diff
@@ -137,12 +137,13 @@ We've seen some of the benefits of Citus for single-tenant queries, but it can a
     FROM ads GROUP BY account_id
   ORDER BY total_clicks DESC;
 
-There is even a way to run DML statements on multiple tenants. As long as the update statement references data local to its own tenant it can be applied simultaneously to all tenants with a helper UDF called :code:`master_modify_multiple_shards`. Here is an example of modifying all image urls to use secure connections.
+Citus is also able to seamlessly run DML statements on multiple tenants. As long as the update statement references data local to its own tenant it can be applied simultaneously to all tenants. Here is an example of modifying all image urls to use secure connections.
```
sumedhpathak: > As long as the update statement references data local to its own tenant it can be applied simultaneously to all tenants.

This line is a bit confusing, and took me a while to understand. It's not part of this PR, but I'd modify it to be a bit clearer. I *think* the intent is to say that an Update statement shouldn't be able to select from a different shard/tenant, and then update another one. I suspect we error out in these scenarios anyway, so I am also OK taking this statement out.

begriffs: Oops, just merged before this comment appeared.

PR #512, opened 2017-10-23
Extol the virtues of the rebalancer

### LINKED ISSUE #477: Make it clear that shard rebalancing happens "online"

We want people to know the capability of our rebalancer, to brag a little bit. Otherwise they might mistakenly believe it entails significant downtime.

Add detail about shard rebalancing under [cluster management](https://docs.citusdata.com/en/v7.0/admin_guide/cluster_management.html).

## How does this work? (devs)

Read queries continue without interruption. Write queries are queued on the coordinator until the shards they reference are in their new places.

* https://www.citusdata.com/blog/2017/04/11/rebalancing-your-database-with-citus/

begriffs:
Fixes #477

### DIFF
--- a/admin_guide/cluster_management.rst
+++ b/admin_guide/cluster_management.rst
@@ -39,12 +39,14 @@ If you want to move existing shards to a newly added worker, Citus Enterprise an
 
   SELECT rebalance_table_shards('github_events');
 
-Many products, like multi-tenant SaaS applications, cannot tolerate downtime, and Citus rebalancing is able to honor this requirement on PostgreSQL 10 or above. This means reads and writes from the application can continue without interruption while data is being moved.
+Many products, like multi-tenant SaaS applications, cannot tolerate downtime, and Citus rebalancing is able to honor this requirement on PostgreSQL 10 or above. This means reads and writes from the application can continue with minimal interruption while data is being moved.
 
 How it Works
 ~~~~~~~~~~~~
 
-Citus' shard rebalancing uses PostgreSQL logical replication to move data from the old shard (called the "publisher" in replication terms) to the new (the "subscriber.") As the PostgreSQL docs `explain <https://www.postgresql.org/docs/current/static/logical-replication-publication.html>`_, the source needs a *replica identity* configured:
+Citus' shard rebalancing uses PostgreSQL logical replication to move data from the old shard (called the "publisher" in replication terms) to the new (the "subscriber.") Logical replication allows application reads and writes to continue uninterrupted while copying shard data. Citus puts a brief write-lock on a shard only during the time it takes to update metadata to promote the subscriber shard as active.
+
+As the PostgreSQL docs `explain <https://www.postgresql.org/docs/current/static/logical-replication-publication.html>`_, the source needs a *replica identity* configured:
 
   A published table must have a "replica identity" configured in
   order to be able to replicate UPDATE and DELETE operations, so
@@ -102,7 +104,7 @@ This example would definitely fix the error with ``rebalance_table_shards`` but 
 
   While ``REPLICA IDENTITY USING INDEX`` is fine, we recommend **against** adding ``REPLICA IDENTITY FULL`` to a table. This setting would result in each update/delete doing a full-table-scan on the subscriber side to find the tuple with those rows. In our testing we’ve found this to result in worse performance than even solution four below.
 
-**Solution 3, force logical replication (on append-only tables)**
+**Solution 3, force logical replication (on insert-only tables)**
 
 If the distributed table doesn't have a primary key or replica identity, and adding one is unclear or undesirable, you can still force the use of logical replication on PostgreSQL 10 or above. It's OK to do this on a table which receives only reads and inserts (no deletes or updates). Include the optional ``shard_transfer_mode`` argument of ``rebalance_table_shards``:
 
--- a/admin_guide/cluster_management.rst
+++ b/admin_guide/cluster_management.rst
@@ -48,7 +48,7 @@ The first option is preferred because it is fully "online," meaning applications
 
 Logical replication requires PostgreSQL version 10 or above. Only tables with a primary key or `replica identity <https://www.postgresql.org/docs/current/static/sql-altertable.html#sql-createtable-replica-identity>`_ support update and delete operations during logical replication.
 
-By default ``rebalance_table_shards`` will use logical replication if Citus is running on PostgreSQL >=10 and the distributed table has a primary key or replica identity. Otherwise it will fall back to COPY with blocked writes.
+By default ``rebalance_table_shards`` will use logical replication if Citus is running on PostgreSQL >=10 and the distributed table has a primary key or replica identity. Otherwise it will error with a message about falling back to using COPY with blocked writes.
 
 .. code-block:: sql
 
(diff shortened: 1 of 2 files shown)


admin_guide/cluster_management.rst
```diff
@@ -34,6 +34,11 @@ In addition to the above, if you want to move existing shards to the newly added
 
 	select rebalance_table_shards('github_events');
 
+.. note::
+
+  Citus' shard rebalancing is fully "online," meaning applications communicating with the database do not experience downtime. Read queries to the cluster can continue without any interruption during the process.
```
lfittl: "Fully Online" contradicts "writes are queued up" in my opinion.

This will not be the case in future versions, but currently the rebalancer is not fully online (in my humble opinion).

cc @sumedhpathak

begriffs: Makes sense. Fully online for reading, but delayed for writing. I guess it will *accept* write queries, so in that way it's not like the system is down, but it's not "fully" up either.

sumedhpathak: So, my understanding may be wrong, but writes are *not* queued up for the majority of the time. Inserts can succeed, and applications can read their writes as well. Once we're close to caught up, then writes are blocked, and the cutover happens. @onderkalaci correct me if I'm wrong.

If there isn't a replica identity, then updates/deletes just error out, but inserts still go through.

admin_guide/cluster_management.rst
```diff
@@ -34,6 +34,11 @@ In addition to the above, if you want to move existing shards to the newly added
...
+
+  Even write queries are supported during rebalancing, but the coordinator node queues them until the shards they reference are in their new places. Although a full rebalance can take hours, it touches individual shards for only a few minutes each and subsequently unblocks writes to those shards.
```
sumedhpathak: See comment above, I think this is inaccurate, and will need updating.

sumedhpathak:
Given the amount of new content, maybe we just create a new section in Cluster Management titled Shard Rebalancing?

admin_guide/cluster_management.rst
lfittl: @sumedhpathak Just for full context, my comment was in regards to the old rebalancer. I believe we can claim fully online for the new one, with the slight caveat of a very short lock at the end (similar to how CREATE INDEX CONCURRENTLY operates as compared to regular CREATE INDEX)

admin_guide/cluster_management.rst
```diff
@@ -28,12 +28,62 @@ To add a new node to the cluster, you first need to add the DNS name or IP addre
...
 
-	select rebalance_table_shards('github_events');
+  SELECT rebalance_table_shards('github_events');
+
+Shard rebalancing is configurable to act by either of two methods:
+
+1. Logical replication or
+2. COPY with blocked writes
+
+The first option is preferred because it is fully "online," meaning applications communicating with the database do not experience downtime for reads or writes, except a very brief block on writes at the end of replication during a metadata update. The second option is online for reads only; writes are blocked while copying data from source to destination shard, which can take time.
```
lfittl: Any chance we can make "brief block on writes" an actual ballpark number? (i.e. "seconds"?) cc @sumedhpathak

admin_guide/cluster_management.rst
```diff
@@ -28,12 +28,62 @@ To add a new node to the cluster, you first need to add the DNS name or IP addre
...
+  ERROR: cannot use logical replication to transfer shards of the
+    relation test_table since it doesn't have a REPLICA IDENTITY or
+    PRIMARY KEY
+  DETAIL:  UPDATE and DELETE commands on the shard will error out during
+    logical replication unless there is a REPLICA IDENTIY or PRIMARY KEY.
+  HINT:  If you wish to continue without a replica identity set the
+    shard_transfer_mode to 'force_logical' or 'block_writes'.
+  */
+
+To allow replication behavior customization, ``rebalance_table_shards`` accepts a ``shard_transfer_mode`` argument with these possible values:
```
lfittl: We should mention that you can create a primary key, or set a replica identity, on the table to fix this error - maybe even with an example.

Only in rare cases would someone with a downtime-sensitive application (e.g. typical SaaS app) want to use any of the other replication behaviors. This is of course less of an issue for analytics/warehousing workloads, but I still think it'd be very good to push people to setup primary keys always.

admin_guide/cluster_management.rst
begriffs: @onderkalaci's pseudo code for the new rebalancing is:

1. Create the shard on the target node
1. Setup the logical replication for the shard from the source node to the target node
1. Wait until the shard on the target catches up with the source
1. Block modifications to the shard
1. Wait until the shard on the target catches up with the source after we’ve blocked the modifications
1. Drop the logical replication related setup
1. Drop the source shard
1. Update the metadata to reflect the shard move 
1. Unblock writes to the shard (the awaiting modifications now sees the new metadata)
1. Continue with the next shard

So we're blocking during operations 5-8, however long that takes.

admin_guide/cluster_management.rst
begriffs: I added a small note suggesting using pks when possible.

admin_guide/cluster_management.rst
lfittl: @begriffs I see - so might be a bit difficult to get a good measurement there, since I assume Step 5 in particular would have variable runtime depending on replication performance.

Maybe leave as is for now, I could see us getting better ballpark numbers on this over time, as we see customers using it.

admin_guide/cluster_management.rst
sumedhpathak: Instead of a note, can we offer the steps as Onder laid out in his document? The steps expand on the example by either:
a) Creating a Primary Key
b) Creating a Unique Index
c) Using the blocking rebalancer.

We also want to add a note around "Replica Identity Full", and note that we recommend *against* using this as it can be catastrophic for performance.

begriffs:
@sumedhpathak I made a new subsection in cluster management, and explained more about each way of making the rebalance function work.

admin_guide/cluster_management.rst
```diff
@@ -28,12 +28,105 @@ To add a new node to the cluster, you first need to add the DNS name or IP addre
...
+  ALTER TABLE test_table REPLICA IDENTITY
+    USING INDEX test_value_unique;
+
+This example would definitely fix the error with ``rebalance_table_shards`` but it's unrealistic insofar as the index includes every column in the table. In a wider table you want to restrict this index to as few columns as you can, which helps replication performance.
+
+.. note::
+
+  While ``REPLICA IDENTITY USING INDEX`` is fine, we recommend **against** adding ``REPLICA IDENTITY FULL`` to a table. This setting would result in each update/delete doing a full-table-scan on the subscriber side to find the tuple with those rows. In our testing we’ve found this to result in worse performance than even solution four below.
+
+**Solution 3, force logical replication (on append-only tables)**
```
sumedhpathak: Can we rename this to maybe "Insert-only"? We have used append-only tables in the context of append-partitioning, and so wanted to keep those a bit separate.

admin_guide/cluster_management.rst
```diff
@@ -28,12 +28,105 @@ To add a new node to the cluster, you first need to add the DNS name or IP addre
...
+-----------------------------------
+
+If you want to move existing shards to a newly added worker, Citus Enterprise and Citus Cloud provide a :ref:`rebalance_table_shards` function to make it easier. This function will move the shards of a given table to distribute them evenly among the workers.
 
 ::
 
-	select rebalance_table_shards('github_events');
+  SELECT rebalance_table_shards('github_events');
+
+Many products, like multi-tenant SaaS applications, cannot tolerate downtime, and Citus rebalancing is able to honor this requirement on PostgreSQL 10 or above. This means reads and writes from the application can continue without interruption while data is being moved.
```
sumedhpathak: We don't quite describe the logic on how we cutover. Instead of "without interruption", should we say with "minimal interruption"? In the "How it works" section, could we also expand a bit on the process, and that we lock for only a short time during the cutover?

PR #515, opened 2017-10-25
master_disable_node correction

### LINKED ISSUE #487: master_disable_node documentation is not correct
Currently, master_disable_node docs describe it as:

```
The master_disable_node function is the opposite of master_activate_node. It marks a node as inactive in the Citus metadata table pg_dist_node, removing it from the cluster temporarily. The function also deletes all reference table placements from the disabled node. To reactivate the node, just run master_activate_node again.
```

The node is not "removed" from the cluster temporarily. INSERTs and SELECTs can still use the node. It removes reference table copies from the node and avoids creation of new shards on that node.

begriffs:
Fixes #487

### DIFF
--- a/reference/user_defined_functions.rst
+++ b/reference/user_defined_functions.rst
@@ -430,11 +430,11 @@ master_disable_node
 $$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$
 
 The :code:`master_disable_node` function is the opposite of
-:code:`master_activate_node`. It marks a node as inactive in
-the Citus metadata table :code:`pg_dist_node`, removing it from
-the cluster temporarily. The function also deletes all reference table
-placements from the disabled node. To reactivate the node, just run
-:code:`master_activate_node` again.
+:code:`master_activate_node`. It marks a node as inactive in the Citus
+metadata table :code:`pg_dist_node`, removing copies of reference
+tables from the node and preventing the creation of new shards there. A
+disabled node is still available for reading and writing data, it is
+just temporarily ineligible for new shards.
 
 Arguments
 ************************


reference/user_defined_functions.rst
```diff
@@ -430,11 +430,11 @@ master_disable_node
...
-:code:`master_activate_node`. It marks a node as inactive in
-the Citus metadata table :code:`pg_dist_node`, removing it from
-the cluster temporarily. The function also deletes all reference table
-placements from the disabled node. To reactivate the node, just run
-:code:`master_activate_node` again.
+:code:`master_activate_node`. It marks a node as inactive in the Citus
+metadata table :code:`pg_dist_node`, removing copies of reference
+tables from the node and preventing the creation of new shards there. A
+disabled node is still available for reading and writing data, it is
+just temporarily ineligible for new shards.
```
sumedhpathak: This does appear to be the opposite of `master_activate_node`, so I'm not sure why this was named `disable` instead of `deactivate`. That said, any reason to remove this line:
>  To reactivate the node, just run`master_activate_node` again.

sumedhpathak: @samay-sharma It's not clear to me the node is available for reads/writes. See https://github.com/citusdata/citus/blob/632d0c675a59bfa3d052f993b8386c9243ca26ab/src/backend/distributed/executor/multi_task_tracker_executor.c#L204. I see similar calls in stuff like DDL pushdown. @begriffs Did you test what happens after disabling a node? It may be that we don't consistently use ActivePrimaryNodeList(), and thus support some operations, but not others.

begriffs: OK, tried it out on my single machine cluster with two worker nodes. I loaded the data from the MT tutorial, and also created a reference table and inserted a few rows into it.

Then I ran

```
❯ select master_disable_node('localhost', 5433);
NOTICE:  00000: Node localhost:5433 has active shard placements. Some queries may fail after this operation. Use SELECT master_activate_node('localhost', 5433) to activate this node back.
LOCATION:  master_disable_node, node_metadata.c:281
```

At this point I was still able to select * from the reference table (and a copy of the reference table remained on the de-activated worker). 

However trying to get all the rows of a distributed table did not work correctly:

```
❯ table campaigns;
ERROR:  failed to assign 16 task(s) to worker nodes
STATEMENT:  table campaigns;
ERROR:  XX000: failed to assign 16 task(s) to worker nodes
LOCATION:  GreedyAssignTaskList, multi_physical_planner.c:4881
```

Citus was able to return results with the router executor for a value of the distribution column held on the still-active worker:

```
select * from campaigns where company_id = 1;
-- this works
```

begriffs: Proposed changes to my description based on this experiment:

> removing copies of reference tables from the node

I didn't observe that, so could remove that claim.

> A disabled node is still available for reading and writing data

True for the router executor, however the real-time executor seems to get completely messed up. Not just for queries involving campaigns, but for any query. At least that's what I'm seeing on my little two-node cluster.

So `master_disable_node` seems potentially disruptive. Not sure how we want to express this in the docs.

sumedhpathak:
@begriffs Should we close this issue? We have an open issue in Citus repo to think about what our implementation should be: https://github.com/citusdata/citus/issues/2063.

begriffs:
Sure, since the changes in this PR aren't accurate, and the implementation may be changing we can close this PR.

PR #516, opened 2017-10-26
adding section for useful queries

craigkerstiens:


reference/useful_queries.rst
```diff
@@ -0,0 +1,109 @@
...
+
+Identifying unused indexes
+--------------------------
+
+This query will run across all worker nodes and identify any unused indexes for a given distributed table, designated here with the placeholder :code:`my_distributed_table`:
+
+.. code-block:: sql
+
+  SELECT *
+  FROM run_command_on_shards('my_distributed_table', $cmd$
```
begriffs: I adjusted this query so that it now runs. However I can't tell whether it works correctly. It indicated that my distributed table had no unused indices. I then created an index that seemed unused and re-ran the query but no change. Perhaps `pg_stat_user_indexes` had not caught up yet...

index.rst
```diff
@@ -146,6 +146,7 @@ Welcome to the documentation for Citus 7.0! Citus horizontally scales PostgreSQL
 
    reference/citus_sql_reference.rst
    reference/sql_workarounds.rst
+   reference/useful_queries.rst
```
sumedhpathak: This section probably belongs in the Administration section, rather than SQL Reference.

reference/useful_queries.rst
```diff
@@ -0,0 +1,117 @@
...
+There are a number of queries that can be useful in monitoring your Postgres cluster. Within Citus many of those useful queries which provide things like cache hit rate or locks are still valid but may need to be adapted to run across all your nodes. This section contains a set of useful utility queries for running against your Citus cluster.
+
+Detecting locks
+---------------
+
+This query will run across all worker nodes and identify locks, how long they've been open, and the offending query:
+
+.. code-block:: sql
+
+  SELECT run_command_on_workers($cmd$
```
sumedhpathak: Can we also provide sample output for these commands?

reference/useful_queries.rst
sumedhpathak: @begriffs What changes did you make to the query? Could you try rerunning the query to see if the results are now different?

admin_guide/cluster_management.rst
```diff
@@ -267,3 +267,162 @@ Output:
...
+       AND blocking_locks.objsubid IS NOT DISTINCT FROM blocked_locks.objsubid
+       AND blocking_locks.pid != blocked_locks.pid
+      JOIN pg_catalog.pg_stat_activity blocking_activity ON blocking_activity.pid = blocking_locks.pid
+      WHERE NOT blocked_locks.GRANTED
+       AND blocking_locks.GRANTED
+      GROUP BY blocked_activity.query,
+               blocking_activity.query
+      ORDER BY 4
+    ) a
+  $cmd$);
```
begriffs: This lock-finding query is the only one I had trouble getting output for. On the coordinator I run:

```
begin;
LOCK TABLE http_request in row share mode;
```

Then in another psql session I run the lock detecting query but it says

```
┌───────────────────────────────────────────────────────────────┐
│                    run_command_on_workers                     │
├───────────────────────────────────────────────────────────────┤
│ (ec2-13-59-96-221.us-east-2.compute.amazonaws.com,5432,t,"")  │
│ (ec2-52-14-226-167.us-east-2.compute.amazonaws.com,5432,t,"") │
└───────────────────────────────────────────────────────────────┘
```
meaning that no locks are active.

begriffs: OK, I was able to get output by running two sessions on a worker, one doing a select for update and another a regular update. That causes one statement to block on another.

begriffs:
@sumedhpathak this takes care of your two feedback items. It's ready for another look.

admin_guide/cluster_management.rst
```diff
@@ -267,3 +267,174 @@ Output:
...
+Querying the size of your shards
+--------------------------------
+
+This query will provide you with the size of every shard of a given distributed table, designated here with the placeholder :code:`my_distributed_table`:
+
+.. code-block:: postgresql
+
+  SELECT pg_size_pretty(result::bigint)
+  FROM run_command_on_shards('my_distributed_table', $cmd$
+    SELECT pg_table_size('%s');
```
sumedhpathak: Can we amend this to show the shard-name (table-name) as well? It might be more useful to see which shard is particularly large or small. I'm guessing the next step in diagnosis will be to anyway get the shard-name.

admin_guide/cluster_management.rst
```diff
@@ -267,3 +267,174 @@ Output:
...
+
+Example output:
+
+::
+
+  ┌─────────┬─────────┬───────────────────────────────────────────────────────────────────────┐
+  │ shardid │ success │                            result                                     │
+  ├─────────┼─────────┼───────────────────────────────────────────────────────────────────────┤
+  │  102008 │ t       │                                                                       │
+  │  102009 │ t       │ {"public.my_distributed_table_102009##stupid_index_102009##28 MB##0"} │
```
sumedhpathak: Is there a reason this only shows unused on one shard? I'm not sure what would be actionable here, since the user can't selectively drop (without manual effort) an index on only one shard of a table. 

Also, what does the 28MB and the 0 represent at the end?

begriffs: Maybe it's actionable in that if the index is well utilized on most shards then keep it, but if used on few (or one) shards then you could consider dropping it.

Looks like `28MB##0` means the index takes up 28MB, and and has been used in an "index scan" zero times. I could modify the query with `json_build_object` so that the result columns have names rather than delimiting with `##`.

sumedhpathak: > I could modify the query with json_build_object so that the result columns have names rather than delimiting with ##.

Could you do that? It might make the output more understandable.

saicitus:
A couple of other useful queries which could be added to the doc:

**Replication factor of distributed tables:**
```
SELECT logicalrelid AS tablename,
       count(*)/count(DISTINCT ps.shardid) AS replication_factor
FROM pg_dist_shard_placement ps
JOIN pg_dist_shard p ON ps.shardid=p.shardid
GROUP BY logicalrelid;
```
**Size of all distributed tables:** (Works only when the replication_factor for all the tables is 1)
```
SELECT
  tablename,
  pg_size_pretty(citus_total_relation_size(tablename ::text))as total_size
FROM pg_tables pt
JOIN pg_dist_partition pp
  ON pt.tablename ::text = pp.logicalrelid ::text
WHERE schemaname = 'public';
```

PR #517, opened 2017-10-26
Show subquery window function workaround

### LINKED ISSUE #475: Window functions on distribution col now supported
Update https://docs.citusdata.com/en/v7.0/reference/sql_workarounds.html#data-warehousing-queries and anywhere else they are mentioned

begriffs:
Fixes #475

### DIFF
--- a/reference/sql_workarounds.rst
+++ b/reference/sql_workarounds.rst
@@ -129,10 +129,10 @@ replicate that shard to every node in the cluster. Now Citus will accept a join 
 Window Functions
 ----------------
 
-Currently Citus does not have out-of-the-box support for window functions on queries involving distributed tables, but there is a straightforward workaround. Window functions will work on a distributed table if
+Currently Citus does not have out-of-the-box support for window functions on cross-shard queries, but there is a straightforward workaround. Window functions will work across shards on a distributed table if
 
-1. They are in a subquery and
-2. They :code:`partition by` the distribution column of a distributed table
+1. The window function is in a subquery and
+2. It includes a :code:`PARTITION BY` clause containing the table's distribution column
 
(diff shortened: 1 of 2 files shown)


reference/sql_workarounds.rst
```diff
@@ -124,41 +124,74 @@ There is a workaround: you can replicate the local table to a single shard on ev
 This will create a table with a single shard (non-distributed), but will
 replicate that shard to every node in the cluster. Now Citus will accept a join query between *here* and *there*, and each worker will have all the information it needs to work efficiently.
 
+.. _window_func_workaround:
+
+Window Functions
+----------------
+
+Currently Citus does not have out-of-the-box support for window functions on queries involving distributed tables, but there is a straightforward workaround. Window functions will work on a distributed table if
```
sumedhpathak: So this is true only for cross-shard queries. For multi-tenant cases where we hit only a single tenant, this would work. Could you update this comment to reflect this in a stronger manner?

sumedhpathak: I realize now we have a comment up top in the section saying we support full SQL for multi-tenant workloads, but this statement still is quite misleading: "Currently Citus does not have out-of-the-box support for window functions". 
We could also rename this as "Cross-shard queries with Window Functions" or something similar to drive home the point.

reference/sql_workarounds.rst
```diff
@@ -124,41 +124,74 @@ There is a workaround: you can replicate the local table to a single shard on ev
...
+
+1. They are in a subquery and
+2. They :code:`partition by` the distribution column of a distributed table
```
sumedhpathak: What part of the window function is that? Can we be a bit more specific by categorizing this class of window functions? (Window functions which include the partition-by clause?

PR #526, opened 2017-11-03
Modernize the real-time guide

### LINKED ISSUE #469: Update the real-time use case guide
The guide https://docs.citusdata.com/en/v7.0/use_case_guide/real_time_dashboards.html is way outdated. Using the latest Citus features will make it a lot simpler.

begriffs:
Uses the Citus 7.1 features to make this guide a lot simpler.

Fixes #469

### DIFF
--- a/use_case_guide/real_time_dashboards.rst
+++ b/use_case_guide/real_time_dashboards.rst
@@ -21,14 +21,6 @@ volume is also high, and both historical and live data are important.
 In this section we'll demonstrate how to build part of the first example, but this
 architecture would work equally well for the second and many other use-cases.
 
-Running It Yourself
--------------------
-
-There will be code snippets in this tutorial but they don't specify a complete system.
-There's `a github repo <https://github.com/citusdata/realtime-dashboards-resources>`_ with
-all the details in one place. If you've followed our installation instructions for running
-Citus on either a single or multiple machines you're ready to try it out.
-
 Data Model
 ----------
 
@@ -71,6 +63,8 @@ data across your cluster after adding new worker nodes.
 
   Citus Cloud uses `streaming replication <https://www.postgresql.org/docs/current/static/warm-standby.html>`_ to achieve high availability and thus maintaining shard replicas would be redundant. In any production environment where streaming replication is unavailable, you should set ``citus.shard_replication_factor`` to 2 or higher for fault tolerance.
 
+With this, the system is ready to accept data and serve queries! Keep the following loop running in a ``psql`` console in the background while you continue with the other commands in this article. It generates fake data every second or two.
+
 .. code-block:: postgres
 
   DO $$
@@ -86,11 +80,7 @@ data across your cluster after adding new worker nodes.
     END LOOP;
   END $$;
 
-With this, the system is ready to accept data and serve queries! We've provided `a data
-ingest script
-<https://github.com/citusdata/realtime-dashboards-resources/blob/master/ingest_example_data.sql>`_
-you can run to generate example data. Once you've ingested data, you can run dashboard
-queries such as:
+Once you're ingesting data, you can run dashboard queries such as:
 
 .. code-block:: sql
 
@@ -154,9 +144,9 @@ and Citus will place matching shards on the same worker. This is called
 .. image:: /images/colocation.png
   :alt: co-location in citus
 
-In order to populate ``http_request_1min`` we're going to periodically run the equivalent
-of an INSERT INTO SELECT. We'll run a function on all the workers which runs INSERT INTO SELECT
-on every matching pair of shards. This is possible because the tables are co-located.
+In order to populate ``http_request_1min`` we're going to periodically run
+an INSERT INTO SELECT. This is possible because the tables are co-located.
+The following function wraps the rollup query up for convenience.
 
 .. code-block:: plpgsql
 
@@ -189,14 +179,6 @@ on every matching pair of shards. This is possible because the tables are co-loc
   END;
   $$ LANGUAGE plpgsql;
 
-Inside this function you can see the dashboard query from earlier. It's been wrapped in
-some machinery which writes the results into ``http_request_1min`` and allows passing in
-the name of the shards to read and write from. It also takes out an advisory lock, to
-ensure there aren't any concurrency bugs where the same rows are written multiple times.
-
-Using that metadata, every minute it runs a script which calls ``rollup_1min`` once for
-each pair of shards:
-
 .. note::
 
   The above function should be called every minute. You could do this by
@@ -205,6 +187,9 @@ each pair of shards:
   .. code-block:: bash
 
     * * * * * psql -c 'SELECT rollup_http_request();'
+
+  Alternately, an extension such as `pg_cron <https://github.com/citusdata/pg_cron>`_
+  allows you to schedule recurring queries directly from the database.
 
 The dashboard query from earlier is now a lot nicer:
 
@@ -223,9 +208,9 @@ Expiring Old Data
 The rollups make queries faster, but we still need to expire old data to avoid unbounded
 storage costs. Once you decide how long you'd like to keep data for each granularity, you
 could easily write a function to expire old data. In the following example, we decided to
-keep raw data for one day and 1-minute aggregations for one month.
+keep raw data for one day, and per-minute aggregations for one month.
 
-.. code-block:: postgres
+.. code-block:: plpgsql
 
   CREATE OR REPLACE FUNCTION expire_old_request_data() RETURNS void
   AS $$
@@ -242,7 +227,7 @@ keep raw data for one day and 1-minute aggregations for one month.
 
     * * * * * psql -c "SELECT expire_old_request_data();"
 
-That's the basic architecture! We provided an architecture that ingests HTTP events and
+Those are the basics! We provided an architecture that ingests HTTP events and
 then rolls up these events into their pre-aggregated form. This way, you can both store
 raw events and also power your analytical dashboards with subsecond queries.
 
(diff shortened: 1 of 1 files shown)


use_case_guide/real_time_dashboards.rst
```diff
@@ -111,15 +117,15 @@ for each of the last 30 days.
...
-        CHECK (ingest_time = date_trunc('minute', ingest_time))
+    site_id INT,
+    ingest_time TIMESTAMPTZ, -- which minute this row represents
+
+    error_count INT,
+    success_count INT,
+    request_count INT,
+    average_response_time_msec INT,
+    CHECK (request_count = error_count + success_count),
+    CHECK (ingest_time = date_trunc('minute', ingest_time))
```
begriffs: By the way, hats off to whoever thought of these CHECK constraints. Great use for them. 😄

use_case_guide/real_time_dashboards.rst
```diff
@@ -282,74 +237,87 @@ which often appear.
...
-  WHERE date_trunc('minute', ingest_time) = date_trunc('minute', now())
-  GROUP BY site_id, minute;
+  ALTER TABLE http_request_1min ADD COLUMN distinct_ip_addresses hll;
+
+Next use our custom aggregation to populate the column. Just add it
+to the query in our rollup function:
+
+.. code-block:: diff
+
+  @@ -1,10 +1,12 @@
```
begriffs: I thought the diff format makes it easier to understand how small the necessary change is. Our old explanation was like, "See this query? Imagine melding it into the rollup."

The drawback of the diff is that it doesn't provide anything to directly copy and run.

begriffs:
There's still an unresolved issue. The jsonb section says to add a clause like this in our rollup function:

```sql
SELECT
  site_id, minute,
  jsonb_object_agg(request_country, country_count)
FROM (
  SELECT
    site_id, date_trunc('minute', ingest_time) AS minute,
    request_country,
    count(1) AS country_count
  FROM http_request
  GROUP BY site_id, minute, request_country
) AS h
GROUP BY site_id, minute;
```

However when I tried it, I ran into a problem where the rollup function needs to read fields like `status_code` which would get obscured by the inner GROUP BY.

use_case_guide/real_time_dashboards.rst
```diff
@@ -21,14 +21,6 @@ volume is also high, and both historical and live data are important.
 In this section we'll demonstrate how to build part of the first example, but this
 architecture would work equally well for the second and many other use-cases.
 
-Running It Yourself
--------------------
-
-There will be code snippets in this tutorial but they don't specify a complete system.
```
begriffs: Now that we aren't doing complicated stuff in the shell I removed the link to the github repo. The guide seems pretty self-contained now.

use_case_guide/real_time_dashboards.rst
```diff
@@ -71,11 +63,24 @@ data across your cluster after adding new worker nodes.
 
   Citus Cloud uses `streaming replication <https://www.postgresql.org/docs/current/static/warm-standby.html>`_ to achieve high availability and thus maintaining shard replicas would be redundant. In any production environment where streaming replication is unavailable, you should set ``citus.shard_replication_factor`` to 2 or higher for fault tolerance.
 
-With this, the system is ready to accept data and serve queries! We've provided `a data
-ingest script
-<https://github.com/citusdata/realtime-dashboards-resources/blob/master/ingest_example_data.sql>`_
```
begriffs: Not sure why we used to link to an external file. I inlined the sql in the guide instead.

use_case_guide/real_time_dashboards.rst
```diff
@@ -138,141 +144,90 @@ and Citus will place matching shards on the same worker. This is called
...
-  as simple as this:
+  The above function should be called every minute. You could do this by
+  adding a crontab entry on the coordinator node:
 
   .. code-block:: bash
 
-     * * * * * /some/path/run_rollups.sh
+    * * * * * psql -c 'SELECT rollup_http_request();'
+
+  Alternately, an extension such as `pg_cron <https://github.com/citusdata/pg_cron>`_
```
begriffs: Thought I'd slip this in here. 😉

use_case_guide/real_time_dashboards.rst
```diff
@@ -362,6 +330,8 @@ aggregate function and its semantics. You do this by running the following:
 .. code-block:: sql
 
   -- this should be run on the workers and coordinator
+  -- (not necessary on Citus Cloud)
```
sumedhpathak: Do we know for sure this is not necessary on Cloud? I am not sure if we automatically create the sum on hll types on Cloud formations.

use_case_guide/real_time_dashboards.rst
```diff
@@ -282,74 +237,87 @@ which often appear.
...
-  CREATE AGGREGATE sum(hll)(
-      sfunc = hll_union_trans,
-      stype = internal,
-      finalfunc = hll_pack
-  );
+.. note::
 
-When doing our rollups, we can now aggregate sessions into an hll column with queries
-like this:
+  This is not necessary on Citus Cloud, which has HLL already installed,
```
sumedhpathak: Again, not sure if the sum(hll_add_trans0) is already defined for us on Citus Cloud. Have you verified this?

use_case_guide/real_time_dashboards.rst
```diff
@@ -282,74 +237,87 @@ which often appear.
...
   CREATE AGGREGATE sum(hll_hashval) (
-       SFUNC = hll_add_trans0,
-       STYPE = internal,
-       FINALFUNC = hll_pack
+    SFUNC = hll_add_trans0,
+    STYPE = internal,
+    FINALFUNC = hll_pack
   );
 
-  -- alias for hll_union_agg
```
sumedhpathak: Did we not need this here for the first example?

use_case_guide/real_time_dashboards.rst
```diff
@@ -138,141 +144,90 @@ and Citus will place matching shards on the same worker. This is called
...
-
-   QUERY=$(cat <<END
-     SELECT * FROM colocated_shard_placements(
-       'http_request'::regclass, 'http_request_1min'::regclass
-     );
-   END
-   )
-
-   COMMAND="psql -h \$2 -p \$3 -c \"SELECT rollup_1min('\$0', '\$1')\""
+  CREATE OR REPLACE FUNCTION rollup_http_request() RETURNS void AS $$
```
sumedhpathak: Is there a reason to put this in a function? Previously the function simplified some work around computing co-located placements, and calling the aggregation script for each pair. Now the function adds complexity to what is a very simple query. We can just note that for *actual* production use, we should consider wrapping this in a function and putting it in cron. Thoughts?

use_case_guide/real_time_dashboards.rst
```diff
@@ -138,141 +144,90 @@ and Citus will place matching shards on the same worker. This is called
...
 The rollups make queries faster, but we still need to expire old data to avoid unbounded
 storage costs. Once you decide how long you'd like to keep data for each granularity, you
 could easily write a function to expire old data. In the following example, we decided to
-keep raw data for one day and 1-minute aggregations for one month.
+keep raw data for one day, and per-minute aggregations for one month.
 
 .. code-block:: plpgsql
 
-  -- another function for the coordinator
   CREATE OR REPLACE FUNCTION expire_old_request_data() RETURNS void
```
sumedhpathak: Same note on whether we need to express this as a function. I think just showing a simple DELETE query would be nice, and then noting that in a production environment to cron this.

use_case_guide/real_time_dashboards.rst
begriffs: In `rollup_http_request` I'm setting a variable `max_rollup_time` in one step, and using the variable in an INSERT statement in another. I had trouble trying to do it all in a single query.

use_case_guide/real_time_dashboards.rst
begriffs: Yep, Cloud provides it. I ran through the this guide again to check.

use_case_guide/real_time_dashboards.rst
begriffs: On Cloud it understood `SUM(distinct_ip_addresses)` without my having to create that aggregate.

use_case_guide/real_time_dashboards.rst
begriffs: The modification to `http_request_1min` in this example uses `SUM(hll_hash_text(ip_address))`. But I've been testing only on Cloud which has done all this stuff for me, so I'll check this guide on a local cluster and verify exactly when we need to create custom aggregates.

begriffs: Verified locally that we do need the custom aggregate.

Also discovered it must be created on all nodes, not just the coordinator, and updated the comment in the code block.

begriffs:
@sumedhpathak I've revised things based on your comments. The remaining thing to look at is what to do about the "Unstructured Data with JSONB" section. There's a comment about that above.

sumedhpathak:
``` SQL
INSERT INTO http_request_1min (
    site_id, ingest_time, 
    country_counters, <---- Insertion column
    request_count,
    success_count, error_count, average_response_time_msec
  ) SELECT
    site_id,
    minute,
    jsonb_objectagg(request_country, country_count), <------- /* HERE */
    COUNT(1) as request_count,
    SUM(CASE WHEN (status_code between 200 and 299) THEN 1 ELSE 0 END) as success_count,
    SUM(CASE WHEN (status_code between 200 and 299) THEN 0 ELSE 1 END) as error_count,
    SUM(response_time_msec) / COUNT(1) AS average_response_time_msec
  FROM (
    SELECT *,
      date_trunc('minute', ingest_time) AS minute
    FROM http_request
  ) AS h
  WHERE minute > max_rollup_time
    AND minute < date_trunc('minute', now())
  GROUP BY site_id, minute
  ORDER BY minute ASC;
```

On your jsbonb question, isn't the above where we'd want to put in the json object aggregation? The group-by on site_id and minute is in the same level as our jsonb object aggregation right?

use_case_guide/real_time_dashboards.rst
begriffs: To test, just now I created a new dev formation. It has hll 2.10 installed. Then:

```
❯ \da
                                                             List of aggregate functions
┌────────┬────────────────┬──────────────────────────────┬────────────────────────────────────────────────┬─────────────────────────────────────────┐
│ Schema │      Name      │       Result data type       │              Argument data types               │               Description               │
├────────┼────────────────┼──────────────────────────────┼────────────────────────────────────────────────┼─────────────────────────────────────────┤
│ public │ hll_add_agg    │ hll                          │ hll_hashval                                    │                                         │
│ public │ hll_add_agg    │ hll                          │ hll_hashval, integer                           │                                         │
│ public │ hll_add_agg    │ hll                          │ hll_hashval, integer, integer                  │                                         │
│ public │ hll_add_agg    │ hll                          │ hll_hashval, integer, integer, bigint          │                                         │
│ public │ hll_add_agg    │ hll                          │ hll_hashval, integer, integer, bigint, integer │                                         │
│ public │ hll_union_agg  │ hll                          │ hll                                            │                                         │
│ public │ max            │ citext                       │ citext                                         │                                         │
│ public │ min            │ citext                       │ citext                                         │                                         │
│ public │ sum            │ group_elements_size_result[] │ group_elements_size_result[]                   │                                         │
│ public │ sum            │ hll                          │ hll                                            │                                         │
│ public │ sum            │ hll                          │ hll_hashval                                    │                                         │
│ public │ sum            │ integer[]                    │ integer[]                                      │                                         │
│ public │ topn_add_agg   │ jsonb                        │ text                                           │ aggregate the items into one counter    │
│ public │ topn_union_agg │ jsonb                        │ jsonb                                          │ aggregate the counters into one counter │
└────────┴────────────────┴──────────────────────────────┴────────────────────────────────────────────────┴─────────────────────────────────────────┘
```

begriffs:
Well, the `jsonb_objectagg` needs access to `country_count`, which would be provided by a modified `h` subquery. The old realtime use case guide suggested adding a group-by to h, like:

```sql
FROM (
  SELECT
    site_id, date_trunc('minute', ingest_time) AS minute,
    request_country,
    count(1) AS country_count  /* <--- needed for jsonb_objectagg */
  FROM http_request
  GROUP BY site_id, minute, request_country
  /* ^---- but this group-by obscures columns like status_code */
) AS h
```

use_case_guide/real_time_dashboards.rst
```diff
@@ -138,141 +154,77 @@ and Citus will place matching shards on the same worker. This is called
...
-   END
-   )
-
-   COMMAND="psql -h \$2 -p \$3 -c \"SELECT rollup_1min('\$0', '\$1')\""
-
-   psql -tA -F" " -c "$QUERY" | xargs -P32 -n4 sh -c "$COMMAND"
+  CREATE OR REPLACE FUNCTION rollup_http_request() RETURNS void AS $$
+  DECLARE max_rollup_time TIMESTAMPTZ;
+  BEGIN
+    SELECT COALESCE(max(ingest_time), timestamp '10-10-1901')
```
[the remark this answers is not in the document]
begriffs: If I take the insert…select out of the function and use a subquery directly (rather than `max_rollup_time`)

```sql
WHERE minute > (
    SELECT COALESCE(max(ingest_time), timestamp '10-10-1901')
    FROM http_request_1min
  )
  AND minute < date_trunc('minute', now())
```

Then it gives me an error

```
ERROR:  0A000: cannot perform distributed planning for the given modification
DETAIL:  Select query cannot be pushed down to the worker.
LOCATION:  RouterModifyTaskForShardInterval, insert_select_planner.c:548
```

I think it's because we're trying to find the max over the whole `http_request_1min` table which is inaccessible from down in a worker.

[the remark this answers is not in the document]
begriffs: Just tried another approach: calculating the max for each tenant rather than globally. Should still process only the new data on each tenant.

```sql
  WHERE minute > (
      SELECT COALESCE(max(ingest_time), timestamp '10-10-1901')
      FROM http_request_1min
      WHERE http_request_1min.site_id = h.site_id
    )
    AND minute < date_trunc('minute', now())
```

This allows the insert…select query to run. What do you think, should I use this instead?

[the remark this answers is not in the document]
sumedhpathak: As we discussed yesterday we are going ahead with leaving the rollup in a function. Then we don't need to simplify and make this one query. So we can retain what we have now.

begriffs:
OK, I've come up with a diff to the rollup that includes the country counting jsonb. Ready for review.

sumedhpathak:
Minor. The 3 queries hitting the rollup table are slightly different:
Two of them use 5 minute intervals, while one uses 1-minute. We use different ways to cast them to 'interval'. Finally, one is >= while the others are only >.
Could we at a minimum update the last one to 5 or 10 minute intervals? Running this for 1-minute interval gave me no data (since we now generate it randomly, its possible we don't ingest USA data in that minute I guess).
 
Original:
``` SQL
SELECT site_id, ingest_time as minute, request_count,
success_count, error_count, average_response_time_msec 
FROM http_request_1min 
WHERE ingest_time > date_trunc('minute', now()) - interval '5 minutes';
```
For HLL:
``` SQL
SELECT hll_cardinality(SUM(distinct_ip_addresses))
FROM http_request_1min
WHERE ingest_time > now() - '5 minutes'::interval;
```

For JSONB:
``` SQL
SELECT
  request_count, success_count, error_count, average_response_time_msec,
  COALESCE(country_counters->>'USA', '0')::int AS american_visitors
FROM http_request_1min
WHERE ingest_time >= date_trunc('minute', now()) - '1 minute'::interval;
```

begriffs:
OK, made the adjustments.

> I presume you've verified that the jsonb query generates the correct expected data?

Yeah it looks right to me -- the visits by country always total `request_count` for each row.

> Did you try it with UPSERT's like we had discussed btw?

I did, but overall it seemed more complicated than just generating the data properly in the initial insert. This is what I had:

```sql
INSERT INTO http_request_1min (
  site_id, ingest_time, country_counters
) SELECT
  site_id, minute,
  jsonb_object_agg(request_country, country_count)
FROM (
  SELECT
    site_id, date_trunc('minute', ingest_time) AS minute,
    request_country,
    count(1) AS country_count
  FROM http_request
  GROUP BY site_id, minute, request_country
) AS h
WHERE minute > max_rollup_time
  AND minute < date_trunc('minute', now())
GROUP BY site_id, minute
ON CONFLICT (site_id, ingest_time) DO UPDATE
  SET country_counters = EXCLUDED.country_counters;
```

Issue #528, opened 2017-11-07
General 7.1 Release Checklist


begriffs:
- [x] s/7\\.0/7.1/g
- [x] Update all installation steps to use new package name (except CloudFormation)
- [x] Update links to new CloudFormation template
- [x] Verify Installation steps with new packages
    - [x] Ubuntu 16.04 (Joe)
    - [x] Ubuntu 14.04
    - [x] Debian Jessie
    - [x] Debian Wheezy
    - [x] Amazon Linux 2017.03.0
    - [x] CentOS 7.3
    - [x] CentOS 6.9
    - [x] RHEL 7.2 
    - [x] RHEL 6.5 
    - [x] Fedora 26
    - [ ] Fedora 25
- [x] Test 7.0 -> 7.1 upgrade instructions
- [x] Test 7.1, pg 9.6 -> pg 10 upgrade instructions
- [x] Test Multi-tenant tutorial and use-case guide (on just one version)
- [x] Test Real-time analytics tutorial and use-case guide (on just one version)
- [x] Test Docker installation. 
- [x] Update citusdata.com/download links.
- [x] Create new 7.1 release branch and set as default ver in RTD (last step)

begriffs:
Done.

PR #541, opened 2017-11-15
Update Citus version to 7.1 throughout text

begriffs:
For #528


cloud/getting_started.rst
```diff
@@ -112,7 +112,7 @@ Generally there is no need to install extensions, however, because every Cloud c
 +--------------------+---------+------------+--------------------------------------------------------------------+
 | citext             | 1.3     | public     | data type for case-insensitive character strings                   |
 +--------------------+---------+------------+--------------------------------------------------------------------+
-| citus              | 7.0-15  | pg_catalog | Citus distributed database                                         |
+| citus              | 7.1-4   | pg_catalog | Citus distributed database                                         |
```
begriffs: This is speculative. Assuming Cloud will show this version soon.

cloud/getting_started.rst
```diff
@@ -150,7 +150,7 @@ Generally there is no need to install extensions, however, because every Cloud c
 +--------------------+---------+------------+--------------------------------------------------------------------+
 | session_analytics  | 1.0     | public     |                                                                    |
 +--------------------+---------+------------+--------------------------------------------------------------------+
-| shard_rebalancer   | 7.0     | public     |                                                                    |
+| shard_rebalancer   | 7.1     | public     |                                                                    |
```
begriffs: Also guessing here

conf.py
```diff
@@ -57,9 +57,9 @@
 # built documents.
 #
 # The short X.Y version.
-version = '7.0'
+version = '7.1'
 # The full version, including alpha/beta/rc tags.
-release = '7.0.3'
```
begriffs: Dropping the patch version here because I don't foresee docs ever showing different content across patch versions.

sumedhpathak: What is this flag used for? Do we expect any issues dropping this precision?

begriffs: Read The Docs makes versions from these numbers. Someone sent a PR a few months ago and bumped this field to "6.2.2." Well unknown to me all the new docs content was being rendered in that version even though v6.2 was the default version visitors would see when they went to docs.citusdata.com. So for a while the newest changes appeared to not be reflecting on the docs. I noticed this extra version in the docs selector in the lower left, then marked 6.2.2 as private and restored the version to 6.2.0.

I'm removing the extra digit so that nobody is tempted to change it and accidentally create a new docs version.

begriffs: Sphinx docs distinguish between `version` and `release` as:

___

### version

The major project version, used as the replacement for |version|. For example, for the Python documentation, this may be something like 2.6.

### release

The full project version, used as the replacement for |release| and e.g. in the HTML templates. For example, for the Python documentation, this may be something like 2.6.0rc1.

If you don’t need the separation provided between version and release, just set them both to the same value.

PR #543, opened 2017-11-16
Improve usability of upgrade steps

velioglu:
Improvements on the usability of upgrade steps. 

If user either start the server before upgrading process or run any query before metadata restore, background daemon may wake up and log chunk of error/warning messages.


sumedhpathak:
The warnings seem innocuous. @begriffs Could you run through those steps once, and if the warnings sound good, hit merge? (And also cherry-pick to 7.1).

PR #544, opened 2017-11-21
Describe our support for views on dist tables

### LINKED ISSUE #227: How to create view containing distributed query

begriffs:
Fixes #227 

This is a replacement PR for another that spoke more to workarounds than support. This PR is more straightforward.

### DIFF
--- a/dist_tables/querying.rst
+++ b/dist_tables/querying.rst
@@ -122,62 +122,11 @@ In general, co-located joins are more efficient than repartition joins as repart
 Views on Distributed Tables
 ###########################
 
-At execution time PostgreSQL internally rewrites queries on views, inlining the view definition as a subquery. To see this in action, assume we have a game database with a distributed table called ``scores`` containing each player's name and the points they earned in runs through the game. For convenience we can make a view to get the average score by player:
+Citus supports all views on distributed tables. For an overview of views' syntax and features, see the PostgreSQL documentation for `CREATE VIEW <https://www.postgresql.org/docs/current/static/sql-createview.html>`_.
 
-.. code-block:: sql
+Note that some views cause a less efficient query plan than others. For more about detecting and improving poor view performance, see :ref:`subquery_perf`. (Views are treated internally as subqueries.)
 
-  CREATE VIEW avg_scores AS
-  SELECT player, avg(points) AS player_avg
-  FROM scores
-  GROUP BY player;
-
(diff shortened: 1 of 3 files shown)


sumedhpathak:
Also, this section probably changes in light of the new recursive planning changes. AFAICS we now support most views I think.

marcocitus:
Yes, tenant ID filters are not really necessary. 

One caveat is that views on distributed tables cannot be used to grant limited read access to another user.

begriffs:
How about we merge this one for 7.1, and let the 7.2 PR take care of the differences? I'd like to get these old PRs wrapped up.

sumedhpathak:
@begriffs Just saw your note. Outside of the 7.2 changes, I was recommending adding a section for Materialized Views. Given that we have 7.2 launched, we should probably just get this updated with that in mind?

begriffs:
Sounds good.

begriffs:
This is ready for another look. It's about how views run in distributed queries, with a link to learning more about poorly performing views. Also a brief section about materialization.

dist_tables/querying.rst
```diff
@@ -119,6 +119,66 @@ In such cases the table(s) to be partitioned are determined by the query optimiz
...
+
+  SELECT pow(2,ceil(log(2,player_avg)))::int AS approx
+  FROM (
+    SELECT player, avg(points) AS player_avg
+    FROM scores
+    GROUP BY player;
+  ) AS avg_scores
+  ORDER BY approx DESC;
+
+The EXPLAIN output above also shows that the query executes pretty efficiently. Because the subquery groups by ``player``, which is also the distribution column of the underlying table, each group can be calculated entirely within a single worker node. The bulk of the processing, including taking approximations, happens in workers under the management of the real-time :ref:`distributed_query_executor`, with the final ordering happening back on the coordinator.
```
begriffs: …I say it executes efficiently, although it does have to sort all the results on one node…

sumedhpathak:
Any reason we dive into how PostgreSQL rewrites Views? I think we want to cover basically:
- Citus supports all Views
- Link to PostgreSQL's page on Views for syntax/features
- Citus also supports materialized views.

We can link to the pull/push section like you did for context on the how, but the above should be sufficient.

This section could be small, but if we have the right SEO and it comes up in the internal search, we should be good. Especially since we now support all Views.

begriffs:
OK, it's just a quick little section now.

PR #547, opened 2017-11-22
Rearrange cloud sections

### LINKED ISSUE #519: Restructure the Cloud section in TOC
### Desired structure

Each item appears in the order it should in the table of contents, and includes a link to the issue describing what content should go inside. Items without an associated issue are already written but perhaps may need to be rearranged in the table of contents.

- Getting started
  - Provisioning
  - Connecting (note here we should explain database connection string)
  - PgBouncer (content in #510)
- Scaling
  - Scaling up
  - Scaling out
  - Move MX?
  - PgBouncer - #491
- Security
  - What we provide for you
  - Connecting with SSL
  - Creating new roles/users
  - IP whitelisting - #507
  - VPC peering - #507 
- High availability and recovery
  - High availability
  - Disaster recovery
  - Point in time recovery
- Additional features
  - Fork
  - Rebalancing - #492
  - Citus MX?
- Monitoring
  - Overview of dashboards within cloud - #506
  - Formation events feed
- Logging
- Support and Billing

begriffs:
Fixes #519


cloud/performance.rst
```diff
@@ -60,10 +57,8 @@ Citus will output the progress as it moves each shard.
 
   The :code:`rebalance_table_shards` function rebalances all tables in the :ref:`colocation group <colocation_groups>` of the table named in its argument. Thus you do not have to call it for every single table, just call it on a representative table from each colocation group.
 
-.. _mx:
-
-Connections
-===========
+PgBouncer
```
sumedhpathak: Should we rename this to Connection Scaling (pgBouncer) or some such? (Instead of just pgBouncer)

index.rst
```diff
@@ -121,7 +121,9 @@ Welcome to the documentation for Citus 7.1! Citus horizontally scales PostgreSQL
    cloud/performance.rst
    cloud/security.rst
    cloud/availability.rst
+   cloud/additional.rst
    cloud/monitoring.rst
+   cloud/support.rst
```
sumedhpathak: We are missing the Logging section here. The content doesn't show up at all.

begriffs:
> Rename "Users and Roles" to "Creating new users/roles"

The section has more than the commands to create users. It also shows how to view and change  permissions for new or existing roles. So "Users and Permissions" felt more accurate to me.

> The "Getting Started" … pgBouncer section … is missing

I moved that under Scaling, since it's about scaling the number of connections. It felt like more detail than someone just getting started would need. Oh and now I see you wrote another comment about it. Yeah, I'll rename the subsection to "Scaling Connections (pgBouncer)" which matches the other subsection names better.

> "What we provide for you" … Is that new content we need to get… ?

Yep, I don't know what we want to add there other than the list of security features implicit in the section itself.

begriffs:
> We currently have one big HA section, which then splits into HA and DR. Craig had HA and DR as separate sections. Is it possible to split those up as well?

I rearranged the content a little to separate it. At some point I'd like to edit it more heavily, but that's out of scope for a this PR.

This is ready for another look.

PR #548, opened 2017-11-22
Note limit pushdown for queries grouped on dist col

### LINKED ISSUE #530: New limit pushdown behavior

It used to be when applying a LIMIT to a query which GROUPs BY the distribution column, the limit would not push down. There was a workaround to make the pushdown happen by setting `limit_clause_row_fetch_count`. However the end-user might think this would produce an approximation, when actually it pushes the complete aggregation to the worker shards.

If I am understanding the fix correctly, Citus will now push down a limit without any need for that GUC.

* Explain that the limit will push down when a GROUP BY is on the distribution column
* Also mention that in that situation the GUC will be disregarded, even though it continues to be useful for other situations.

Update https://docs.citusdata.com/en/v7.0/dist_tables/querying.html#limit-pushdown

* https://github.com/citusdata/citus/issues/1589
  * Reported more generally in https://github.com/citusdata/citus/issues/606
  * Attempted fix: https://github.com/citusdata/citus/pull/1635
  * Actual fix: https://github.com/citusdata/citus/pull/1641

begriffs:
Fixes #530

### DIFF
--- a/dist_tables/querying.rst
+++ b/dist_tables/querying.rst
@@ -76,17 +76,47 @@ You can then call sum(hll_column) to roll up those columns within the database. 
 Limit Pushdown
 #####################
 
-Citus also pushes down the limit clauses to the shards on the workers wherever possible to minimize the amount of data transferred across network.
+Citus pushes limit clauses down to the shards on workers wherever possible to minimize the amount of data transferred across the network. For example, if a select query with limit clause groups by a table's distribution column, then Citus will push the query down and run it directly on each worker. For instance, in a distributed table of deliveries partitioned geographically, we can take a sample of the number of drivers delivering in a few areas:
 
-However, in some cases, SELECT queries with LIMIT clauses may need to fetch all rows from each shard to generate exact results. For example, if the query requires ordering by the aggregate column, it would need results of that column from all shards to determine the final aggregate value. This reduces performance of the LIMIT clause due to high volume of network data transfer. In such cases, and where an approximation would produce meaningful results, Citus provides an option for network efficient approximate LIMIT clauses.
+.. code-block:: sql
 
-LIMIT approximations are disabled by default and can be enabled by setting the configuration parameter citus.limit_clause_row_fetch_count. On the basis of this configuration value, Citus will limit the number of rows returned by each task for aggregation on the coordinator. Due to this limit, the final results may be approximate. Increasing this limit will increase the accuracy of the final results, while still providing an upper bound on the number of rows pulled from the workers.
+  EXPLAIN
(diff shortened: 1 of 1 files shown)


dist_tables/querying.rst
```diff
@@ -86,6 +86,8 @@ LIMIT approximations are disabled by default and can be enabled by setting the c
 
     SET citus.limit_clause_row_fetch_count to 10000;
 
+If, however, a SELECT query with LIMIT clause groups by a table's distribution column then Citus will push the query down and run it directly on each worker. The pushed-down query will include its limit as well, and Citus will disregard the aforementioned GUC.
```
sumedhpathak: Can we move this up front? (Maybe somewhere like line 79). It would be good to start off with LIMIT's where we are able to push down entirely, followed by where we end up pulling data. That probably flows better?

dist_tables/querying.rst
```diff
@@ -76,16 +76,48 @@ You can then call sum(hll_column) to roll up those columns within the database.
...
+  │             Group Key: location_id                                                                                               │
+  │             ->  Index Only Scan using deliveries_pkey_102360 on deliveries_102360 companies  (cost=0.15..58.65 rows=700 width=8) │
+  └──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
+  (12 rows)
+
+Notice how the limit occurs twice in the plan: once at the final step in the coordinator, and also down in the query run on the workers. This limits how much data each worker returns to the coordinator.
+
+However, in some cases SELECT queries with LIMIT clauses may need to fetch all rows from each shard to generate exact results. For example, if we had added an ORDER BY the driver count then the coordinator would need results from all shards to determine the final aggregate value. No individual worker knows which rows are safe to omit, and would send back all rows. This reduces performance of the LIMIT clause due to high volume of network data transfer.
```
begriffs: Actually I didn't quite follow the logic of why limits cannot be pushed down when ordering by the aggregate. For example wouldn't the five  locations with most drivers be chosen from among the best five from each shard? A sixth place location on shard 23 has no chance in the global competition, right?

I must have not parsed the claim properly when rewriting the paragraph.

begriffs:
Tried to make this section more concrete with an example.

dist_tables/querying.rst
sumedhpathak: @begriffs Let's assume the top-5 in each shard are different. However, the 6th place might be the same across all shards (i.e., just miss out on making top-5). However, in aggregate across all shards, it might come out on top. 

Is there anything remaining on this PR btw? I had approved this one.

PR #549, opened 2017-11-24
Update write_throughput_benchmark.rst

steffenix:
Fix example

### DIFF
--- a/articles/semi_structured_data.rst
+++ b/articles/semi_structured_data.rst
@@ -53,7 +53,7 @@ using a
 column for url\_params is much more convenient than creating columns for
 each parameter. With JSONB we can get both the flexibility of schema,
 and combined with `GIN
-indexing <https://www.postgresql.org/docs/9.5/static/gin.html>`__ we can
+indexing <https://www.postgresql.org/docs/10/static/gin.html>`__ we can
 still have performant queries against all keys and values without having
 to index them individually.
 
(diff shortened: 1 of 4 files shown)


begriffs:
@steffenix thanks for the pull request. Is `setrandom` deprecated?

steffenix:
Got replaced in 9.6
https://www.postgresql.org/docs/9.6/static/release-9-6.html

begriffs:
Just reproduced the problem you reported. With postgres-contrib 10 I got the error `insert.sql:4: invalid command in command "setrandom"`.

Your change looks good but there are a few other things to fix:

* Our [install pgbench](https://docs.citusdata.com/en/v7.1/extra/write_throughput_benchmark.html#install-pgbench) instructions specify postgres 9.5. Following those instructions would make the benchmarking guide work without a problem. However I agree with you that we should use a newer version because it's more likely what people have on their machines already. Could you update the install instructions?
* The [update throughput](https://docs.citusdata.com/en/v7.1/extra/write_throughput_benchmark.html#id2) section also has sql with `setrandom`.

Would you like to update your pull request with these additions?

steffenix:
OK I will have alook and update my pull request.

steffenix:
I have changed the version to 10.

steffenix:
I have done and cleanup the mess that was introduced by the merge.

begriffs:
Thank you!

PR #550, opened 2017-11-25
Disable server-side cursors in django apps

gsgalloway:


begriffs:
@gsgalloway thanks for contributing to the docs!

For context, can you tell me how you discovered the need for this change?

gsgalloway:
Certainly. Here's a demo for eliciting the error message I saw seeing:

```sql
CREATE TABLE cursor_test (id integer PRIMARY KEY);

-- Works
SELECT id FROM cursor_test ORDER BY id ASC;
DECLARE "cursor" NO SCROLL CURSOR WITH HOLD FOR SELECT id FROM cursor_test ORDER BY id ASC;

SELECT create_reference_table('cursor_test');

-- Works
SELECT id FROM cursor_test ORDER BY id ASC;

-- Error
DECLARE "cursor2" NO SCROLL CURSOR WITH HOLD FOR SELECT id FROM cursor_test ORDER BY id ASC;

-- ERROR:  rescan is unsupported
-- DETAIL:  We don't expect this code path to be executed.
```

I noticed the error message seemed to suggest an internal error from a Citus node. Or at least the error message was a bit abstruse.

The `DECLARE NO SCROLL CURSOR` query was being generated by the Django rest framework as part of its built-in pagination functionality

PR #551, opened 2017-11-28
Make docs responsive

### LINKED ISSUE #466: Fix on mobile?
Currently the docs theme is not responsive and looks broken on mobile devices. Do we want to fix that, and if so how? (New docs system, or same system but improved theme.) Could examine analytics to see whether many people attempt reading the docs on mobile.

begriffs:
Resetting to vanilla Read The Docs style, and re-applying our customizations more cleanly. This new code organization will allow us to upgrade the `sphinx_rtd_theme` theme independently of our own customizations, meaning our code is not mixed with theirs.

Opening the PR early to ensure that our doc preview pipeline works with it. I'll add a comment when it's ready for review.

Fixes #466


begriffs:
OK, this is good for a review now. It behaves nicely at different screen sizes. If we want the original header back it'll be more complicated. Also that header no longer matched our main site.

begriffs:
TODO: add heap analytics code into this template.

begriffs:
Google analytics is not being added, but on the real RTD it might. I'll merge this to master to check. If Google is not added then I'll add it manually the way I'm doing for Heap.

PR #553, opened 2017-11-30
Furkan's suggestions

### LINKED ISSUE #439: Co-Location page query fix
In the [co-location page](https://docs.citusdata.com/en/latest/sharding/colocation.html) we have the following query
```
SELECT page_id, count(*) AS count
FROM event
WHERE page_id IN (ARRAY[...page IDs from first query...]) AND
      tenant_id = 6 AND
      payload->>'time' >= now() - interval '1 week'
GROUP BY page_id ORDER BY count DESC LIMIT 10;
```
Unfortunately there is no IN operator for arrays in postgresql. 
Additionally, postgres wants to see a type cast on the field ```payload->>'time'```.
The right query would be
```
SELECT page_id, count(*) AS count
FROM event
WHERE page_id = ANY(ARRAY[...page IDs from first query...]) AND
      tenant_id = 6 AND
      (payload->>'time')::date >= now() - interval '1 week'
GROUP BY page_id ORDER BY count DESC LIMIT 10;
```

### LINKED ISSUE #442: Append reference query fix
In the [reference page for append distribution](https://docs.citusdata.com/en/latest/reference/append.html), there is this query introducing ```master_append_table_to_shard```.
```
SELECT * from master_append_table_to_shard(102089, 'github_events_temp', 'master-101', 5432);
```
However, when we follow the instructions, we can see that we did not yet created a table called ```github_events_temp```. So, it might be better to run the query with a table that we are sure that it exists. For example;
```
SELECT * from master_append_table_to_shard(102089, 'github_events', 'master-101', 5432);
```

begriffs:
Fixes #439 
Fixes #442

### DIFF
--- a/sharding/colocation.rst
+++ b/sharding/colocation.rst
@@ -90,7 +90,7 @@ Across all shards of the event table (Q2):
 
   SELECT page_id, count(*) AS count
   FROM event
-  WHERE page_id = ANY(ARRAY[...page IDs from first query...])
+  WHERE page_id IN (…page IDs from first query…)
     AND tenant_id = 6
     AND (payload->>'time')::date >= now() - interval '1 week'
   GROUP BY page_id ORDER BY count DESC LIMIT 10;
--- a/sharding/colocation.rst
+++ b/sharding/colocation.rst
@@ -90,11 +90,10 @@ Across all shards of the event table (Q2):
 
   SELECT page_id, count(*) AS count
   FROM event
-  WHERE page_id IN (ARRAY[...page IDs from first query...]) AND
-        tenant_id = 6 AND
-        payload->>'time' >= now() - interval '1 week'
+  WHERE page_id = ANY(ARRAY[...page IDs from first query...])
+    AND tenant_id = 6
+    AND (payload->>'time')::date >= now() - interval '1 week'
   GROUP BY page_id ORDER BY count DESC LIMIT 10;
-
 
 Afterwards, the results from the two steps need to be combined by the application.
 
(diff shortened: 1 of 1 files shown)


sharding/colocation.rst
```diff
@@ -90,12 +90,11 @@ Across all shards of the event table (Q2):
 
   SELECT page_id, count(*) AS count
   FROM event
-  WHERE page_id IN (ARRAY[...page IDs from first query...]) AND
-        tenant_id = 6 AND
-        payload->>'time' >= now() - interval '1 week'
+  WHERE page_id = ANY(ARRAY[...page IDs from first query...])
```
sumedhpathak: @begriffs Have you tested this query to check that it works as expected (and makes sense in the context of this section)?

begriffs:
I rebased away the change to the append section.

The co-location section actually has other things in its queries that don't make sense, like a reference to `tenant_id` which doesn't exist. I'll look the section over, supply some sample data, and run the queries to get any bugs out.

begriffs:
I rebased away the change to the append section.

Also simplified the code in co-location section. Ran the example, adding some data in and verified that it runs.

sharding/colocation.rst
```diff
@@ -90,12 +91,11 @@ Across all shards of the event table (Q2):
 
   SELECT page_id, count(*) AS count
   FROM event
-  WHERE page_id IN (ARRAY[...page IDs from first query...]) AND
-        tenant_id = 6 AND
-        payload->>'time' >= now() - interval '1 week'
+  WHERE page_id IN (…page IDs from first query…)
```
sumedhpathak: This overall looks good, but I'm wondering if this section needs some rewrite now that we have broader SQL support in 7.2. For example, we could actually write this query as a subquery in 7.2, which sort of diminishes the point we're trying to make.

PR #563, opened 2017-12-04
Update faq entries

### LINKED ISSUE #498: Improve FAQ question on adding nodes to include cloud
https://docs.citusdata.com/en/v7.0/faq/faq.html#how-do-i-add-nodes-to-an-existing-citus-cluster

Currently, it asks people to run master_add_node. Cloud customers never have to run that. So, maybe we can add a few lines which say that for cloud customers this is run automatically when you add a node.

### LINKED ISSUE #499: Update FAQ answer to indicate that we support cross node transactions
https://docs.citusdata.com/en/v7.0/faq/faq.html#are-there-any-postgresql-features-not-supported-by-citus

The answer says that we do not support "Transactional semantics for queries that span across multiple shards". We do support that for DDLs, COPYs and for single row inserts now which span multiple shards. We're also expanding support to cross shard updates and deletes. We should update this answer to indicate that.

### LINKED ISSUE #500: Update answer on handling of failure node according to streaming replication
https://docs.citusdata.com/en/v7.0/faq/faq.html#how-does-citus-handle-failure-of-a-worker-node

Currently, it says "If a worker node fails during e.g. a SELECT query, jobs involving shards from that server will automatically fail over to replica shards located on healthy hosts. "

With streaming replication (which is now our preferred way of replication), there are no replica shards. So, we should update the answer accordingly.

begriffs:
Fixes #498 
Fixes #499 
Fixes #500

### DIFF
--- a/faq/faq.rst
+++ b/faq/faq.rst
@@ -19,13 +19,11 @@ Either way, after adding a node to an existing cluster it will not contain any d
 How does Citus handle failure of a worker node?
 -----------------------------------------------
 
-Using one of two methods depending on configuration:
+Citus supports two modes of replication, allowing it to tolerate worker-node failures. In the first model, we use PostgreSQL's streaming replication to replicate the entire worker-node as-is. In the second model, Citus can replicate data modification statements, thus replicating shards across different worker nodes. They have different advantages depending on the workload and use-case as discussed below:
 
-**Streaming replication** maintains a standby node for every worker, allowing a seamless failover to the standby on failure. Citus Community edition is compatible with streaming replication but does not automatically configure it on a local cluster, whereas Citus Cloud has streaming replication built in.
+1. **PostgreSQL streaming replication.** This option is best for heavy OLTP workloads. It replicates entire worker nodes by continuously streaming their WAL records to a standby. You can configure streaming replication on-premise yourself by consulting the `PostgreSQL replication documentation <https://www.postgresql.org/docs/current/static/warm-standby.html#STREAMING-REPLICATION>`_ or use :ref:`Citus Cloud <cloud_overview>` which is pre-configured for replication and high-availability.
 
-**Shard replication** keeps replicas of each shard on a configurable number of other workers for use in re-routing queries if a host becomes unhealthy. Citus Community edition supports shard replication through easy configuration.
-
-Using either technique means intermittent failures will not require restarting potentially long-running analytical queries. You can find more information about Citus' failure handling logic in :ref:`dealing_with_node_failures`.
+2. **Citus shard replication.** This option is best suited for an append-only workload. Citus replicates shards across different nodes by automatically replicating DML statements and managing consistency. If a node goes down, the co-ordinator node will continue to serve queries by routing the work to the replicas seamlessly. To enable shard replication simply set :code:`SET citus.shard_replication_factor = 2;` (or higher) before distributing data to the cluster.
 
 How does Citus handle failover of the coordinator node?
 -------------------------------------------------------
--- a/faq/faq.rst
+++ b/faq/faq.rst
@@ -12,13 +12,14 @@ Currently Citus imposes primary key constraint only if the distribution column i
 How do I add nodes to an existing Citus cluster?
 ------------------------------------------------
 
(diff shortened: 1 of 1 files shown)


faq/faq.rst
```diff
@@ -45,12 +46,11 @@ Are there any PostgreSQL features not supported by Citus?
...
 
-Other queries which, by contrast, combine data from multiple nodes, do not support the entire spectrum of PostgreSQL features. However they still enjoy broad SQL coverage, including semi-structured data types (like jsonb, hstore), full text search, operators, functions, and foreign data wrappers. Note that the following constructs aren't supported natively for cross-node queries:
+Other queries which, by contrast, combine data from multiple nodes, do not support the entire spectrum of PostgreSQL features. However they still enjoy broad SQL coverage, including semi-structured data types (like jsonb, hstore), full text search, operators, functions, foreign data wrappers, and in many cases even transactional semantics. Note that the following constructs aren't supported natively for cross-node queries:
 
 * Window Functions
 * CTEs
```
sumedhpathak: This will be changing as of 7.2. We should also probably amend this list a bit. We support window functions in certain scenarios now with subquery pushdown.

faq/faq.rst
```diff
@@ -12,13 +12,14 @@ Currently Citus imposes primary key constraint only if the distribution column i
...
 
 How does Citus handle failure of a worker node?
 -----------------------------------------------
 
-If a worker node fails during e.g. a SELECT query, jobs involving shards from that server will automatically fail over to replica shards located on healthy hosts. This means intermittent failures will not require restarting potentially long-running analytical queries, so long as the shards involved can be reached on other healthy hosts.
-You can find more information about Citus' failure handling logic in :ref:`dealing_with_node_failures`.
+If a worker node fails during e.g. a SELECT query, Citus seamlessly fails over to a streaming replication standby node. (If streaming replication is disabled then Citus keeps replica shards for use in re-routing queries if a host becomes unhealthy.) This means intermittent failures will not require restarting potentially long-running analytical queries. You can find more information about Citus' failure handling logic in :ref:`dealing_with_node_failures`.
```
sumedhpathak: >Citus seamlessly fails over to a streaming replication standby node.
Citus doesn't automatically fail over to streaming replica's. This happens in Cloud, but not in community. The Auto-HA project aims to do that, but isn't yet out. The users will need to manually setup failover via repmgr and their own monitoring to detect and perform the failover. 

> If streaming replication is disabled then Citus keeps replica shards for use in re-routing queries
We do that only if shard_replication_factor is set to 2. We don't automatically re-replicate if there is no streaming replication.

faq/faq.rst
begriffs: Maybe we should leave this as-is on the current branch and update the section separately for 7.2, that way all the changes of this PR can be cherry-picked to 7.1

faq/faq.rst
```diff
@@ -12,13 +12,20 @@ Currently Citus imposes primary key constraint only if the distribution column i
...
 
 How does Citus handle failure of a worker node?
 -----------------------------------------------
 
-If a worker node fails during e.g. a SELECT query, jobs involving shards from that server will automatically fail over to replica shards located on healthy hosts. This means intermittent failures will not require restarting potentially long-running analytical queries, so long as the shards involved can be reached on other healthy hosts.
-You can find more information about Citus' failure handling logic in :ref:`dealing_with_node_failures`.
+Using one of two methods depending on configuration:
+
+**Streaming replication** maintains a standby node for every worker, allowing a seamless failover to the standby on failure. Citus Community edition is compatible with streaming replication but does not automatically configure it on a local cluster, whereas Citus Cloud has streaming replication built in.
```
sumedhpathak: I think https://docs.citusdata.com/en/v7.1/admin_guide/cluster_management.html#worker-node-failures captures it well. 
- I think we do want to note that Shard replication is better for append-only workloads.
- Instead of "compatible with", we should note that you can just use the built-in PostgreSQL streaming replication. 
Could you update this with content from that link? It probably captures the same amount of information, but worth noting here as well.

faq/faq.rst
sumedhpathak: I'm OK to update it in #571. We'll probably want to rework this FAQ in some good detail based on that.

begriffs: Yep, the recursive planning docs PR updates this question. I'll rebase away the change here when the other PR is merged.

begriffs:
I just duplicated content from the Cluster Management worker failure section. It was well-written and about the right length for a FAQ answer.

sumedhpathak:
@begriffs  Is this ready for another look?

begriffs:
@sumedhpathak yep, it's ready.

PR #566, opened 2017-12-05
Explain reference tables in more concrete terms

### LINKED ISSUE #505: Motivate when to use reference tables

Reference tables are used to store data that needs to be shared between tenants or is not naturally associated with a single tenant.

We currently discuss *how* to use reference tables but we don't put it in concrete terms. Explain when it usually comes up for multi-tenant apps.

Modify https://docs.citusdata.com/en/v7.0/dist_tables/ddl.html#reference-tables

begriffs:
Do you think this suffices?

Fixes #505

### DIFF
--- a/dist_tables/ddl.rst
+++ b/dist_tables/ddl.rst
@@ -51,13 +51,12 @@ You are now ready to insert data into the distributed table and run queries on i
 Reference Tables
 ~~~~~~~~~~~~~~~~
 
-The above method distributes tables into multiple horizontal shards, but another possibility is distributing tables into a single shard and replicating the shard to every worker node.
+The above method distributes tables into multiple horizontal shards, but another possibility is distributing tables into a single shard and replicating the shard to every worker node. Tables distributed this way are called *reference tables.* They are used to store data that needs to be frequently accessed by multiple nodes in a cluster.
 
-Tables distributed this way are called *reference tables.* They are used to store data that needs to be frequently accessed by multiple nodes in a cluster. In multi-tenant apps this means data shared between tenants or data that is not naturally associated with a single tenant.
-
+Common candidates for reference tables include:
 
 * Smaller tables which need to join with larger distributed tables.
-* Tables which do not have the tenant column or aren't associated with the tenant. (In some cases, users might do it even for tables which are associated with tenant if they don't have tenant id to reduce migration effort.)
+* Tables in multi-tenant apps which lack a tenant id column or which aren't associated with a tenant. (In some cases, to reduce migration effort, users might even choose to make reference tables out of tables associated with a tenant but which currently lack a tenant id.)
 * Tables which need unique constraints across multiple columns and are small enough.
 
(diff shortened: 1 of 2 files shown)


dist_tables/ddl.rst
```diff
@@ -51,7 +51,11 @@ You are now ready to insert data into the distributed table and run queries on i
 Reference Tables
 ~~~~~~~~~~~~~~~~
 
-The above method distributes tables into multiple horizontal shards, but it's also possible to distribute tables into a single shard and replicate it to every worker node. Tables distributed this way are called *reference tables.*  They are typically small non-partitioned tables which we want to locally join with other tables on any worker. One US-centric example is information about states.
+The above method distributes tables into multiple horizontal shards, but another possibility is distributing tables into a single shard and replicating the shard to every worker node.
```
sumedhpathak: Can we start with the motivation here? (e.g. "What about data that is shared amongst tenants?").

I like the opening paragraph here: https://docs.citusdata.com/en/v7.2/use_case_guide/multi_tenant.html#sharing-data-between-tenants.

Also, can we link to the above section from here? (e.g. " See an example app and how we might use reference tables there") (or some such).

begriffs: I was trying to make your suggestion fit with the flow but having trouble. The page as a whole isn't centered around the MT use case, so opening this section talking about sharing data amongst tenants didn't feel right.

However this section does swing around to MT by the second paragraph, and gets into an example in the third.

Good idea about linking to the use case guide. I added the link.

dist_tables/ddl.rst
```diff
@@ -51,7 +51,11 @@ You are now ready to insert data into the distributed table and run queries on i
...
+
+Tables distributed this way are called *reference tables.* They are used to store data that needs to be frequently accessed by multiple nodes in a cluster. In multi-tenant apps this means data shared between tenants or data that is not naturally associated with a single tenant.
```
sumedhpathak: The SE Doc we listed said adding a section like "When to use Reference Tables". I don't know if there is more to add than this paragraph. @samay-sharma Any thoughts?

samay-sharma: A few things on "When to use reference tables"

(a) Smaller tables which need to join with larger distributed tables.
(b) Tables which do not have the tenant column or aren't associated with the tenant. (In some cases, users might do it even for tables which are associated with tenant if they don't have tenant id to reduce migration effort.)
(c) Tables which need unique constraints across multiple columns and are small enough.

sumedhpathak: Thanks @samay-sharma  
@begriffs Could you add these points (bullet listed as Samay wrote up) to the paragraph?

sumedhpathak: @begriffs Is this PR updated with Samay's comments? Thanks

begriffs:
Incorporated Samay's feedback, ready for another look.

begriffs:
This is a fairly minor change, merging it.

PR #573, opened 2017-12-12
Add LogDNA provider

### LINKED ISSUE #569: Add LogDNA configuration to Cloud logging page
Add a section to [External Log Destinations](https://docs.citusdata.com/en/v7.1/cloud/logging.html#external-log-destinations). The settings we want are listed here - https://docs.logdna.com/docs/syslog-ng

begriffs:
@lfittl I tested this on LogDNA and the logs go through.

Couple notes:

* IETF did not work, I need to use BSD so that our syslog conf file had `network()` in its destination section.
* If you were copying-pasting-modifying a message template from another provider in our docs, you might have gotten a directional quote (`”` vs `"`) because Sphinx converted it. That messed me up until I saw the weird unicode escape code in the syslog conf. This pull request wraps all the templates in code boxes to prevent Sphinx from interpreting the quote.

Fixes #569


begriffs:
@lfittl does this look good to merge?

lfittl:
@begriffs If you tested it and it works then +1 on merging it.

PR #574, opened 2017-12-13
setrandom is deprecate in postgres 9.6

chaintng:
In postgres 9.6, they deprecate \setrandom command
They suggest to use built-in random(x,y) function

reference: https://www.postgresql.org/docs/9.6/static/release-9-6.html


begriffs:
Thanks for the contribution!

However someone else recently noticed the same problem and send a fix in #549, which has been merged.

PR #584, opened 2018-01-03
Misc updates for recursive query planning

### LINKED ISSUE #571: Document new recursive query planner

Citus is adding a new query planning mode for Improved SQL support of subqueries and CTEs.

Recursive execution allows new queries, but sometimes isn't a good choice for performance. When the subquery returns too much information, the network overhead to send it to the coordinator can be too high. We want to help people determine if this is the case, and if so modify the query to be router executable if possible. Include examples of a slow recursive executor query and its modification.

### Docs changes needed

* Adjust workarounds section
  * Remove https://docs.citusdata.com/en/v7.1/reference/sql_workarounds.html#subqueries-in-where
  * Just wrap local table in subquery/CTE https://docs.citusdata.com/en/v7.1/reference/sql_workarounds.html#join-a-local-and-a-distributed-table
  * https://docs.citusdata.com/en/v7.1/reference/sql_workarounds.html#window-functions
    * No change for this one. Still requires the workaround.
  * Test and remove section https://docs.citusdata.com/en/v7.1/reference/sql_workarounds.html#data-warehousing-queries
    * The Citus on my system which identifies itself as `7.2-3` is still unable to partition a window function on a non-distribution column. Tried moving that part into a subquery but got the same error, so leaving this section unchanged.
  * Add sections for features where router query workaround exists with perf implications. **Question**: I can see how filtering a query for router execution will make it work, but why are there "perf implications?" Seems like it would run quickly, albeit limiting information to a single tenant.
    * Aggregate distinct other than count(distinct), for instance sum(distinct) and avg(distinct)
    * array_agg with ORDER BY
    * window functions on non-partition keys
* Expand claims in SQL coverage section https://docs.citusdata.com/en/v7.1/reference/citus_sql_reference.html
  * Reverse the section: it's much shorter now to list the things we *don't* support than those we *do*.
    * Correlated subqueries (that are not pushdownable)
    * Recursive/modifying CTEs
    * Table sample
    * SELECT … FOR UPDATE
    * Grouping sets
    * There are a few exceptional combinations of CTEs and subqueries that do not yet work. However they eventually will. Do we document them now or just wait?
* Query Processing https://docs.citusdata.com/en/v7.1/performance/query_processing.html
  * Add the router executor
  * Show how executors can be mixed with recursive planning and their intermediate results brought together as short-lived reference tables
  * Give an outline of how Citus decides which execution strategy to use
    * Like, "first it tries to push down, and if it can't then … Finally the logic applies recursively to CTEs and subqueries"
* https://docs.citusdata.com/en/v7.1/performance/performance_tuning.html
  * Talk about discovering and fixing bottlenecks that are due to inadvertent use of recursive planning
  * Where the recursive planner *is* required, talk about limiting/filtering in ctes/subqueries when possible
    * `WITH cte_slow AS (SELECT * FROM users_table) SELECT * FROM cte_slow LIMIT 10;`
    * vs `WITH cte_fast AS (SELECT * FROM users_table LIMIT 10) SELECT * FROM cte_fast;`
  * The `citus.max_intermediate_result_size` GUC can help cancel accidental `SELECT * FROM big_table` subqueries
  * Beware of queries whose top-level is router executable but has large intermediate results -- this will force computation on a single core. Give an example.
* FAQ
  * Handling errors due to `max_intermediate_result_size` violation
  * Things are looking better in "Are there any PostgreSQL features not supported by Citus?"
* GUC reference
  * Add max_intermediate_result_size
* Other pull requests which are slightly invalidated (update them before merging)
  * #553 - the "naive" query will actually work
  * #544 - the suggestion to make per-tenant views is no longer necessary
  * #563 - touches an area of the FAQ that will have to be changed

https://github.com/citusdata/citus/issues/1804
(The main issue about it, with PRs linking to it)

begriffs:
This will touch a lot, and fixes #571 

Opening it early

### DIFF
--- a/faq/faq.rst
+++ b/faq/faq.rst
@@ -43,14 +43,15 @@ If you want to do joins between small dimension tables (regular Postgres tables)
 Are there any PostgreSQL features not supported by Citus?
 ---------------------------------------------------------
 
-Since Citus provides distributed functionality by extending PostgreSQL, it uses the standard PostgreSQL SQL constructs. It provides full SQL support for queries which access a single node in the database cluster. These queries are common, for instance, in multi-tenant applications where different nodes store different tenants (see :ref:`when_to_use_citus`).
+Since Citus provides distributed functionality by extending PostgreSQL, it uses the standard PostgreSQL SQL constructs. The vast majority of queries are supported, even when they combine data across the network from multiple database nodes. Currently all SQL is supported except:
 
-Other queries which, by contrast, combine data from multiple nodes, do not support the entire spectrum of PostgreSQL features. However they still enjoy broad SQL coverage, including semi-structured data types (like jsonb, hstore), full text search, operators, functions, and foreign data wrappers. Note that the following constructs aren't supported natively for cross-node queries:
+* Correlated subqueries
+* Recursive/modifying CTEs
+* Table sample
+* SELECT … FOR UPDATE
+* Grouping sets
 
-* Window Functions
-* CTEs
-* Set operations
-* Transactional semantics for queries that span across multiple shards
+What's more, Citus has 100% SQL support for queries which access a single node in the database cluster. These queries are common, for instance, in multi-tenant applications where different nodes store different tenants (see :ref:`when_to_use_citus`).
 
 .. _faq_choose_shard_count:
 
--- a/faq/faq.rst
+++ b/faq/faq.rst
@@ -40,6 +40,8 @@ Can I join distributed and non-distributed tables together in the same query?
 
 If you want to do joins between small dimension tables (regular Postgres tables) and large tables (distributed), then wrap the local table in a subquery. Citus' subquery execution logic will allow the join to work. See :ref:`join_local_dist`.
 
+.. _unsupported:
+
 Are there any PostgreSQL features not supported by Citus?
 ---------------------------------------------------------
 
@@ -51,7 +53,11 @@ Since Citus provides distributed functionality by extending PostgreSQL, it uses 
 * SELECT … FOR UPDATE
 * Grouping sets
 
+
 What's more, Citus has 100% SQL support for queries which access a single node in the database cluster. These queries are common, for instance, in multi-tenant applications where different nodes store different tenants (see :ref:`when_to_use_citus`).
+
+Remember that -- even with this extensive SQL coverage -- data modeling can have a significant impact on query performance. See the section on :ref:`citus_query_processing` for details on how Citus executes queries.
+
 
 .. _faq_choose_shard_count:
 
@@ -109,7 +115,9 @@ The data in distributed tables lives on the worker nodes (in shards), not on the
 Why am I seeing an error about max_intermediate_result_size?
 ------------------------------------------------------------
 
-Citus has to use more than one step to run some queries having subqueries or CTEs. It copies subquery results into temporary :ref:`reference_tables` for use by the main query. If these results are too large, this might cause unacceptable network overhead. Citus has a configurable setting, ``citus.max_intermediate_result_size`` to specify a subquery result size threshold at which the query will be canceled. If you run into the error, it looks like:
+Citus has to use more than one step to run some queries having subqueries or CTEs. Using :ref:`push_pull_execution`, it pushes subquery results to all worker nodes for use by the main query. If these results are too large, this might cause unacceptable network overhead, or even insufficient storage space on the coordinator node which accumulates and distributes the results.
+
+Citus has a configurable setting, ``citus.max_intermediate_result_size`` to specify a subquery result size threshold at which the query will be canceled. If you run into the error, it looks like:
 
 ::
 
@@ -117,7 +125,7 @@ Citus has to use more than one step to run some queries having subqueries or CTE
   DETAIL:  Citus restricts the size of intermediate results of complex subqueries and CTEs to avoid accidentally pulling large result sets into once place.
   HINT:  To run the current query, set citus.max_intermediate_result_size to a higher value or -1 to disable.
 
-As the error message suggests, you can increase this limit by altering the variable:
+As the error message suggests, you can (cautiously) increase this limit by altering the variable:
 
 .. code-block:: sql
 
--- a/faq/faq.rst
+++ b/faq/faq.rst
@@ -38,7 +38,7 @@ ingest workflows.
 Can I join distributed and non-distributed tables together in the same query?
 -----------------------------------------------------------------------------
 
-If you want to do joins between small dimension tables (regular Postgres tables) and large tables (distributed), then you can distribute the small tables as "reference tables." This creates a single shard replicated across all worker nodes. Citus will then be able to push the join down to the worker nodes. If the local tables you are referring to are large, we generally recommend to distribute the larger tables to reap the benefits of sharding and parallelization which Citus offers. For a deeper discussion, see :ref:`reference_tables` and our :ref:`joins` documentation.
+If you want to do joins between small dimension tables (regular Postgres tables) and large tables (distributed), then wrap the local table in a subquery. Citus' subquery execution logic will allow the join to work. See :ref:`join_local_dist`.
 
 Are there any PostgreSQL features not supported by Citus?
 ---------------------------------------------------------
--- a/faq/faq.rst
+++ b/faq/faq.rst
@@ -105,6 +105,23 @@ Why does pg_relation_size report zero bytes for a distributed table?
 
 The data in distributed tables lives on the worker nodes (in shards), not on the coordinator. A true measure of distributed table size is obtained as a sum of shard sizes. Citus provides helper functions to query this information. See :ref:`table_size` to learn more.
 
+Why am I seeing an error about max_intermediate_result_size?
+------------------------------------------------------------
+
+Citus has to use more than one step to run some queries having subqueries or CTEs. It copies subquery results into temporary :ref:`reference_tables` for use by the main query. If these results are too large, this might cause unacceptable network overhead. Citus has a configurable setting, ``citus.max_intermediate_result_size`` to specify a subquery result size threshold at which the query will be canceled. If you run into the error, it looks like:
+
+::
+
+  ERROR:  the intermediate result size exceeds citus.max_intermediate_result_size (currently 1 GB)
+  DETAIL:  Citus restricts the size of intermediate results of complex subqueries and CTEs to avoid accidentally pulling large result sets into once place.
+  HINT:  To run the current query, set citus.max_intermediate_result_size to a higher value or -1 to disable.
+
+As the error message suggests, you can increase this limit by altering the variable:
+
+.. code-block:: sql
+
+  SET citus.max_intermediate_result_size = '3GB';
+
 Can I run Citus on Heroku or Amazon RDS?
 ----------------------------------------
 
--- a/performance/performance_tuning.rst
+++ b/performance/performance_tuning.rst
@@ -192,9 +192,9 @@ Subquery/CTE Network Overhead
 
 In the best case Citus can execute queries containing subqueries and CTEs in a single step. This is usually because both the main query and subquery filter by tables' distribution column in the same way, and can be pushed down to worker nodes together. However Citus is sometimes forced to execute subqueries *before* executing the main query, copying the intermediate subquery results to other worker nodes for use by the main query. This technique is called :ref:`push_pull_execution`.
 
-It's important to be aware when subqueries are executed in a separate step, and avoid sending too much data between worker nodes. The network overhead will hurt performance.
+It's important to be aware when subqueries are executed in a separate step, and avoid sending too much data between worker nodes. The network overhead will hurt performance. The EXPLAIN command allows you to discover how queries will be executed, including whether multiple steps are required. For a detailed example see :ref:`push_pull_execution`.
 
-To detect if this is happening, adjust the ``max_intermediate_result_size`` limit in a new connection to the coordinator node. By default the max intermediate result size is 1GB, which is large enough to allow some inefficient queries. Try turning it down and running your queries:
+Also you can defensivly set a safeguard against large intermediate results. Adjust the ``max_intermediate_result_size`` limit in a new connection to the coordinator node. By default the max intermediate result size is 1GB, which is large enough to allow some inefficient queries. Try turning it down and running your queries:
 
 .. code-block:: sql
 
--- a/performance/performance_tuning.rst
+++ b/performance/performance_tuning.rst
@@ -190,7 +190,7 @@ Other than the above, there are two configuration parameters which can be useful
 Subquery/CTE Network Overhead
 =============================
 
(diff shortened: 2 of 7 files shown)


begriffs:
These changes cover the items I outlined in the issue. There may be more left to say, but I think the PR is at a good place to review and discuss.

sumedhpathak:
@begriffs Could you re-push to Heroku? Thanks!

performance/query_processing.rst
```diff
@@ -34,7 +34,9 @@ Distributed Query Executor
 
 Citus’s distributed executors run distributed query plans and handle failures that occur during query execution. The executors connect to the workers, send the assigned tasks to them and oversee their execution. If the executor cannot assign a task to the designated worker or if a task execution fails, then the executor dynamically re-assigns the task to replicas on other workers. The executor processes only the failed query sub-tree, and not the entire query while handling failures.
 
-Citus has two executor types - real time and task tracker. The former is useful for handling simple key-value lookups and INSERT, UPDATE, and DELETE queries, while the task tracker is better suited for larger SELECT queries.
+Citus has three basic executor types: real time, router, and task tracker. It can use more than one at once on a single query, assigning different executors to different subqueries/CTEs as needed to support the SQL functionality. This process is recursive: if Citus cannot determine how to run a subquery then it examines sub-subqueries.
```
sumedhpathak: Do we want to note somewhere that all of this is handled by Citus automatically? (i.e., the executor choice is "Dynamic"), and the user needs not take any action.
We can also show an `EXPLAIN` example to allow users to determine which executor a query is using. E.g., 
"if you want to know which executor a query is using to debug performance issues, you can check our explain plan".

performance/query_processing.rst
```diff
@@ -50,7 +50,7 @@ Furthermore, when the real time executor detects simple INSERT, UPDATE or DELETE
 Router Executor
 ---------------
 
-When all data required for a query is stored on a single node, Citus can route the entire query to the node and run it there. The result-set is then relayed through the coordinator node back to the client. The router executor takes care of this type of execution, so named because a query is *routed* to a single node.
```
sumedhpathak: Why the edit here?

performance/query_processing.rst
```diff
@@ -61,6 +61,28 @@ The task tracker executor is well suited for long running, complex data warehous
 
 Each task tracker daemon on the workers also makes sure to execute at most citus.max_running_tasks_per_node concurrently. This concurrency limit helps in avoiding disk I/O contention when queries are not served from memory. The task tracker executor is designed to efficiently handle complex queries which require repartitioning and shuffling intermediate data among workers.
 
+Subquery/CTE Push-Pull Execution
+--------------------------------
+
+If necessary Citus can push results from subqueries and CTEs into short-lived :ref:`reference_tables` for use by an outer query. This allows Citus to support a greater variety of SQL constructs, and even mix executor types between a query and its subqueries.
```
sumedhpathak: Maybe note that "For CTE's and Subqueries which cannot be flattened or otherwise pushed down, Citus will recursively plan and execute these independently. The results of these are gathered on the coordinator and then pushed back to the workers in order to join or combine the results".

As noted in a separate comment, these aren't reference tables. So we should remove the link here, and also the text. We push them as files and read from them directly. Conceptually they are like reference tables, but if we call them that, users might get confused about what they could do with it.

performance/query_processing.rst
```diff
@@ -61,6 +61,28 @@ The task tracker executor is well suited for long running, complex data warehous
...
+  WHERE page_id IN (
+    SELECT page_id
+    FROM visits
+    GROUP BY page_id
+    ORDER BY count(*) DESC
+    LIMIT 20
+  )
+  GROUP BY page_id;
+
+The real-time executor would like to run a fragment of this query against each shard by page_id, counting distinct session_ids, and combining the results on the coordinator. However the LIMIT in the subquery means the subquery cannot be executed as part of the fragment. However by recursively planning the query Citus can run the subquery separately, push the results to all workers as a temporary reference table, run the main fragment query, and pull the results back to the coordinator. The "push-pull" design supports a subqueries like the one above.
```
sumedhpathak: Again note that the results aren't pushed back as a reference table. 
Also, would an Explain plan help illustrate this example?

faq/faq.rst
```diff
@@ -105,6 +105,23 @@ Why does pg_relation_size report zero bytes for a distributed table?
 
 The data in distributed tables lives on the worker nodes (in shards), not on the coordinator. A true measure of distributed table size is obtained as a sum of shard sizes. Citus provides helper functions to query this information. See :ref:`table_size` to learn more.
 
+Why am I seeing an error about max_intermediate_result_size?
+------------------------------------------------------------
+
+Citus has to use more than one step to run some queries having subqueries or CTEs. It copies subquery results into temporary :ref:`reference_tables` for use by the main query. If these results are too large, this might cause unacceptable network overhead. Citus has a configurable setting, ``citus.max_intermediate_result_size`` to specify a subquery result size threshold at which the query will be canceled. If you run into the error, it looks like:
```
sumedhpathak: Again the note that the data isn't stored in reference tables. Remove the link/text here.
Should we link to the section which talks about the push-pull architecture? This can help expand on their understanding.

sumedhpathak: > If these results are too large, this might cause unacceptable network overhead.

We should note that the results are actually pulled to the coordinator. So apart from network overhead, an unbounded data size could cause the disk to fill up on the master, causing crashes and availability outages.

faq/faq.rst
```diff
@@ -105,6 +105,23 @@ Why does pg_relation_size report zero bytes for a distributed table?
...
+
+::
+
+  ERROR:  the intermediate result size exceeds citus.max_intermediate_result_size (currently 1 GB)
+  DETAIL:  Citus restricts the size of intermediate results of complex subqueries and CTEs to avoid accidentally pulling large result sets into once place.
+  HINT:  To run the current query, set citus.max_intermediate_result_size to a higher value or -1 to disable.
+
+As the error message suggests, you can increase this limit by altering the variable:
```
sumedhpathak: We should probably again Warn or Note the implications of just increasing this value, and do it with care.

faq/faq.rst
```diff
@@ -43,14 +43,15 @@ If you want to do joins between small dimension tables (regular Postgres tables)
...
+* Recursive/modifying CTEs
+* Table sample
+* SELECT … FOR UPDATE
+* Grouping sets
 
-* Window Functions
-* CTEs
-* Set operations
-* Transactional semantics for queries that span across multiple shards
+What's more, Citus has 100% SQL support for queries which access a single node in the database cluster. These queries are common, for instance, in multi-tenant applications where different nodes store different tenants (see :ref:`when_to_use_citus`).
```
sumedhpathak: Should we add another paragraph which says something like:
"Citus does support almost all SQL, but data modeling can have a significant impact on query performance. See the section on `query processing` to see how Citus executes queries and performance implications" (or some such).

performance/performance_tuning.rst
```diff
@@ -185,6 +185,47 @@ Citus has two executor types for running SELECT queries. The desired executor ca
 
 Other than the above, there are two configuration parameters which can be useful in cases where approximations produce meaningful results. These two parameters are citus.limit_clause_row_fetch_count and citus.count_distinct_error_rate. The former sets the number of rows to fetch from each task while calculating limits while the latter sets the desired error rate when calculating approximate distinct counts. You can learn more about the applicability and usage of these parameters in the user guide sections: :ref:`count_distinct` and :ref:`limit_pushdown`.
 
+.. _subquery_perf:
+
+Subquery/CTE Network Overhead
+=============================
+
+In the best case Citus can execute queries containing subqueries and CTEs in a single step. This is usually because both the main query and subquery filter by tables' distribution column in the same way, and can be pushed down to worker nodes together. However Citus is sometimes forced to execute subqueries *before* executing the main query, and then copy the intermediate subquery results to other worker nodes for use by the main query.
```
sumedhpathak: Could we link to the push/pull executor section in Query processing?

performance/performance_tuning.rst
```diff
@@ -185,6 +185,47 @@ Citus has two executor types for running SELECT queries. The desired executor ca
...
+
+It's important to be aware when subqueries are executed in a separate step, and avoid sending too much data between worker nodes. The network overhead will hurt performance.
+
+To detect if this is happening, adjust the ``max_intermediate_result_size`` limit in a new connection to the coordinator node. By default the max intermediate result size is 1GB, which is large enough to allow some inefficient queries. Try turning it down and running your queries:
```
sumedhpathak: Can they also not look at the Explain plan, rather than playing around with this parameter?

reference/sql_workarounds.rst
```diff
@@ -13,116 +13,37 @@ subset of SQL statements. We are continuously working to increase SQL coverage
...
-  SELECT create_reference_table('here');
+  -- or
 
-This will create a table with a single shard (non-distributed), but will
-replicate that shard to every node in the cluster. Now Citus will accept a join query between *here* and *there*, and each worker will have all the information it needs to work efficiently.
+  WITH x AS (SELECT * FROM local)
+  SELECT * FROM x
+  JOIN dist USING (id);
 
 .. _window_func_workaround:
```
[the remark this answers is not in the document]
begriffs: Yes, the window function workaround is still required.

I can also update the temp table to a CTE instead.

performance/performance_tuning.rst
```diff
@@ -185,6 +185,47 @@ Citus has two executor types for running SELECT queries. The desired executor ca
...
+.. _subquery_perf:
+
+Subquery/CTE Network Overhead
+=============================
+
+In the best case Citus can execute queries containing subqueries and CTEs in a single step. This is usually because both the main query and subquery filter by tables' distribution column in the same way, and can be pushed down to worker nodes together. However Citus is sometimes forced to execute subqueries *before* executing the main query, copying the intermediate subquery results to other worker nodes for use by the main query. This technique is called :ref:`push_pull_execution`.
+
+It's important to be aware when subqueries are executed in a separate step, and avoid sending too much data between worker nodes. The network overhead will hurt performance.
+
+To detect if this is happening, adjust the ``max_intermediate_result_size`` limit in a new connection to the coordinator node. By default the max intermediate result size is 1GB, which is large enough to allow some inefficient queries. Try turning it down and running your queries:
```
sumedhpathak: I had a previous comment that I'll reiterate here. Instead of detecting this via modification of this parameter, can we use EXPLAIN to show how the plan would be executed? I'd want to highlight and promote EXPLAIN as a way to diagnose/debug Citus execution and performance issues.

begriffs: Oops, I added an example of EXPLAIN in query_processing.rst and overlooked the one here.

performance/query_processing.rst
```diff
@@ -53,6 +63,69 @@ The task tracker executor is well suited for long running, complex data warehous
...
+             Group Key: visits.page_id, visits.session_id
+             ->  Hash Join  (cost=17.00..651.99 rows=16508 width=8)
+               Hash Cond: (visits.page_id = intermediate_result.page_id)
+               ->  Seq Scan on visits_102200 visits  (cost=0.00..509.17 rows=33017 width=8)
+               ->  Hash  (cost=14.50..14.50 rows=200 width=4)
+                 ->  HashAggregate  (cost=12.50..14.50 rows=200 width=4)
+                   Group Key: intermediate_result.page_id
+                   ->  Function Scan on read_intermediate_result intermediate_result  (cost=0.00..10.00 rows=1000 width=4)
+
+We can see that Citus uses the real-time executor to first execute the subquery as "Distributed Subplan 7_1," then runs another query to do the ``count(distinct session_id)`` with a HashAggregate.
```
sumedhpathak: Is that what people should look for then? The presence of something like "Distributed Sub Plan"? Or can we show snippets of the plan, and show how data is flowing a bit clearer?

begriffs:
@sumedhpathak I dissected the explain output, do you I went too detailed with it? If you like it I'll merge.

performance/query_processing.rst
```diff
@@ -53,6 +63,179 @@ The task tracker executor is well suited for long running, complex data warehous
...
+    FROM visits
+    GROUP BY page_id
+    ORDER BY count(*) DESC
+    LIMIT 20
+  )
+  GROUP BY page_id;
+
+The real-time executor would like to run a fragment of this query against each shard by page_id, counting distinct session_ids, and combining the results on the coordinator. However the LIMIT in the subquery means the subquery cannot be executed as part of the fragment. By recursively planning the query Citus can run the subquery separately, push the results to all workers, run the main fragment query, and pull the results back to the coordinator. The "push-pull" design supports a subqueries like the one above.
+
+Let's see this in action by reviewing the EXPLAIN output. First we'll try the query and remove the LIMIT clause. Then we'll add the LIMIT clause back and see what happens to the execution plan.
```
sumedhpathak: So if I understand this right, the first example should get pushed down entirely to the workers? The second example with the LIMIT added back will use the pull-push planner? If so, I'd skip the first example, and go directly to the one with LIMIT. I do see value in comparing them, but we haven't made it explicit or clear that this query gets pushed down.

performance/query_processing.rst
```diff
@@ -53,6 +63,179 @@ The task tracker executor is well suited for long running, complex data warehous
...
+    FROM visits
+    GROUP BY page_id
+    ORDER BY count(*) DESC
+    LIMIT 20
+  )
+  GROUP BY page_id;
+
+::
+
+  GroupAggregate  (cost=0.00..0.00 rows=0 width=0)
```
sumedhpathak: Can we show the full plan first? And then start breaking it down?

performance/query_processing.rst
```diff
@@ -53,6 +63,179 @@ The task tracker executor is well suited for long running, complex data warehous
...
+                    Node: host=localhost port=5433 dbname=postgres
+                    ->  Limit  (cost=1883.00..1883.05 rows=20 width=12)
+                      ->  Sort  (cost=1883.00..1965.54 rows=33017 width=12)
+                        Sort Key: (count(*)) DESC
+                        ->  HashAggregate  (cost=674.25..1004.42 rows=33017 width=12)
+                          Group Key: page_id
+                          ->  Seq Scan on visits_102264 visits  (cost=0.00..509.17 rows=33017 width=4)
+  .
+
+This big chunk governs running the subquery and making its results available.
```
sumedhpathak: Maybe note that we use the real-time executor for this.

performance/query_processing.rst
```diff
@@ -53,6 +63,179 @@ The task tracker executor is well suited for long running, complex data warehous
...
+::
+
+              ->  Seq Scan on visits_102264 visits  (cost=0.00..509.17 rows=33017 width=8)
+              ->  Hash  (cost=14.50..14.50 rows=200 width=4)
+                ->  HashAggregate  (cost=12.50..14.50 rows=200 width=4)
+                  Group Key: intermediate_result.page_id
+                  ->  Function Scan on read_intermediate_result intermediate_result  (cost=0.00..10.00 rows=1000 width=4)
+  .
+
+The intermediate results are internally retrieved from a function which loads data from a file that was passed between the worker nodes.
```
sumedhpathak: Actually, the data from the first query is retrieved back to the master and then pushed back down to all workers. They don't transfer between the workers. Could you note that?

PR #600, opened 2018-01-29
Update benchmark instructions

### LINKED ISSUE #589: Benchmark instructions on our website don't work
I tried running the benchmark instructions on our website. For this, I logged into the AWS console, and picked the RedHat system (RHEL-7.4_HVM) recommended by AWS. I then picked m4.16xlarge as the instance type.

When I run the instructions on our [docs](https://docs.citusdata.com/en/v7.1/extra/write_throughput_benchmark.html#citus-write-throughput-benchmark), they fail.

<code>
[ec2-user@ip-10-0-0-28 ~]$ sudo yum install postgresql10-contrib
Loaded plugins: amazon-id, rhui-lb, search-disabled-repos
rhui-REGION-client-config-server-7                                                                                                    | 2.9 kB  00:00:00     
rhui-REGION-rhel-server-releases                                                                                                      | 3.5 kB  00:00:00     
rhui-REGION-rhel-server-rh-common                                                                                                     | 3.8 kB  00:00:00     
No package postgresql10-contrib available.
Error: Nothing to do
</code>

begriffs:
Pg 10 hasn't made it into many debian or rhel based distros. This updates the steps to use PGDG for the install.

Fixes #589

### DIFF
--- a/extra/write_throughput_benchmark.rst
+++ b/extra/write_throughput_benchmark.rst
@@ -37,14 +37,13 @@ Once we create a new EC2 instance, we need to install pgbench on this instance. 
 
 If you are running a **Debian** based system, simply type::
 
+  sudo sh -c 'echo "deb http://apt.postgresql.org/pub/repos/apt/ $(lsb_release -cs)-pgdg main" > /etc/apt/sources.list.d/pgdg.list'
+  sudo apt-get install wget ca-certificates
+  wget --quiet -O - https://www.postgresql.org/media/keys/ACCC4CF8.asc | sudo apt-key add -
   sudo apt-get update
   sudo apt-get install postgresql-10
 
-If you are running a **RedHat** based system, simply type::
-
-  sudo yum update
-  sudo yum install postgresql10-contrib
-
+If you are running a **RedHat** based system, follow the `yum installation guide <https://wiki.postgresql.org/wiki/YUM_Installation>`_.
 
 Benchmark INSERT Throughput
 ---------------------------
--- a/extra/write_throughput_benchmark.rst
+++ b/extra/write_throughput_benchmark.rst
@@ -43,7 +43,7 @@ If you are running a **Debian** based system, simply type::
   sudo apt-get update
   sudo apt-get install postgresql-10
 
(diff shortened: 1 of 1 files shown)


extra/write_throughput_benchmark.rst
```diff
@@ -37,14 +37,13 @@ Once we create a new EC2 instance, we need to install pgbench on this instance.
...
+  wget --quiet -O - https://www.postgresql.org/media/keys/ACCC4CF8.asc | sudo apt-key add -
   sudo apt-get update
   sudo apt-get install postgresql-10
 
-If you are running a **RedHat** based system, simply type::
-
-  sudo yum update
-  sudo yum install postgresql10-contrib
-
+If you are running a **RedHat** based system, follow the `yum installation guide <https://wiki.postgresql.org/wiki/YUM_Installation>`_.
```
begriffs: For Debian systems I was able to write a single script that worked. However the systems using the `yum` package manager vary too much for a simple script to work. Thus I point people at the postgres wiki about it.

Another option would be to drop the instructions for RHEL and just tell people to use Debian.

sumedhpathak: Also, pgbench is available in the contrib package. We removed those specific instructions, so if someone just installs PostgreSQL, they wouldn't have pgbench?

begriffs:
> if someone just installs PostgreSQL, they wouldn't have pgbench?

Actually in my tests with the instructions above, a postgres install includes pgbench as well.

* Ubuntu: `/usr/bin/pgbench -> ../share/postgresql-common/pg_wrapper`
* RHEL: `/usr/pgsql-10/bin/pgbench`

It would be a slimmer install if we instructed people to install only contrib items, but the current instructions work OK. Can investigate more if you like.

begriffs:
Actually I found that the `-client` packag installs pgbench, and is obviously slimmer.

begriffs:
Oh no! Nevermind, postgresql-10-client installs a broken pgbench:

```
Error: pg_wrapper: pgbench was not found in /usr/lib/postgresql/10/bin
```

Also the contrib package just looks like an alias nowadays for the full package:

```
Note, selecting 'postgresql-10' instead of 'postgresql-contrib-10'
```

So I'm putting the instructions back to how they were, excepting using the redhat installer link you suggested rather than the wiki.

PR #605, opened 2018-02-06
Bump Citus to 7.2.1

velioglu:


begriffs:
* [x] Verified that I can deploy this CF template
* [x] Verified that https://s3.amazonaws.com/citus-deployment/aws/citus7/cloudformation/citus-7.2.1.json is accessible from an anonymous browser

@velioglu one question: `\dx` reported citus 7.2-3, is that expected? (It's the same version as 7.2.0 reports on my local cluster.)

velioglu:
Since we didn't change anything on sql files, it is expected. You can check the version with `show citus.version`

begriffs:
Cool, it reports as 7.2.1.

This looks good to merge.

(Note that the description in the CloudFormation dashboard still says "Citus 7.2," but maybe we're intentionally omitting the patch version.)

PR #606, opened 2018-02-06
Simplify main menu

begriffs:
I based this on notes from our meeting where we decided to use simple action-oriented topics.

See how you like this in general, I think it looks pretty clean. After we settle on the exact table of contents, then I'll adjust some of the file names which are slightly haphazard. I didn't sweat the file organization too much since this is a proof of concept. One consequence of adjusting the menu is that some pages had to be merged into one in order to put their contents under the same menu sub-headings.

Once we agree on the menu structure I will:

* [x] Rebase on the latest v7.3 changes
* [x] Apply changes from my and Sumedh's discussion
* [x] Rename/reorganize files slightly
* [x] Create a mapping from old to new paths for RTD's page redirection rules. If I don't set up redirects then our search engine rankings will get decimated because page urls will break.
* [x] Fix sphinx warnings
* [x] Scan moved sections for missing syntax highlighting (because some of them relied on the default language lexer being specified at the top of their pages, but now separate pages are merged together.)
* [ ] After merging, test page redirection rules on "latest" and when that works smoothly, cherry pick the menu redesign to v7.3 and apply all those same redirection rules there.


sumedhpathak:
@begriffs Sorry about the delay on this one. Do the conflicts need to be resolved before I can look at it?

begriffs:
You can look at it as-is. Resolving the conflicts won't change the menu structure too much.

begriffs:
@sumedhpathak this is ready for another look. If you like the structure I can work on the next steps.

sumedhpathak:
Ok, I prefer the previous version to this. :) 

I don't quite like the Apps/Queries section, as it seems a bit obscure. I think we were closer in the previous iteration. Can we revert to that version?

I had a few main thoughts on the previous version:
- Installation being buried within Manage/Administration
- Choosing the partition-column section falling in a much lower section.
- Menu seemingly derivative. 

If we revert to the previous setting, I think we could probably sit together and figure out some of these issues.

begriffs:
@sumedhpathak take a last look at this, and if you like the names and order of things I'll fix up the filenames for each page which are currently a bit scrambled by the reshuffling.

Depending on your browser zoom level, the "Cloud" section is just barely above the fold with the new top level sections.

Also someday I would be overjoyed to remove the "Articles" section which I feel is disorganized and doesn't match the rest of the flow. The articles' content could be grafted into the appropriate existing sections.

begriffs:
@sumedhpathak as an experiment I broke all the content under Develop into smaller pages. The way to do it was to make some pages that tie others together. (For instance "Reference.") We can adjust the depth of the menu displayed in such pages, and I tried a few different depths in some of them so you can see what it looks like.

I could also opt to make the menu disappear from those intermediate pages and we can add an intro paragraph or something, but I don't know how to remove the intermediate pages entirely. Not seeing other techniques in the sphinx docs, issues, or stack overflow answers. Also asked in their IRC room but got no response. This is a sphinx issue rather than an RTD issue.

begriffs:
I've renamed a bunch of paths in this PR. Wanted to note that Moz [suggests](https://moz.com/learn/seo/url) that stuffing keywords into URLs is not a big factor for SEO:

> URLs themselves generally do not have a major impact on a page’s ability to rank. So, while it’s worth thinking about, don’t create otherwise unuseful URLs simply to include a keyword in them.

PR #614, opened 2018-02-14
Cloud metrics

### LINKED ISSUE #506: Cloud metrics, what is collected

Cloud users want to see how much load their cluster and individual nodes are experiencing. This can help debug bottlenecks, and identify whether scaling out is beneficial.

Show the different metrics gathered about each node, and how the statistics are displayed on the Cloud dashboard. Include screenshots.

A new section inside Cloud -> Monitoring

Docs in a similar style: [this](http://docs.aws.amazon.com/redshift/latest/mgmt/metrics-listing.html).

begriffs:
Fixes #506


begriffs:
Waiting to add docs for Datadog support. That feature should be available to test by next week.

cloud/monitoring.rst
```diff
@@ -1,6 +1,34 @@
...
+
+Citus Cloud metrics enable you to get information about your cluster's health and performance. The "Metrics" tab of the Cloud Console provides graphs for a number of measurements, all viewable per node:
+
+* CPU Utilization (Percent)
+   .. image:: ../images/metric-cpu.png
+* Network - Bytes In / Second
+   .. image:: ../images/metric-network-in.png
+* Network - Bytes Out / Second
+   .. image:: ../images/metric-network-out.png
+* Read IOPS
```
sumedhpathak: Hmm, I was expecting more information on what these mean. Can we link to someplace which describes this metric? Or is there information on AWS metrics pages which go into some detail?

Some are simple enough, e.g. CPU or Network bytes, but others like Avg Queue Length might be useful to have a little bit more information. 

Also, in the past, not all of these graphs have shown up for me. Is this because my cluster didn't have enough activity? Do we need to put some information around that?

lfittl: @sumedhpathak There are usually problems on the dev instances with some of this data - did you test on a dev instance, or production?

sumedhpathak: By dev, do you mean the 99$ cluster? It's been a while, so I don't recall exactly the cluster type. As a question, would all of these graphs be visible for all Cloud customers then?

lfittl: Yeah, thats the one I mean - I believe some of the disk statistics are not available for the developer plan customers. I don't think its worth mentioning, because this is subject to change at any time anyway.

begriffs: @lfittl 

* Can you tell me what "Avg Queue Length" means? I searched the postgres docs  but nothing obvious jumped out at me.
* For "IOPs" I'll see how AWS docs describe that measurement, and use similar language. "Input/output operations per second" doesn't really mean anything to me unless I know the size of data those operations are moving at a time.

begriffs:
OK, I added more explanations of the metrics, and grouped them together. For instance the grouping puts related EBS I/O metrics together to better relate IOPS, the IO queue, bytes/sec etc.

PR #618, opened 2018-02-22
Use %s rather than %I

### LINKED ISSUE #617: Documentation for run_command_on_shards() doesn't match example
@taybin opened up an issue in the Citus repository: https://github.com/citusdata/citus/issues/2014 I'm copy/pasting the issue below.

https://docs.citusdata.com/en/v7.2/extra/query_propagation.html#running-on-all-shards

The documentation says:
> Notice the %I which will be replaced with each shard’s name.

but the example code uses `%s`

This also makes the documentation for `run_command_on_colocated_placements()` confusing because it uses both `%s` and `%I` and doesn't explain which is for which table.

begriffs:
Because schema-qualified names should not quote around the dot.

Fixes #617 
Fixes https://github.com/citusdata/citus/issues/2014


begriffs:
I made the text match the example, and tested the examples.

PR #620, opened 2018-02-26
Configuring server params from cloud console

### LINKED ISSUE #601: Document what postgres/pgbouncer settings can be configured in Cloud

When managing a server locally, an administrator can tune settings in postgresql.conf. Some of these settings will become configurable on Citus Cloud as well, at least settings which don't require a server restart to take effect.

* Provide a list of the configurable settings, with links to PostgreSQL docs to explain what they do
* Instructions for how to change them on Cloud, possibly with a screenshot

New subsection under Cloud

## How does this work? (devs)

@aamederen can you tell me which settings will be configurable?

begriffs:
Fixes #601 

The original issue specified including a list of the configurable settings, with links to PostgreSQL docs explaining what they do. However the UI in Cloud will provide those links, right @aamederen? I'm currently not seeing the postgres docs link under the settings selection box, but assuming that the change has not yet been deployed.

Also I have a section about the units for numerical settings, but perhaps units will soon be allowed and my documentation about that won't be necessary.


aamederen:
Hey @begriffs, You are right, the UI is going to provide the links and small descriptions, but that change is not deployed yet.

About the units, it is not a super easy change so it may take a while to put it in place. As I mentioned during the presentation, I tried to mention the units in the short descriptions of the parameters.

sumedhpathak:
@aamederen Could you review this change? Thanks!

aamederen:
hey @begriffs, sorry for not responding. There are a few other things that is likely to change about custom configs. I will review this after the pending changes are deployed.

I assume this has no urgency since custom configs feature is not enabled to any of the users yet.

cloud/performance.rst
```diff
@@ -86,3 +86,30 @@ To measure the number of active connections at a given time, run:
...
+
+The options are grouped by which server and system they control:
+
+* Coordinator PostgreSQL
+* Coordinator Inbound PgBouncer
+* Worker PostgreSQL
+
+Selecting an option opens an input box that accepts the appropriate values, whether numerical, textual, or a pre-set list. The selected option also shows a link to learn more about the configuration parameter.
+
+.. image:: ../images/cloud-pg-conf-option.png
```
aamederen: It would be good if we can re-select the images. For example, we removed the `debug_print_plan` option because it can lead to oversized logs and out-of-disk issues.

aamederen: Also, this screenshot can cover the short description and the link.

cloud/performance.rst
```diff
@@ -86,3 +86,30 @@ To measure the number of active connections at a given time, run:
...
+
+Numerical parameters do not yet allow specifying units in this interface, and are interpreted as their default unit. A reference of the default units is available in the database from the ``pg_settings`` table:
```
aamederen: The `short description` of tune settings mention the units most of the time, even if the actual documentation doesn't.

cloud/performance.rst
begriffs: Done

begriffs: I updated the screenshot and added some info about how each numeric option's description displays the default unit.

begriffs:
@aamederen this is ready for another review.

cloud/performance.rst
```diff
@@ -88,3 +88,24 @@ To measure the number of active connections at a given time, run:
...
+In addition to adding nodes and scaling their hardware, the Cloud Console allows changing database server configuration parameters. Adjusting the parameters can help tune the server for particular workloads. Cloud currently exposes just those parameters which do not require a server restart to take effect.
+
+To access these settings, go to the Configuration tab in your Cloud formation and click "Change a Parameter." It will present a dropdown list of config parameters:
+
+.. image:: ../images/cloud-pg-conf.png
+
+The options are grouped by which server and system they control:
+
+* Coordinator PostgreSQL
+* Coordinator Inbound PgBouncer
```
aamederen: We will also allow Coordinator Outbound PgBouncer soon but we can add it once we enable it.

PR #621, opened 2018-02-27
Out with the broadcast, in with the reference

### LINKED ISSUE #238: Remove references to broadcast joins and large_table_shard_count
#### Broadcast joins

When we talk about joins and broadcasted tables in our docs, we refer to the dated ``citus.large_table_shard_count`` config setting. Citus uses the new reference table API and doesn't rely on this setting anymore.

https://docs.citusdata.com/en/v6.0/dist_tables/querying.html#joins

We should update the following paragraph to reflect this change:

"To determine the best join strategy, Citus treats large and small tables differently while executing JOINs. The distributed tables are classified as large and small on the basis of the configuration entry citus.large_table_shard_count (default value: 4). The tables whose shard count exceeds this value are considered as large while the others small. In practice, the fact tables are generally the large tables while the dimension tables are the small tables."

#### Colocated joins

We have the following Note in the colocated joins section:

"In order to benefit most from colocated joins, you should hash distribute your tables on the join key and use the same number of shards for both tables.  ..."

I had two questions on this note. First, we changed our table creation APIs in v6.0 and we could probably reframe this sentence to the following: "In order to benefit most from colocated joins, you should hash distribute your tables on the join key, where the join keys have the same data types."

Second, this note communicates information that's very similar to the paragraph before it. Do we need to mention it as a Note?

begriffs:
Fixes #238

### DIFF
--- a/dist_tables/querying.rst
+++ b/dist_tables/querying.rst
@@ -91,33 +91,28 @@ LIMIT approximations are disabled by default and can be enabled by setting the c
 Joins
 #####
 
-Citus supports equi-JOINs between any number of tables irrespective of their size and distribution method. The query planner chooses the optimal join method and join order based on the statistics gathered from the distributed tables. It evaluates several possible join orders and creates a join plan which requires minimum data to be transferred across network.
-
-To determine the best join strategy, Citus treats large and small tables differently while executing JOINs. The distributed tables are classified as large and small on the basis of the configuration entry citus.large_table_shard_count (default value: 4). The tables whose shard count exceeds this value are considered as large while the others small. In practice, the fact tables are generally the large tables while the dimension tables are the small tables.
-
-Broadcast joins
-----------------
-
-This join type is used while joining small tables with each other or with a large table. This is a very common use case where you want to join the keys in the fact tables (large table) with their corresponding dimension tables (small tables). Citus replicates the small table to all workers where the large table's shards are present. Then, all the joins are performed locally on the workers in parallel. Subsequent join queries that involve the small table then use these cached shards.
+Citus supports equi-JOINs between any number of tables irrespective of their size and distribution method. The query planner chooses the optimal join method and join order based on how tables are distributed. It evaluates several possible join orders and creates a join plan which requires minimum data to be transferred across network.
 
 Co-located joins
-----------------------------
+----------------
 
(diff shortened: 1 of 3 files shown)


dist_tables/querying.rst
```diff
@@ -91,34 +91,29 @@ LIMIT approximations are disabled by default and can be enabled by setting the c
...
+Citus supports equi-JOINs between any number of tables irrespective of their size and distribution method. The query planner chooses the optimal join method and join order based on how tables are distributed. It evaluates several possible join orders and creates a join plan which requires minimum data to be transferred across network.
 
 Co-located joins
-----------------------------
+----------------
 
 To join two large tables efficiently, it is advised that you distribute them on the same columns you used to join the tables. In this case, the Citus coordinator knows which shards of the tables might match with shards of the other table by looking at the distribution column metadata. This allows Citus to prune away shard pairs which cannot produce matching join keys. The joins between remaining shard pairs are executed in parallel on the workers and then the results are returned to the coordinator.
```
ozgune: Two quick questions.

1. Instead of removing this note, could we merge it into the previous paragraph?
2. We saw a user error with colocated joins recently. Users decide to distribute their tables on the join key, but the join keys have different data types (for example, integer and bigint). Do we want to mention in this note that the join key data types have to be the same? For example, what if we said the following instead: "In order to benefit most from colocated joins, you should hash distribute your tables on the join key, where the join keys have the same data types."

dist_tables/querying.rst
```diff
@@ -91,34 +91,29 @@ LIMIT approximations are disabled by default and can be enabled by setting the c
...
-----------------------------
+-----------------
 
 In some cases, you may need to join two tables on columns other than the distribution column. For such cases, Citus also allows joining on non-distribution key columns by dynamically repartitioning the tables for the query.
 
 In such cases the table(s) to be partitioned are determined by the query optimizer on the basis of the distribution columns, join keys and sizes of the tables. With repartitioned tables, it can be ensured that only relevant shard pairs are joined with each other reducing the amount of data transferred across network drastically.
 
 In general, co-located joins are more efficient than repartition joins as repartition joins require shuffling of data. So, you should try to distribute your tables by the common join keys whenever possible.
 
+Reference joins
```
ozgune: Two minor notes.

1. Rename "Reference joins" -> "Reference table joins"?
2. Maybe move the "Reference table joins" section to before "Repartition joins"?

PR #625, opened 2018-03-05
Include tips for seeing good COPY performance

### LINKED ISSUE #599: Instructions for getting maximal \copy benchmark numbers

When people measure the COPY command they might observe lower numbers than we list. This is because their cluster or test setup is not well configured. We should provide pointers to help people get good performance.

https://docs.citusdata.com/en/v7.2/performance/scaling_data_ingestion.html#bulk-copy-250k-2m-s

## How does this work? (devs)

Suggestions:

* We recommend a large batch size (~ 50000-100000). You can benchmark with multiple files (1,10,1000,10000 etc), each of that batch size
* Parallel ingestion - Increase number of threads/ingestors to 2,4,8,16 and run benchmarks.
* Use a compute optimized coordinator - say 8-16 vcpus. For the workers you could choose memory optimized boxes with decent number of cpus (say 4-8vcpus etc).
* Shard count - go with a smaller count - 32 should suffice, you could benchmark with 64 too.
* You should ingest data for a suitable amount of time (say 2hrs, 4hrs ,8 hrs,24 hrs etc). This test would be more representative of production setup.

begriffs:
Fixes #599

### DIFF
--- a/performance/scaling_data_ingestion.rst
+++ b/performance/scaling_data_ingestion.rst
@@ -119,7 +119,7 @@ From a throughput standpoint, you can expect data ingest ratios of 250K - 2M row
 
   * We recommend a large batch size (~ 50000-100000). You can benchmark with multiple files (1, 10, 1000, 10000 etc), each of that batch size.
   * Use parallel ingestion. Increase the number of threads/ingestors to 2, 4, 8, 16 and run benchmarks.
-  * Use a compute-optimized coordinator, 8-16 vcpus. For the workers choose memory-optimized boxes with a decent number of vcpus (4-8).
+  * Use a compute-optimized coordinator. For the workers choose memory-optimized boxes with a decent number of vcpus.
   * Go with a relatively small shard count, 32 should suffice but you could benchmark with 64, too.
   * Ingest data for a suitable amount of time (say 2, 4, 8, 24 hrs). Longer tests are more representative of a production setup.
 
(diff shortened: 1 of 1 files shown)


performance/scaling_data_ingestion.rst
```diff
@@ -115,7 +115,13 @@ From a throughput standpoint, you can expect data ingest ratios of 250K - 2M row
 
 .. note::
 
-    To avoid opening too many connections to worker nodes, we recommend running only two COPY commands on a distributed table at a time. In practice, running more than four at a time rarely results in performance benefits. An exception is when all the data in the ingested file has a specific partition key value, which goes into a single shard. COPY will only open connections to shards when necessary.
+  Make sure your benchmarking setup is well configured so you can observe optimal COPY performance. Follow these tips:
+
+  * We recommend a large batch size (~ 50000-100000). You can benchmark with multiple files (1, 10, 1000, 10000 etc), each of that batch size.
+  * Use parallel ingestion. Increase the number of threads/ingestors to 2, 4, 8, 16 and run benchmarks.
+  * Use a compute-optimized coordinator, 8-16 vcpus. For the workers choose memory-optimized boxes with a decent number of vcpus (4-8).
```
saicitus: Can we not specify the actual numbers 8-16 vcpus and 4-8 etc, just because hardware differs from usecase to usecase.

begriffs: Just leave it at compute-optimized for coordinator and memory-optimized for workers?

sumedhpathak:
@saicitus I presume you are reviewing this PR? Have assigned this to you.

begriffs:
I'm going to merge this and we can adjust it if needed.

PR #626, opened 2018-03-06
Two little fixes

### LINKED ISSUE #580: Misinformation about disabling distributed deadlock detection
The following excerpt has a misinformation about disabling distributed deadlock detection: 

> citus.distributed_deadlock_detection_factor (floating point)
Sets the time to wait before checking for distributed deadlocks. In particular the time to wait will be this value multiplied with PostgreSQL’s deadlock_timeout setting. **If the detection factor is set to its maximum value of 1000, then distributed deadlock detection is disabled**. The default value is 2.

The correct form is, distributed deadlock detection is disabled when `citus.distributed_deadlock_detection_factor = -1`

### LINKED ISSUE #582: Minor query fix on dml page
As part of the release testing I have been going through the queries in docs. I found the below SQL in [dml page](https://docs.citusdata.com/en/latest/dist_tables/dml.html#updates-and-deletion)

```SQL
DELETE FROM github_events
WHERE repo_id IN (24509048, 24509049);

UPDATE github_events
SET event_public = TRUE
WHERE org = 'happy_devs';
```

which errors out with 

```SQL
ERROR:  22P02: invalid input syntax for type json
LINE 3: WHERE org = 'happy_devs';
                    ^
DETAIL:  Token "happy_devs" is invalid.
CONTEXT:  JSON data, line 1: happy_devs
LOCATION:  report_invalid_token, json.c:1248
```

obviously it is not happy with org (which is defined as jsonb type previously in the same page.

begriffs:
Fixes #580 
Fixes #582

### DIFF
--- a/reference/configuration.rst
+++ b/reference/configuration.rst
@@ -57,7 +57,7 @@ Whether to log distributed deadlock detection related processing in the server l
 citus.distributed_deadlock_detection_factor (floating point)
 $$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$
 
-Sets the time to wait before checking for distributed deadlocks. In particular the time to wait will be this value multiplied by PostgreSQL's ``deadlock_timeout`` setting. The default value is ``2``. A value of ``-1`` disables distributed deadlock detection.
+Sets the time to wait before checking for distributed deadlocks. In particular the time to wait will be this value multiplied by PostgreSQL's `deadlock_timeout <https://www.postgresql.org/docs/current/static/runtime-config-locks.html>`_ setting. The default value is ``2``. A value of ``-1`` disables distributed deadlock detection.
 
 citus.max_task_string_size (integer)
 ------------------------------------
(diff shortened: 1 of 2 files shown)


reference/configuration.rst
```diff
@@ -57,7 +57,7 @@ Whether to log distributed deadlock detection related processing in the server l
 citus.distributed_deadlock_detection_factor (floating point)
 $$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$
 
-Sets the time to wait before checking for distributed deadlocks. In particular the time to wait will be this value multiplied with PostgreSQL's deadlock_timeout setting. If the detection factor is set to its maximum value of 1000, then distributed deadlock detection is disabled. The default value is 2.
+Sets the time to wait before checking for distributed deadlocks. In particular the time to wait will be this value multiplied by PostgreSQL's ``deadlock_timeout`` setting. The default value is ``2``. A value of ``-1`` disables distributed deadlock detection.
```
sumedhpathak: Can this link to PostgreSQL's page? https://www.postgresql.org/docs/current/static/runtime-config-locks.html (for deadlock_timeout).

begriffs: (added link)

PR #629, opened 2018-03-07
Workaround for upsert into … select

### LINKED ISSUE #560: INSERT SELECT via coordinator doesn’t support upserts

Warn people it won't work. `ON CONFLICT` fails when you run the query on the coordinator. Scan for any places we might contradict that. Also check whether plain `INSERT .. ON CONFLICT` is good, because we mention that too.

* Might be time to add INSERT SELECT to the DML section as a subsection. It is mentioned in passing other places but is a significant form of DML. Include the on conflict warning as a highlighted note.
* The workarounds section if there's a workaround (perhaps filtering by a single tenant).

begriffs:
Fixes #560

### DIFF
--- a/develop/reference_dml.rst
+++ b/develop/reference_dml.rst
@@ -149,8 +149,6 @@ The situation changes when dealing with late arriving data, or running the rollu
     ON CONFLICT (day, url, site_id) DO UPDATE SET
       view_count = daily_page_views.view_count + EXCLUDED.view_count;
 
-(In reality this particular construction needs a slight modification on Citus, see the :ref:`upsert_into_select` workaround.)
-
 Updates and Deletion
 --------------------
 
--- a/develop/reference_dml.rst
+++ b/develop/reference_dml.rst
@@ -149,6 +149,8 @@ The situation changes when dealing with late arriving data, or running the rollu
     ON CONFLICT (day, url, site_id) DO UPDATE SET
       view_count = daily_page_views.view_count + EXCLUDED.view_count;
 
+(In reality this particular construction needs a slight modification on Citus, see the :ref:`upsert_into_select` workaround.)
+
 Updates and Deletion
 --------------------
 
--- a/develop/reference_workarounds.rst
+++ b/develop/reference_workarounds.rst
@@ -72,13 +72,15 @@ Remember that the coordinator will send the results in the subquery or CTE to al
 INSERT…SELECT upserts
 ---------------------
 
-Currently "on conflict" upserts are not supported on the coordinator for "insert into … select" statements. Attempting to upsert into distributed tables this way produces an error:
+For those INSERT…SELECT statements including a GROUP BY clause which contains the distribution column, Citus can push the whole statement down to workers for execution. Otherwise Citus uses :ref:`push_pull_execution` to pull the results of the SELECT query from workers to the coordinator, and push the rows for insertion back down to the workers. This latter mode does not support the ON CONFLICT clause (Postgres' way to do upserts).
+
+Attempting to execute these kind of queries will generate an error message:
 
 ::
 
   ERROR: ON CONFLICT is not supported in INSERT ... SELECT via coordinator
 
-The workaround is to materialize the select query in a temporary distributed table, and upsert from there.
+The ideal solution is to model the data with a distribution column that can be part of the GROUP BY clause in the upsert statement. However if this is not feasible then the workaround is to materialize the select query in a temporary distributed table, and upsert from there.
 
 .. code-block:: postgresql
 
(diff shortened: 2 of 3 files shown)


develop/reference_dml.rst
```diff
@@ -149,6 +149,8 @@ The situation changes when dealing with late arriving data, or running the rollu
     ON CONFLICT (day, url, site_id) DO UPDATE SET
       view_count = daily_page_views.view_count + EXCLUDED.view_count;
 
+(In reality this particular construction needs a slight modification on Citus, see the :ref:`upsert_into_select` workaround.)
```
sumedhpathak: Only Insert/Selects which don't get pushed down entirely need the construct. Do you know if this query needs to be pulled up? I think site_id is the partition column, so it seems this query will just work. Could you try it out?

develop/reference_workarounds.rst
```diff
@@ -67,6 +67,65 @@ Although you can't join such tables directly, by wrapping the local table in a s
...
+  CREATE UNLOGGED TABLE temp_table (LIKE dest_table);
+  SELECT create_distributed_table('temp_table', 'tenant_id');
+  INSERT INTO temp_table <query>;
+  INSERT INTO dest_table SELECT * FROM temp_table <upsert clause>;
+  DROP TABLE temp_table;
+  END;
+
+.. _window_func_workaround:
+
+Window Functions
```
sumedhpathak: Why do we have this section? Is this new content, or just something which was moved around?

develop/reference_workarounds.rst
```diff
@@ -67,6 +67,65 @@ Although you can't join such tables directly, by wrapping the local table in a s
 
 Remember that the coordinator will send the results in the subquery or CTE to all workers which require it for processing. Thus it's best to either add the most specific filters and limits to the inner query as possible, or else aggregate the table. That reduces the network overhead which such a query can cause. More about this in :ref:`subquery_perf`.
 
+.. _upsert_into_select:
+
+INSERT…SELECT upserts
+---------------------
+
+Currently "on conflict" upserts are not supported on the coordinator for "insert into … select" statements. Attempting to upsert into distributed tables this way produces an error:
```
sumedhpathak: We should probably expand this a bit. INSERT/SELECT has two modes: 1) It gets pushed down entirely if we are grouping by the partition column. or 2) We run the select on the workers, pull data to the coordinator, and then run inserts from the coordinator to the workers again. This is chosen if we aren't grouping by the partition column.

Its option 2 above which doesn't support ON CONFLICT. So we should clarify that and recommend modeling the data/choosing partition column so we can choose option 1. (Since that is also most efficient). And offer this workaround as a backup if data modeling isn't possible.

develop/reference_dml.rst
begriffs: You're right, it worked. I should have tested it earlier.

develop/reference_workarounds.rst
```diff
@@ -67,6 +67,34 @@ Although you can't join such tables directly, by wrapping the local table in a s
 
 Remember that the coordinator will send the results in the subquery or CTE to all workers which require it for processing. Thus it's best to either add the most specific filters and limits to the inner query as possible, or else aggregate the table. That reduces the network overhead which such a query can cause. More about this in :ref:`subquery_perf`.
 
+.. _upsert_into_select:
+
+INSERT…SELECT upserts
```
sumedhpathak: I'd rename this to include "not grouped on partition column" or some such.

develop/reference_workarounds.rst
```diff
@@ -67,6 +67,34 @@ Although you can't join such tables directly, by wrapping the local table in a s
...
+---------------------
+
+For those INSERT…SELECT statements including a GROUP BY clause which contains the distribution column, Citus can push the whole statement down to workers for execution. Otherwise Citus uses :ref:`push_pull_execution` to pull the results of the SELECT query from workers to the coordinator, and push the rows for insertion back down to the workers. This latter mode does not support the ON CONFLICT clause (Postgres' way to do upserts).
+
+Attempting to execute these kind of queries will generate an error message:
+
+::
+
+  ERROR: ON CONFLICT is not supported in INSERT ... SELECT via coordinator
```
sumedhpathak: Could we also add this in Common Error Messages? (And link to this section for workaround?)

develop/reference_workarounds.rst
```diff
@@ -67,6 +67,34 @@ Although you can't join such tables directly, by wrapping the local table in a s
 
 Remember that the coordinator will send the results in the subquery or CTE to all workers which require it for processing. Thus it's best to either add the most specific filters and limits to the inner query as possible, or else aggregate the table. That reduces the network overhead which such a query can cause. More about this in :ref:`subquery_perf`.
 
+.. _upsert_into_select:
+
+INSERT…SELECT upserts
+---------------------
+
+For those INSERT…SELECT statements including a GROUP BY clause which contains the distribution column, Citus can push the whole statement down to workers for execution. Otherwise Citus uses :ref:`push_pull_execution` to pull the results of the SELECT query from workers to the coordinator, and push the rows for insertion back down to the workers. This latter mode does not support the ON CONFLICT clause (Postgres' way to do upserts).
```
sumedhpathak: A correction here. We need to make sure we select partition column from one table into the partition column of another table. If there is an agg, then the partition column needs to be in the group-by clause.
See https://docs.google.com/presentation/d/1mxoLq5d8-jhO9XjYR7I4oJCUo5r2eaqSZPNkepO4yQc/edit#slide=id.g18d75ba220_0_28. Could we amend this description?

PR #631, opened 2018-03-08
Document more UDFs

### LINKED ISSUE #595: Document master_update_node()

### LINKED ISSUE #604: master_move_shard_placement is not documented
master_move_shard_placement is not documented in the section Citus Utility Function Reference.
https://docs.citusdata.com/en/v7.2/reference/user_defined_functions.html

begriffs:
Fixes #595 
Fixes #604

### DIFF
--- a/reference/user_defined_functions.rst
+++ b/reference/user_defined_functions.rst
@@ -355,6 +355,7 @@ Example
           7 |       7 | new-node |    12345 | default  | f           | t        |       0 | primary  | default
     (1 row)
 
+.. _master_update_node:
 
 master_update_node
 $$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$
--- a/reference/user_defined_functions.rst
+++ b/reference/user_defined_functions.rst
@@ -831,7 +831,7 @@ $$$$$$$$$$$$$$$$$$$$$$$$$$$$$
 
   The master_move_shard_placement function is a part of Citus Enterprise. Please `contact us <https://www.citusdata.com/about/contact_us>`_ to obtain this functionality.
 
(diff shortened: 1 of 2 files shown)


reference/user_defined_functions.rst
```diff
@@ -355,6 +355,33 @@ Example
           7 |       7 | new-node |    12345 | default  | f           | t        |       0 | primary  | default
     (1 row)
 
+
+master_update_node
+$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$
+
+The master_update_node() function changes the hostname and port for a node registered in the Citus metadata table :ref:`pg_dist_node <pg_dist_node>`.
```
sumedhpathak: This came up in a few customer conversations. Could you also add an FAQ Entry for this? What scenarios would one want to do this?

reference/user_defined_functions.rst
```diff
@@ -796,6 +823,50 @@ The example below will repair an inactive shard placement of shard 12345 which i
 
     SELECT master_copy_shard_placement(12345, 'good_host', 5432, 'bad_host', 5432);
 
+master_move_shard_placement
```
sumedhpathak: While there might be some scenarios, this function isn't generally meant to be called manually. That said, I'm not opposed to adding this in. Might be useful to note though that this is called as part of rebalance-table-shards, and its not necessary to be called directly.

reference/user_defined_functions.rst
begriffs: ✅

reference/user_defined_functions.rst
begriffs: Added a sentence about that to the intro paragraph.

PR #633, opened 2018-03-15
Bump Citus CF template to 7.3.0

velioglu:


begriffs:
I deployed the template and confirmed the citus version.

Issue #634, opened 2018-03-19
Show example of citus.multi_task_query_log_level


begriffs:
We discuss the valid options for this enum in https://docs.citusdata.com/en/v7.3/reference/configuration.html#citus-multi-task-query-log-level-enum

However Craig suggested that we also give examples of what appears in the logs as a result of setting the GUC.

begriffs:
Somewhat related observation, just adding it to this issue as a place to take care of it:

A lot of people seem to run `set foo = bar` when they should run `alter database set foo = bar`. Find where to clarify this in the docs.

PR #638, opened 2018-03-26
adding example of how to alter db and adjusting log levels

craigkerstiens:


begriffs:
There are also some uses of this in https://citus-doc-preview-pr-638.herokuapp.com/migration/transitioning.html#validating-query-migration that probably should be updated too?

reference/configuration.rst
```diff
@@ -213,16 +220,15 @@ The supported values for this enum are:
 
 * **debug:** Logs statement at DEBUG severity level.
 
-* **log:** Logs statement at LOG severity level.
+* **log:** Logs statement at LOG severity level. Log severity level will contain both the error, followed by :code:`STATEMENT` which contains the SQL query that was run.
```
begriffs: I'm not sure I understand the added comment here. Is there special behavior for `multi_task_query_log_level=log` not shared by the other levels?

begriffs:
Fixes #634

reference/configuration.rst
craigkerstiens: Log includes the statement following it, warning it is suppressed by default on cloud.

PR #648, opened 2018-04-05
Publicize manual query propagation

### LINKED ISSUE #150: Add FAQ about having to run per-server commands on all nodes
Citus does not automatically propagate commands like `CREATE DATABASE, CREATE EXTENSION, CREATE TYPE, CREATE FUNCTION etc` to the workers if run on the master.

If the user runs these commands on the workers, then these constructs are supported. However, at times users run these commands only on the master and hence get the impression that Citus doesn't support the above.

We should document this and guide the users to run these commands on all nodes.

### LINKED ISSUE #559: How to propagate role/function creation

In our Cloud docs we talk about creating new users by using the GUI. However in custom installations people need to know that the coordinator currently does not propagate `CREATE USER` to the workers. Show how to do this.

Same thing with `CREATE FUNCTION.`

* Toward the end of the DDL section, near https://docs.citusdata.com/en/v7.1/dist_tables/ddl.html#manual-modification
* If we had a non-Cloud section about users and permissions it would go there. Not sure this is enough content to warrant creating such a section.

### LINKED ISSUE #643: Move manual propagation section out of extra/
Add it into a discoverable place in docs because it's frequently useful.

This is already done in the menu reorganization PR, but we'd like to get this merged sooner.

begriffs:
Fixes #150 
Fixes #559 
Fixes #643 

This PR retains the propagation section in `extra/` to preserve any links we have shared with people to those sections. In the menu redesign we might consider adding a page redirect for that section.

### DIFF
--- a/develop/reference_ddl.rst
+++ b/develop/reference_ddl.rst
@@ -303,11 +303,4 @@ Adding an index takes a write lock, which can be undesirable in a multi-tenant "
 Manual Modification
 ~~~~~~~~~~~~~~~~~~~
 
-Currently other DDL commands are not auto-propagated, however you can propagate the changes manually using this general four-step outline:
-
-1. Begin a transaction and take an ACCESS EXCLUSIVE lock on coordinator node against the table in question.
-2. In a separate connection, connect to each worker node and apply the operation to all shards.
-3. Disable DDL propagation on the coordinator and run the DDL command there.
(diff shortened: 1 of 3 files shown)


begriffs:
Rebased this on master which now has the query prop section exposed. This PR now just adds an FAQ entry about it.

PR #654, opened 2018-05-02
Simplify rollups

### LINKED ISSUE #586: Simplify rollups in realtime use-case guide

The queries in our use-case guide demonstrate cool features, but they should be simplified to make them easier to understand and also to match best-practices better.

* what we have for rollups are pretty complex, the insert into select has a join… what about a simpler one?
* Save the `minute` as a variable first, and remove subquery:
   ```sql
   WHERE minute > (
     SELECT COALESCE(max(ingest_time), timestamp '10-10-1901')
     FROM http_request_1min
     WHERE http_request_1min.site_id = h.site_id
   )
   ```
* Usually people just use a single-row table to hold the previous rollup time and consult it rather than running a `max()` scan.
* Remove the other subquery that creates a `minute` column and just use the literal `date_trunc('minute', ingest_time)` everywhere.

begriffs:
Fixes #586

### DIFF
--- a/use_cases/realtime_analytics.rst
+++ b/use_cases/realtime_analytics.rst
@@ -160,31 +160,39 @@ The following function wraps the rollup query up for convenience.
 
 .. code-block:: plpgsql
 
+  -- single-row table to store when we rolled up last
+  CREATE TABLE latest_rollup (
+    minute timestamptz PRIMARY KEY,
+
+    EXCLUDE USING gist (minute WITH <>),
+    CHECK (minute = date_trunc('minute', minute))
+  );
+
+  -- initialize to a time long ago
+  INSERT INTO latest_rollup VALUES ('10-10-1901');
+
+  -- function to do the rollup
   CREATE OR REPLACE FUNCTION rollup_http_request() RETURNS void AS $$
+  DECLARE
+    this_minute timestamptz := date_trunc('minute', now());
+    since       timestamptz := minute from latest_rollup;
   BEGIN
     INSERT INTO http_request_1min (
       site_id, ingest_time, request_count,
       success_count, error_count, average_response_time_msec
     ) SELECT
       site_id,
-      minute,
+      date_trunc('minute', ingest_time),
       COUNT(1) as request_count,
       SUM(CASE WHEN (status_code between 200 and 299) THEN 1 ELSE 0 END) as success_count,
       SUM(CASE WHEN (status_code between 200 and 299) THEN 0 ELSE 1 END) as error_count,
       SUM(response_time_msec) / COUNT(1) AS average_response_time_msec
-    FROM (
-      SELECT *,
-        date_trunc('minute', ingest_time) AS minute
-      FROM http_request
-    ) AS h
-    WHERE minute > (
-      SELECT COALESCE(max(ingest_time), timestamp '10-10-1901')
-      FROM http_request_1min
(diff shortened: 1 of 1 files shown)


use_cases/realtime_analytics.rst
```diff
@@ -160,31 +160,39 @@ The following function wraps the rollup query up for convenience.
 
 .. code-block:: plpgsql
 
+  -- single-row table to store when we rolled up last
```
begriffs: Really [schema variables](http://www.postgresql-archive.org/proposal-schema-variables-td5989266.html) would be better, but they're not yet available.

use_cases/realtime_analytics.rst
```diff
@@ -160,31 +160,39 @@ The following function wraps the rollup query up for convenience.
...
-    WHERE minute > (
-      SELECT COALESCE(max(ingest_time), timestamp '10-10-1901')
-      FROM http_request_1min
-      WHERE http_request_1min.site_id = h.site_id
-    )
-      AND minute <= date_trunc('minute', now())
-    GROUP BY site_id, minute
-    ORDER BY minute ASC;
+    FROM http_request
+    WHERE date_trunc('minute', ingest_time) <@
```
begriffs: Should probably add an index for the expressaion `date_trunc('minute', ingest_time)` on http_request, but date_trunc is not immutable on timestamptz so Postgres forbids it.

use_cases/realtime_analytics.rst
begriffs: Actually there's another technique: directly inspect the timestamp of the last insertion commit.

```sql
select pg_xact_commit_timestamp(xmin) from http_request_1min;
```

Then we wouldn't have to maintain our own metadata table...

use_cases/realtime_analytics.rst
```diff
@@ -160,31 +160,39 @@ The following function wraps the rollup query up for convenience.
 
 .. code-block:: plpgsql
 
+  -- single-row table to store when we rolled up last
+  CREATE TABLE latest_rollup (
+    minute timestamptz PRIMARY KEY,
+
+    EXCLUDE USING gist (minute WITH <>),
```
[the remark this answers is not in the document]
begriffs: The pkey prevents there being two rows with differing values of `minute` and the exclusion constraint prevents there being two rows with the same value. The two together enforce that there is at most one row in the table, so it acts like more of a single variable than a table.

Is this overkill and does it obscure the point of this example? Maybe…

use_cases/realtime_analytics.rst
```diff
@@ -160,31 +160,39 @@ The following function wraps the rollup query up for convenience.
...
+    CHECK (minute = date_trunc('minute', minute))
+  );
+
+  -- initialize to a time long ago
+  INSERT INTO latest_rollup VALUES ('10-10-1901');
+
+  -- function to do the rollup
   CREATE OR REPLACE FUNCTION rollup_http_request() RETURNS void AS $$
+  DECLARE
+    this_minute timestamptz := date_trunc('minute', now());
```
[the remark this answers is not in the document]
begriffs: Makes sense, sure.

use_cases/realtime_analytics.rst
begriffs: Alternately I could use two comparisons rather than checking inclusion in a range. I thought the range was clever and the code shorter, but not if it confuses people.

PR #656, opened 2018-05-04
Document TopN

### LINKED ISSUE #555: How to use TopN approximation

Finding an approximation of the most frequent items of a column of data, using a limited amount of memory.

Show how to use it, how to customize the memory limit, and how to take the union of two measurements.

Perhaps a new section under https://docs.citusdata.com/en/v7.1/dist_tables/querying.html#aggregate-functions

### Example sql

There is a full example in the readme for https://github.com/citusdata/topn

begriffs:
Fixes #555


develop/reference_sql.rst
```diff
@@ -70,6 +70,132 @@ As an example, if you want to run the hll_union aggregate function on your data
 
 You can then call sum(hll_column) to roll up those columns within the database. Please note that these custom aggregates need to be created both on the coordinator and the workers.
 
+.. _topn:
+
+Estimating Top N Items
+~~~~~~~~~~~~~~~~~~~~~~
+
+Similar to how HLL efficiently estimates the number of distinct items in a set, the `TopN PostgreSQL extension <https://github.com/citusdata/postgresql-topn>`_ simultaneously counts many items. It, too, stores an approximation in a bounded amount of space. It is also pre-installed on Citus Cloud.
```
sumedhpathak: Could you provide a better description of TopN. The current one just says: "the `TopN PostgreSQL extension <https://github.com/citusdata/postgresql-topn>`_ simultaneously counts many items.". 

Maybe describe that it is an approximation algorithm, and provide motivation on why it might be useful:
 For example, the section here: https://github.com/citusdata/postgresql-topn#why-use-topn has some good information on why top-N.

develop/reference_sql.rst
```diff
@@ -70,6 +70,132 @@ As an example, if you want to run the hll_union aggregate function on your data
...
+
+Fundamentally the extension tracks a multi-set as a JSON object.
```
sumedhpathak: Are the below two points necessary? Are you thinking of these as being more "reference" type information for the extension?

develop/reference_sql.rst
```diff
@@ -70,6 +70,132 @@ As an example, if you want to run the hll_union aggregate function on your data
...
+  │    item    │ frequency │
+  ├────────────┼───────────┤
+  │ 0375404368 │       217 │
+  │ 0345417623 │       217 │
+  │ 0375404376 │       217 │
+  │ 0375408738 │       217 │
+  │ 043936213X │       204 │
+  └────────────┴───────────┘
+
+For more details and examples see the TopN readme.
```
sumedhpathak: Link to the readme?

develop/reference_sql.rst
```diff
@@ -70,6 +70,132 @@ As an example, if you want to run the hll_union aggregate function on your data
...
+.. code-block:: bash
+
+  curl -L https://examples.citusdata.com/customer_reviews_2000.csv.gz | \
+    gunzip > reviews.csv
+
+Next, ingest it into the database:
+
+.. code-block:: postgresql
+
+  CREATE TABLE customer_reviews
```
sumedhpathak: This example doesn't distribute the table.

develop/reference_sql.rst
begriffs: I wanted to clarify what the extension really does, in addition to a scenario where it's used. When I first saw the repo I wasn't totally sure. Then I experimented with it a little and saw what the basic operation is, and that it is pretty simple.

begriffs:
> The example we have doesn't create a distributed table

Oops, I got so focused on the operation of the extension that I forgot that the point is working on distributed tables. Tried to modify the example and ran into a problem with `topn_union_agg` as you predicted. My attempted workaround of creating an alias for it called `sum` also ran into a problem as I elaborated in Slack.

I'll continue writing the section the way I anticipate it will work in 7.5 when the aggregate becomes supported, and we can test/merge then.

begriffs:
OK, I revised the intro and tested the code with distributed tables on Citus built from Furkan's pull request.

PR #666, opened 2018-05-10
Billing instructions

### LINKED ISSUE #659: ACH billing for cloud

A walkthrough of setting up ACH billing, with screenshots of UI.

https://docs.citusdata.com/en/v7.3/cloud/support.html

begriffs:
Fixes #659 

The current change is pretty minimal, just some screenshots of an arguably self-explanatory part of the UI. Is there something else we want to explain here?


begriffs:
@craigkerstiens assigning to you for review since you suggested the topic.

sumedhpathak:
I'll let Craig review it, but visually, the image and the titles blend together. Its hard to distinguish where the CC ends and the ACH section begins. You could break it up by putting some text, but I don't know what text we would add. Is there one method over the other we prefer? Also, are there other keywords we can add so this is searchable?

cloud/support.rst
```diff
@@ -25,6 +28,25 @@ Citus Cloud bills on a per minute basis. We bill for a minimum of one hour of us
...
+
+.. image:: ../images/cloud-bill-menu.png
+
+Then fill out the form for the desired payment option.
+
+Credit Card Option
+~~~~~~~~~~~~~~~~~~
+
+.. image:: ../images/cloud-bill-credit.png
+
```
sumedhpathak: Add more space or text to break up these two sections?

craigkerstiens:
Overall good with this, we need to add a bit around the ACH process, feel free to ship this and we can iterate on the other.

PR #682, opened 2018-05-18
Menu options collapsed by default

### LINKED ISSUE #405: Allow the top-level menu sections to be collapsible, so we can navigate the index easier.

begriffs:
Fixes #405 

Certainly looks different, takes a little getting used to.

### DIFF
--- a/articles/index.rst
+++ b/articles/index.rst
@@ -1,0 +1,18 @@
+Related Articles
+################
+
+.. toctree::
+   :maxdepth: 1
+
+   heroku_addon.rst
+   efficient_rollup.rst
+   hll_count_distinct.rst
+   scale_on_aws.rst
+   parallel_indexing.rst
+   aggregation.rst
+   outer_joins.rst
+   designing_saas.rst
+   metrics_dashboard.rst
+   sharding_mt_app.rst
+   semi_structured_data.rst
+   faceted_search.rst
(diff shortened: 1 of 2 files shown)


sumedhpathak:
Interesting. Is it possible to selectively open up a few of those? E.g, I'd like to keep FAQ/Articles closed, but maybe open up Get Started and/or others.

begriffs:
Looks like the only way to force them to expand is by including my own javascript. The code would add a `.current` class to menu options at time of page load.

However experimenting with that by running ad-hoc code reveals some problems.

screen shot 2018-05-21 at 4 15 51 pm

The screenshot shows me browsing the real-time dashboard guide and opening other menus with javascript. With the other menu items expanded it becomes confusing to determine my current place in the nav. Also, when I collapse one of the items with my mouse, the theme detects that something isn't right and collapses the other menu items too.

I can try splitting the menu into multiple `toctree` elements, each with one entry. That might fix the problem where the theme tries to enforce that just one is expanded at once. Doesn't fix the confusing looking UI though.

begriffs:
Maybe I could have my code add `.current` to the items to open them, and also another class which I can match with special CSS to differentiate it from the true current section.

begriffs:
@sumedhpathak I redid this branch to collapse only the articles menu item.

PR #683, opened 2018-05-18
Warn about setting shard count too high

begriffs:


admin_guide/cluster_management.rst
```diff
@@ -32,6 +32,12 @@ Real-Time Analytics Use-Case
 
 In the :ref:`rt_blurb` use-case, shard count should be related to the total number of cores on the workers. To ensure maximum parallelism, you should create enough shards on each node such that there is at least one shard per CPU core. We typically recommend creating a high number of initial shards, e.g. **2x or 4x the number of current CPU cores**. This allows for future scaling if you add more workers and CPU cores.
 
+However be careful to keep the shard count small enough that distributed queries won't starve for connections. Put another way, the upper bound on necessary connections must not exceed the max connections possible in the system:
+
+::
+
+  (max concurrent queries * shard count) ≤ (number of workers * max_connections per worker)
```
sumedhpathak: What was the motivation for this calculation? The phrasing seems a bit strong, and pgBouncer can make a difference. I'm wondering if we should revisit this with SE for a better recommendation?

begriffs: Craig suggested this in an email:
> We typically recommend creating a high number of initial shards, e.g. 2x or 4x the number of current CPU cores as long as maximum concurrent queries / shard count / workers < worker max connections.

What all those slashes meant wasn't clear to me, whether it meant division or "and" and I chatted with Sai to clarify. He told me:
> the idea is `(maximum concurrent queries * shard count) / workers`  would be the max number of connections generated to each worker

I turned that into the final inequality that you saw above. To double-check it with him I assigned Sai as a reviewer to this PR as well.

saicitus: Hey @begriffs the wording seems to be strong i.e `must not exceed`. We could soften and keep it more open ended by saying something like below:
"Just keep in mind that the queries that get parallelized across shards can generate `maximum concurrent queries * shard count` connections across all the nodes. If you have too high a shard count there is potential that this setup generates large number of connections to the worker nodes which might not be optimal."

begriffs:
@sumedhpathak 5⏰

PR #681, opened 2018-05-17
More info about metrics

begriffs:
but remove pictures, not too valuable

Fixes #674

PR #684, opened 2018-05-23
More info about metrics

### LINKED ISSUE #674: Explain meaning of cloud metrics
In https://docs.citusdata.com/en/v7.4/cloud/monitoring.html we list a bunch of metrics, but some of them are obscure. Provide description of what they mean (available from AWS docs).

Also improve style a bit, remove bullet points.

begriffs:
Fixes #674

Same as PR #681 except uses a branch name without unicode so that RTD does not die.


cloud/monitoring.rst
```diff
@@ -4,41 +4,42 @@ Monitoring
...
 Amazon EBS Volume Metrics
 -------------------------
 
-* Read IOPS. The average number of read operations per second.
-   .. image:: ../images/metric-iops-read.png
-* Write IOPS. The average number of write operations per second.
-   .. image:: ../images/metric-iops-write.png
-* Average Queue Length (Count). The number of read and write operation requests waiting to be completed.
-   .. image:: ../images/metric-queue.png
+IOPS are a unit of measure representing input/output operations per second. The operations are measured in KiB, and the underlying drive technology determines the maximum amount of data that a volume type counts as a single I/O.
```
begriffs: @lfittl you mentioned that this paragraph is confusing. It comes from https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ebs-io-characteristics.html

I find IOPS kind of confusing as well, since what those "operations" can accomplish in terms of moving data around seems to depend on the physical drive.

The intention in this PR is to try and add more detail to the terms that people will see in our metrics dashboard. Is there a different description you'd give to IOPS, or perhaps leave it undefined? It seemed like one of the terms that needs a little explanation.

lfittl: I see. Maybe we can link to the AWS page so its clear where that is coming from?

I don't think its worth spending much more time on rewording it otherwise. Since the metrics are exactly the same as provided by EBS, we should use the same or similar wording.

sumedhpathak: Agreed. Can we provide links back to the original source for this and other metrics as well?

sumedhpathak:
#681 was approved by Lukas. So I see no reason to not merge this PR.

PR #688, opened 2018-06-05
Upgrading citus through the cloud UI

### LINKED ISSUE #661: Upgrading Citus version through cloud GUI

We have a guide for [on prem upgrades](https://docs.citusdata.com/en/v7.3/admin_guide/upgrading_citus.html) but the process will be easier on Cloud. Add a guide when the feature is ready. Could note that it is currently done via a support ticket.

New section in https://docs.citusdata.com/en/v7.3/cloud/manage.html

begriffs:
Fixes #661


sumedhpathak:
@byucesoy Could you take a quick look?

cloud/upgrades.rst
```diff
@@ -0,0 +1,20 @@
...
+
+.. note::
+
+  We strongly recommend :ref:`forking <cloud_forking>` a formation and trying the upgrade on the fork before upgrading the production formation. It takes more time, but provides a dry run for the real upgrade.
+
+When you click upgrade, Citus Cloud creates new servers for the coordinator and worker nodes with the requested software versions, and replays the write-ahead log from the original nodes to transfer data. This can take a fair amount of time depending on the amount of data in existing nodes. During the transfer, the formation overview tab will contain a notice:
+
+.. image:: ../images/cloud-upgrading.png
+
+When the new nodes have caught up (and during a maintenance window if one has been set), Citus Cloud automatically redirects traffic to the new nodes. During the redirection there may be a few minutes of downtime. After everything is done, the upgrade section in Settings will show:
```
sumedhpathak: Do we have information on setting the maintenance window part? When does that happen in the process? Should we have a recommendation around doing it in one?

cloud/upgrades.rst
```diff
@@ -0,0 +1,20 @@
+Upgrades
+########
+
+On a self-hosted instance of Citus, all upgrades to PostgreSQL and/or the Citus extension must be done :ref:`manually <upgrading>`.  However on Citus Cloud the work is automated. When desired, visit the Settings tab in the cloud console, and scroll down to the upgrade section:
```
sumedhpathak: Could we take out the part about 'self-hosted Citus needs manual upgrade?". Maybe instead just replacing it with spiel about Citus Cloud (Citus Cloud allows you to upgrade to the latest versions of Citus and PostgreSQL with a few clicks of the button, etc".  Also maybe something like "we recommend upgrading to the latest versions so you get all the new features and bug-fixes etc".

byucesoy:
Looks pretty good to me.

We are adding some new features to the UI such as rescheduling failover time. I will open a separate issue for those new features when they are done. This one covers all the existing details.

PR #697, opened 2018-06-08
add mark_tables_colocated in reference section

### LINKED ISSUE #398: Add mark_tables_colocated in UDF reference section
We document it here: https://docs.citusdata.com/en/v6.2/dist_tables/ddl.html?highlight=mark_tables_colocated#upgrading-from-citus-5-x, but the arguments could use some explaining.

begriffs:
Fixes #398

### DIFF
--- a/develop/api_udf.rst
+++ b/develop/api_udf.rst
@@ -110,6 +110,39 @@ reference table
 
 	SELECT upgrade_to_reference_table('nation');
 
+.. _mark_tables_colocated:
+
+mark_tables_colocated
+$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$
+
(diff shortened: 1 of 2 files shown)


develop/api_udf.rst
```diff
@@ -110,6 +110,39 @@ reference table
 
 	SELECT upgrade_to_reference_table('nation');
 
+.. _mark_tables_colocated:
+
+mark_tables_colocated
+$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$
+
+The mark_tables_colocated() function takes a distributed table (the source), and a list of others (the targets), and puts the targets into the same co-location group as the source. If the source is not yet in a group, this function creates one, and assigns the source and targets to it.
```
sumedhpathak: Can we add a few notes:
- That it will error out if the attributes don't match?
- That usually you should use colocate-with option in create_distributed_table(), rather than doing it post-facto with this UDF.

begriffs: Added those notes.

PR #698, opened 2018-06-08
When not to use citus -- match website suggestions

### LINKED ISSUE #679: Be more specific in "when citus is inappropriate"

We mention "data warehousing" as something that doesn't work well, but the term is broad enough that people may self-identify as doing data warehousing even though Citus could work well for them.

Look at what Claire wrote on the main site and use something like that.

begriffs:
Fixes #679 

Pulls in bullet points from https://www.citusdata.com/product and also adjusts existing bullet points to avoid talking about data warehousing. That terms means different things to different people.

### DIFF
--- a/get_started/what_is_citus.rst
+++ b/get_started/what_is_citus.rst
@@ -51,13 +51,17 @@ A good way to think about tools and SQL features is the following: if your workl
 When Citus is Inappropriate
 ---------------------------
 
-Workloads which require a large flow of information between worker nodes generally do not work as well. For instance:
+Some workloads require a large flow of information between worker nodes and generally do not work well on Citus. For instance:
 
-* Traditional data warehousing with long, free-form SQL
-* Many distributed transactions across multiple shards
+* Long, free-form SQL statements joining on diverse columns
+* Prevalent distributed transactions across multiple shards
 * Queries that return data-heavy ETL results rather than summaries
 
-These constraints come from the fact that Citus operates across many nodes (as compared to a single node database), giving you easy horizontal scaling as well as high availability.
+Other workloads don't need a powerful distributed database, such as:
+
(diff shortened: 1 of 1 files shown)


get_started/what_is_citus.rst
```diff
@@ -51,13 +51,17 @@ A good way to think about tools and SQL features is the following: if your workl
 When Citus is Inappropriate
 ---------------------------
 
-Workloads which require a large flow of information between worker nodes generally do not work as well. For instance:
+Some workloads require a large flow of information between worker nodes and generally do not work well on Citus. For instance:
 
-* Traditional data warehousing with long, free-form SQL
-* Many distributed transactions across multiple shards
+* Long, free-form SQL statements joining on diverse columns
```
sumedhpathak: To clarify, these are just modifications to the existing language, and the below ones are taken from the website? Is that accurate? Thanks.

begriffs: That's right. Removed the term "data warehousing" in case people mistakenly identified with it. Then copied in items from https://www.citusdata.com/product to match the website better.

sumedhpathak: Why not go with just the list we have below? 
"Queries that return data-heavy ETL results rather than summaries" sounds OK to me, but "Long, free-form SQL statements joining on diverse columns" might be possible with things like reference tables. I'd also remove "prevalent distributed transactions", since we highlight distributed transactions a good bit on the product side. 
Could we redo with the below 4 points:
 *When single-node Postgres can support your application and you do not expect to grow
* Offline analytics, without the need for real-time ingest nor real-time queries
* Analytics apps that do not need to support a large number of concurrent users
*Queries that return data-heavy ETL results rather than summaries

Issue #525, opened 2017-11-02
Section all about roll-ups


begriffs:
## Communication goals (e.g. detailed howto vs orientation)

We often get questions about best practices for doing roll-ups. Although we have a little bit about this in the existing docs, we could give more detailed advice:

* Best practices
* Common patterns
* When to do rollups (Hourly, daily)
* How do deal with delayed data
* Upsert/Insert

### Good locations for content in docs structure

Would probably fit in the Administration section by the upcoming section about useful queries.

Note that this could also be a blog post, since it's relevant to PostgreSQL in general, not just Citus.

begriffs:
There are now two places that have this kind of content:
* http://docs.citusdata.com/en/v7.5/develop/reference_dml.html#caching-aggregations-with-rollups
* http://docs.citusdata.com/en/v7.5/use_cases/realtime_analytics.html#rollups

We decided to remove the first section in DML, and improve the real-time use case guide. We want to add a "rolling up delayed data" subsection that uses ON CONFLICT. Do we want to keep average_response_time_msec and update it, or remove the average to make the example simpler?

PR #702, opened 2018-06-19
Problems with rollups into table with serial column

### LINKED ISSUE #558: INSERT SELECT doesn't support rollups with (big)serial

Running a rollup fails when the destination table contains a (big)serial column to populate. The column must be removed to proceed. In practice it hasn't been a problem to remove such columns from rollup tables, as they are often the result of an ORM or of speculative desire for a surrogate key.

A highlighted note in https://docs.citusdata.com/en/v7.1/dist_tables/dml.html#distributed-aggregations

## How does this work? (devs)

The rollups are pushed down to workers, but [nextval](https://www.postgresql.org/docs/current/static/functions-sequence.html) must be run on the coordinator, so the dml statement isn't able to populate a value into the serial column.

begriffs:
Fixes #558 

Also contains a fix in the `rollup_http_request` that I ran into while attempting to reproduce the rollup problem.


develop/reference_dml.rst
```diff
@@ -127,11 +127,25 @@ Once we create this new distributed table, we can then run :code:`INSERT INTO ..
     WHERE site_id = 5 AND
       day >= date '2016-01-01' AND day < date '2017-01-01';
 
-It's worth noting that for :code:`INSERT INTO ... SELECT` to work on distributed tables, Citus requires the source and destination table to be co-located. In summary:
+It's worth noting that for :code:`INSERT INTO ... SELECT` to work on distributed tables, Citus requires the source and destination table to be co-located.
```
sumedhpathak: This statement is probably not true anymore since we can perform insert/select via the coordinator.

develop/reference_dml.rst
```diff
@@ -127,11 +127,25 @@ Once we create this new distributed table, we can then run :code:`INSERT INTO ..
...
+.. note::
+
+  INSERT INTO…SELECT queries going from one distributed table to another cannot generate values for serial columns in the destination. Doing so requires a ``nextval()`` that is unique across the whole cluster, which is not available to a query pushed down to a single worker node. Attempting to run such a query will raise an error:
+
+  .. code-block:: text
+
+    ERROR:  0A000: INSERT ... SELECT cannot generate sequence values when selecting from a distributed table
+    LOCATION:  CoordinatorInsertSelectSupported, insert_select_planner.c:1221
+
+  We recommend removing serial columns, and instead using a UUID or manually generated integer column.
```
sumedhpathak: Can we elaborate like "removing serial columns on the destination table of an insert/select."?

begriffs:
@sumedhpathak OK, I rearranged things quite a bit, so would appreciate if you could give it another look.

develop/reference_dml.rst
```diff
@@ -3,41 +3,60 @@
...
-      2489373118,'PublicEvent','t',24509048,'{}','{"id": 24509048, "url": "https://api.github.com/repos/SabinaS/csee6868", "name": "SabinaS/csee6868"}','{"id": 2955009, "url": "https://api.github.com/users/SabinaS", "login": "SabinaS", "avatar_url": "https://avatars.githubusercontent.com/u/2955009?", "gravatar_id": ""}',NULL,'2015-01-01 00:09:13'
+    INSERT INTO github_events VALUES
+      (
+        2489373118,'PublicEvent','t',24509048,'{}','{"id": 24509048, "url": "https://api.github.com/repos/SabinaS/csee6868", "name": "SabinaS/csee6868"}','{"id": 2955009, "url": "https://api.github.com/users/SabinaS", "login": "SabinaS", "avatar_url": "https://avatars.githubusercontent.com/u/2955009?", "gravatar_id": ""}',NULL,'2015-01-01 00:09:13'
       ), (
         2489368389,'WatchEvent','t',28229924,'{"action": "started"}','{"id": 28229924, "url": "https://api.github.com/repos/inf0rmer/blanket", "name": "inf0rmer/blanket"}','{"id": 1405427, "url": "https://api.github.com/users/tategakibunko", "login": "tategakibunko", "avatar_url": "https://avatars.githubusercontent.com/u/1405427?", "gravatar_id": ""}',NULL,'2015-01-01 00:00:24'
       );
 
-Bulk Loading
+From Select
```
sumedhpathak: Maybe something like 'FROM SELECT Clause"? (And also possibly "Distributed Rollups" in parenthesis?

develop/reference_dml.rst
```diff
@@ -3,41 +3,60 @@
...
+~~~~~~~~~~~
+
+Citus also supports ``INSERT … SELECT`` statements -- which insert rows based on the results of a select query. This is a convenient way to fill tables and also allows "upserts" with the ``ON CONFLICT`` clause. One useful application is :ref:`rollups`.
+
+In Citus there are two ways that inserting from a select statement can happen. The first is if the source tables and destination table are :ref:`colocated <colocation>`, and the select/insert statements both include the distribution column. In this case Citus can push the ``INSERT … SELECT`` statement down for parallel execution on all nodes. This way supports the ``ON CONFLICT`` clause, but does not support setting default values for serial columns in the destination table.
```
sumedhpathak: > but does not support setting default values for serial columns in the destination table.
Let's take this out from this section.

sumedhpathak: Can we also add that this is useful for distributed rollups of data? Maybe link to our rollup-guide as an example?

develop/reference_dml.rst
```diff
@@ -3,41 +3,60 @@
...
+
+The second way of executing an ``INSERT … SELECT`` statement is selecting the results from worker nodes, pulling the data up to the coordinator node, and then issuing an INSERT statement from the coordinator with the data. Citus is forced to use this approach when the source and destination tables are not colocated. This method does not support ``ON CONFLICT``, but does support setting default values for serial columns in the destination table.
```
sumedhpathak: > does support setting default values for serial columns in the destination table.

This only works if we SELECT from a non-distributed table. Let's take it out given that it's somewhat subtle.

develop/reference_dml.rst
```diff
@@ -3,41 +3,60 @@
...
+
+When in doubt about which method Citus is using, use the EXPLAIN command, as described in :ref:`postgresql_tuning`.
+
+COPY Command
```
sumedhpathak: In parenthesis (Bulk Load)?

develop/reference_rollups.rst
```diff
@@ -0,0 +1,83 @@
+.. _rollups:
+
+Caching Aggregations with Rollups
```
sumedhpathak: This section doesn't quite belong in the 'reference' section for SQL. Is there anything here that isn't covered in our real-time use-case guide? Maybe ON_CONFLICT? 
I'm wondering if we nix this and just redirect as I mentioned above to the roll-up guide if we think this is covered elsewhere.

begriffs: Actually this section used to be buried in https://docs.citusdata.com/en/v7.4/develop/reference_dml.html#distributed-aggregations and had the title "Distributed Aggregations." It was deep inside "Inserting Data" so this moves it higher into its own section.

We could move it higher still, as a sibling of SQL Reference. I feel bad cutting out all that content.

begriffs:
I was experimenting with updating the real time use case guide, adding a part about rolling up delayed data with "on conflict." It's complicated because the use case guide rolls up an average, and updating the average is harder than incrementing counts.

We could remove the average from the use case example, or figure out how to do it. This may be getting out of scope for this PR which is about the serial columns. Should I delete `develop/reference_rollups.rst` as part of this PR or just leave it? Could consider prioritizing the use case guide update later for #525

develop/reference_dml.rst
```diff
@@ -127,12 +145,6 @@ Once we create this new distributed table, we can then run :code:`INSERT INTO ..
     WHERE site_id = 5 AND
       day >= date '2016-01-01' AND day < date '2017-01-01';
 
-It's worth noting that for :code:`INSERT INTO ... SELECT` to work on distributed tables, Citus requires the source and destination table to be co-located. In summary:
```
sumedhpathak: Why delete these?

sumedhpathak: Maybe reword this to for "on-conflict clause" to work.. we require colocation etc.

sumedhpathak:
If adding an UPSERT example to the use-case guide isn't as simple, let's proceed with what you suggest (i.e. leaving the existing section as-is).

After adding back the "colocation required" warning sentences, you are good to 🚢

develop/reference_dml.rst
begriffs: Moved the colocation required warning a little later where it reads better.

PR #703, opened 2018-06-20
Bump Citus version to 7.4.1

velioglu:
Bump Citus version

### DIFF
--- a/installation/multi_machine_aws.rst
+++ b/installation/multi_machine_aws.rst
@@ -17,7 +17,7 @@ Below, we explain in detail the steps required to setup a multi-node Citus clust
 
 .. raw:: html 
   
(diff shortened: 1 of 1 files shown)


sumedhpathak:
@begriffs Could you verify this PR (that the CF template correctly deploys, has the right permissions, and has the right version)?

Feel free to merge after that.

begriffs:
Deployed and verified:

```
# select citus_version();
                                            citus_version
-----------------------------------------------------------------------------------------------------
 Citus 7.4.1 on x86_64-pc-linux-gnu, compiled by gcc (GCC) 4.4.7 20120313 (Red Hat 4.4.7-18), 64-bit
```

Also verified I could curl https://citus-deployment.s3.amazonaws.com/aws/citus7/cloudformation/citus-7.4.1.json and thus that the file is accessible to the general public.

PR #705, opened 2018-06-21
Add citus_create_restore_point to udfs

begriffs:
Documents https://github.com/citusdata/citus/issues/1451

Later on we could make a more general section about distributed backups.


sumedhpathak:
Approved, but can we use content Marco wrote up here: https://github.com/citusdata/citus/pull/1452, including the limitations? After that, feel free to merge.

PR #706, opened 2018-06-25
Updated Restore point text, and added note for EE edition

sumedhpathak:
Added a bit more context from Marco's note, and also have added a note that support of it is part of EE.


begriffs:
Closing, since these changes got committed another way.

PR #708, opened 2018-06-28
Note misleading error for table non-owners

begriffs:
I added this as a note to the error message since it seems like just one of many cases. However if you think in reality it's the star of the show then I can work it into the regular description and resolution parts.

### DIFF
--- a/reference/common_errors.rst
+++ b/reference/common_errors.rst
@@ -132,6 +132,10 @@ Cannot establish a new connection for placement *n*, since DML has been executed
 
 This is a current limitation. In a single transaction Citus does not support running insert/update statements with the :ref:`router_executor` that reference multiple shards, followed by a read query that consults both of the shards.
 
+.. note::
+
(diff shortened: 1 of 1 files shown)


reference/common_errors.rst
```diff
@@ -132,6 +132,17 @@ Cannot establish a new connection for placement *n*, since DML has been executed
 
 This is a current limitation. In a single transaction Citus does not support running insert/update statements with the :ref:`router_executor` that reference multiple shards, followed by a read query that consults both of the shards.
 
+.. note::
+
+  A similar error also occurs (misleadingly) when the :ref:`create_distributed_table` function is executed on a table by a role other than the table's owner. See this `github discussion <https://github.com/citusdata/citus/issues/2094>`_ for details. To resolve this particular problem, identify the table's owner, switch roles, and try again.
```
begriffs: I say "similar" because that error mentions DDL rather than DML.

begriffs:
@sumedhpathak I'm putting a 5-day ⏰  on this one since it's pretty small.

reference/common_errors.rst
```diff
@@ -132,6 +132,17 @@ Cannot establish a new connection for placement *n*, since DML has been executed
...
+
+  .. code-block:: sql
+
+    -- find the role
+    SELECT tablename, tableowner FROM pg_tables;
+    -- switch into it
+    SET ROLE table_owner_name;
```
sumedhpathak: Did you test that this works?

PR #716, opened 2018-07-03
Unique index on non-distribution column

### LINKED ISSUE #363: Document why UNIQUE indexes on non-partition columns are unsupported
Relevant support thread:
https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/citus-users/O3eDLxcsEfw/141NJea7BAAJ

People see an error but don't know why.

Document why its not supported. What are the technical reasons why it would not work well.

DDL/FAQ or both?

### LINKED ISSUE #685: Add alternatives/workarounds for Unique Index on non-partition column.
See https://www.reddit.com/r/PostgreSQL/comments/8l04sy/have_any_of_you_migrated_from_postgresql_to_a/dzeagcg/

I had a few problems in Docs:
- Searching for "unique index" doesn't actually bring up anything meaningful.
https://docs.citusdata.com/en/v7.4/search.html?q=unique+index&check_keywords=yes&area=default
We should definitely add some content so that this search would be useful.

- We don't AFAICS have workarounds or options for unique index on non-partition column. Maybe this is in a separate section? @begriffs Do you know?

begriffs:
Fixes #363 
Fixes #685 

Adds a section to our common error messages about it, with two workarounds. Also adds a note to the DDL section.

### DIFF
--- a/reference/common_errors.rst
+++ b/reference/common_errors.rst
@@ -257,6 +257,8 @@ Resolution
 
 Add the table distribution column to both the select and insert statements, as well as the statement GROUP BY if applicable. For more info as well as a workaround, see :ref:`upsert_into_select`.
 
+.. _non_distribution_uniqueness:
+
 Cannot create uniqueness constraint
 -----------------------------------
 
--- a/reference/common_errors.rst
+++ b/reference/common_errors.rst
@@ -256,3 +256,26 @@ Resolution
 ~~~~~~~~~~
 
 Add the table distribution column to both the select and insert statements, as well as the statement GROUP BY if applicable. For more info as well as a workaround, see :ref:`upsert_into_select`.
+
+Cannot create uniqueness constraint
+-----------------------------------
+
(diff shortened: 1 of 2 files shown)


reference/common_errors.rst
```diff
@@ -243,3 +243,28 @@ Resolution
...
+  LOCATION:  ErrorIfUnsupportedConstraint, multi_utility.c:2505
+
+Enforcing uniqueness on a non-distribution column would require Citus to check every shard on every INSERT to validate, which defeats the goal of scalability.
+
+Resolution
+~~~~~~~~~~
+
+There are two ways to enforce uniqueness on a non-distribution column:
+
+1. Create a composite unique index or primary key that includes the desired column (*C*), but also includes the distribution column (*D*). This is not quite as strong a condition as uniqueness on *C* alone, but will ensure that the values of *C* are unique for each value of *D*. For instance if distributing by ``company_id`` in a multi-tenant system, this approach would make *C* unique within each company.
```
begriffs: Maybe I should add an example here?

PR #717, opened 2018-07-09
Update claims about altering dist col

### LINKED ISSUE #668: Distribution column can be renamed, documentation bug
Under the chapter [Adding/Modifying Columns](https://docs.citusdata.com/en/latest/develop/reference_ddl.html#adding-modifying-columns) it incorrectly states the rename of the distribution column does not work.

This has been implemented in https://github.com/citusdata/citus/pull/1312 which seems to be first included in v6.2.0: https://github.com/citusdata/citus/commit/5272c2c44b98fff531ee2a820d375abedf2da6fb

I successfully renamed the distribution column on a hash distributed column in the version that is to be released as 7.4

begriffs:
Fixes #668


begriffs:
@sumedhpathak I'm adding a 5⏰ to this one.

PR #732, opened 2018-07-27
Trigger workarounds

### LINKED ISSUE #722: Document that Triggers are Unsupported
For example:
```SQL
drop table if exists insert_target;
drop table if exists audit_table;

CREATE TABLE insert_target (
    value           text
);

CREATE TABLE audit_table(
    stamp             timestamp ,
    value            text       
);

CREATE OR REPLACE FUNCTION process_audit() RETURNS TRIGGER AS $emp_audit$
    BEGIN
	insert into audit_table(stamp,value) VALUES (now(),'value');
        RETURN NEW;
    END;
$emp_audit$ LANGUAGE plpgsql;

CREATE TRIGGER emp_audit
AFTER INSERT OR UPDATE OR DELETE ON insert_target
    EXECUTE PROCEDURE process_audit();

select create_reference_table('insert_target');
select create_reference_table('audit_table');

explain analyze insert into insert_target (value) VALUES ('inserted value');

select * from insert_target;
select * from audit_table;

```

* We need to document that Triggers on Distributed Tables are unsupported
* Ideally we should document a workaround or alternative as well. Will need to check with @samay-sharma or @saicitus on workaround.

begriffs:
Fixes #722


develop/reference_workarounds.rst
```diff
@@ -128,3 +128,201 @@ There is another trick though. We can pull the relevant information to the coord
...
+
+Include the shard ids (written below as "xxxxxx" and "yyyyyy") in custom queries:
+
+.. code-block:: postgresql
+
+  SELECT run_command_on_workers($cmd$
+    CREATE OR REPLACE FUNCTION process_audit() RETURNS TRIGGER AS $$
+      BEGIN
+        INSERT INTO audit_table_xxxxxx(stamp,value) VALUES (now(),'value');
+        RETURN NEW;
```
marcocitus: Note that this is unsafe in many cases because it by-passes the write locks on the reference table. Therefore, concurrent writes to audit_table may be applied in a different order on different replicas which would cause replicas to diverge.

marcocitus: It's also unsafe for replication factor > 1

sumedhpathak: Is it because of the now()? @begriffs In a customer example, we had a timestamp on the `insert_target` table, and used that value in the `audit_table` using `NEW.`. That one gets evaluated on the master, so the value remains the same across all placements.

begriffs: @marcocitus is there a better workaround, or is there no safe way to do triggers between reference tables?

develop/reference_workarounds.rst
```diff
@@ -128,3 +128,201 @@ There is another trick though. We can pull the relevant information to the coord
...
+      END;
+    $$ LANGUAGE plpgsql;
+  $cmd$);
+
+  SELECT run_command_on_workers($cmd$
+    CREATE TRIGGER emp_audit
+    AFTER INSERT OR UPDATE OR DELETE ON insert_target_yyyyyy
```
sumedhpathak: Do we need the YYYY here for insert target?

develop/reference_workarounds.rst
marcocitus: It's because the order in which concurrent update/delete/upsert commands are applied to replica's is not guaranteed. Technically, an insert-only example is safe because insertion order doesn't really matter, but small deviations can cause it not to be. There is no workaround right now. In MX you will be able to update another reference table from a trigger on a single shard placement. That has some other downsides (it creates hierarchies of cached connections), but is safe.

develop/reference_workarounds.rst
begriffs: Yeah, it's being run on the worker so it can't see plain old insert_target. (Tried it to double check and got an error)

begriffs:
* I moved the trigger stuff under `/extra` as Samay suggested
* Still using `now()` in the `process_audit()` trigger because `NEW` won't have any timestamp column in it (`insert_target` doesn't have a timestamp field that someone would be including, so nothing to carry over in NEW)
* Do we keep the ref<->ref workaround at all? If we want to suggest using it in an insert-only workload then I'll need to switch up the audit table example since it's only there to keep track of a table that gets updated

begriffs:
Meh, on second thought the ref table stuff was kind of sloppy. I removed it and just said, "This is not yet possible."

begriffs:
OK, the ref-ref trigger no longer uses `now()`. It is kind of a weak example, since it's similar to the trigger on a single distributed table but not as useful. Hopefully it gives people what they need to set up ref-ref in more useful ways in their own projects.

extra/triggers.rst
```diff
@@ -0,0 +1,204 @@
...
+  ├─────┼─────┤
+  │   1 │  84 │
+  │   2 │ 202 │
+  └─────┴─────┘
+
+**Trigger between reference tables.**
+
+.. note::
+
+  **This workaround is only safe in limited situations.** When using such a trigger to insert into a reference table, make sure that no concurrent updates happen on the destination table. The order in which concurrent update/delete/insert commands are applied to replicas is not guaranteed, and replicas of the reference table can get out of sync with one another.
```
sumedhpathak: To reinforce, I'd say that any DML to the destination then should be via the trigger only.

extra/triggers.rst
```diff
@@ -0,0 +1,204 @@
...
+        INSERT INTO audit_table_xxxxxx(author,value)
+          VALUES (current_user,NEW.value);
+        RETURN NEW;
+      END;
+    $$ LANGUAGE plpgsql;
+  $cmd$);
+
+  SELECT run_command_on_workers($cmd$
+    CREATE TRIGGER emp_audit
+    AFTER INSERT OR UPDATE ON insert_target_yyyyyy
```
sumedhpathak: I'm still not sure why we need shard-id's for both reference tables? We do need it for the destination table, but the source table shouldn't be needed. Can we do run_command_on_shards (or placements) instead?

begriffs: The tables `insert_target` and `audit_table` don't exist on the workers. Referencing either of those tables within `run_command_on_workers()` will cause a "relation does not exist" error.

For instance:

```sql
-- produces a value
select count(*) from insert_target;

-- no good
select run_command_on_workers('select count(*) from insert_target');
┌────────────────────────────────────────────────────────────────────────┐
│                         run_command_on_workers                         │
├────────────────────────────────────────────────────────────────────────┤
│ (localhost,5433,f,"ERROR:  relation ""insert_target"" does not exist") │
│ (localhost,5434,f,"ERROR:  relation ""insert_target"" does not exist") │
└────────────────────────────────────────────────────────────────────────┘

-- but this works
select run_command_on_workers('select count(*) from insert_target_102425');
```

sumedhpathak: Yes, but run_command_on_shards (or placements) should resolve the shard_id for the table. Why can't we use the other API's? Do we have to use the generic run_command_on_workers?

begriffs: I see what you're saying. I'll try to simplify it with one of the other UDFs.

begriffs:
@sumedhpathak good idea about using a different function to fill in the names. I updated the example and tested it.

PR #734, opened 2018-07-31
Sometimes colocating with "none" is best

### LINKED ISSUE #733: Warn of the hazards of implicit colocation

1) someone creates a table on two nodes, 2) adds a node 3) distributes another table which only goes on two nodes because it is implicitly colocated

* http://docs.citusdata.com/en/v7.5/develop/api_udf.html#create-distributed-table
* http://docs.citusdata.com/en/v7.5/develop/reference_ddl.html#co-locating-tables

### How does this work? (devs)

Use `colocate_with=>none`

begriffs:
Fixes #733

### DIFF
--- a/develop/reference_ddl.rst
+++ b/develop/reference_ddl.rst
@@ -138,7 +138,9 @@ If you would prefer a table to be in its own co-location group, specify :code:`'
 
   SELECT create_distributed_table('A', 'foo', colocate_with => 'none');
 
-To co-locate a number of tables, distribute one and then put the others into its co-location group. For example:
+Using co-location group ``'none'`` can potentially take advantage of more worker nodes than existing groups. A table distributed into an existing co-location group will be placed on exactly the same nodes as other tables in the group, which doesn't include nodes added after the group was created.
+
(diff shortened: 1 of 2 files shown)


develop/reference_ddl.rst
```diff
@@ -138,7 +138,9 @@ If you would prefer a table to be in its own co-location group, specify :code:`'
 
   SELECT create_distributed_table('A', 'foo', colocate_with => 'none');
 
-To co-locate a number of tables, distribute one and then put the others into its co-location group. For example:
+Using co-location group ``'none'`` can potentially take advantage of more worker nodes than existing groups. A table distributed into an existing co-location group will be placed on exactly the same nodes as other tables in the group, which doesn't include nodes added after the group was created.
```
sumedhpathak: I wouldn't frame it like that, given we can rebalance shards to take advantage of other nodes. I'd just say, if a table isn't related to others, just use 'none'. If not we might implicitly colocate with other tables, and can have adverse impacts. E.g. when rebalancing, we might move more than just that one shard. Or for shard placement, we'll be limited to locations of the other colocated table.

PR #735, opened 2018-08-01
Common error message about stable functions

begriffs:


begriffs:
@marcocitus can I get your feedback about this? Want to be sure I'm saying the right things.

PR #739, opened 2018-08-06
Installing enterprise edition

begriffs:


extra/enterprise.rst
```diff
@@ -0,0 +1,87 @@
...
+5. Continue by following the standard :ref:`multi-machine debian <post_enterprise_deb>` installation steps, **starting at step 3.**
+
+.. _enterprise_rhel:
+
+Fedora, CentOS, or Red Hat
+--------------------------
+
+1. Contact us to obtain a Citus Enterprise repository token. This token grants download access to the packages.
+
+2. Create a package tarball:
```
sumedhpathak: @samay-sharma This assumes we always create a tarball. Do we want to have separate instructions for "if you don't have internet access on your machines"?

extra/enterprise.rst
```diff
@@ -0,0 +1,87 @@
...
+===================================
+
+.. _enterprise_debian:
+
+Ubuntu or Debian
+----------------
+
+1. Contact us to obtain a Citus Enterprise repository token. This token grants download access to the packages.
+
+2. Create a package tarball:
```
sumedhpathak: Hey Joe, this is missing a few items we discussed:
- Two methods: If you have internet, or if you don't. This only captures if you don't have internet.
- Upgrading from citus community. We should document steps needed for that.

extra/enterprise.rst
```diff
@@ -0,0 +1,87 @@
...
+
+  .. code-block:: bash
+
+    # replace XYZ with your token
+    curl https://install.citusdata.com/enterprise/deb.sh | \
+      sudo CITUS_REPO_TOKEN=XYZ bash
+
+    sudo apt-get clean
+    sudo apt-get install -y --download-only postgresql-10-citus-rebalancer-7.5
```
sumedhpathak: Does this pull in Enterprise / PG etc? Should we comment that?

PR #754, opened 2018-08-14
Add entry to pg_dist_local_group after upgrade

### LINKED ISSUE #738: pg_dist_local_group on workers needs to be preserved during pg_upgrade
We're messing a step in the [upgrade instructions](http://citus-docs.readthedocs.io/en/v7.5/admin_guide/upgrading_citus.html) to preserve `pg_dist_local_group`, which causes an issue:
https://github.com/citusdata/citus/issues/2329

We should also consider fixing this in the extension.

begriffs:
Fixes #738

### DIFF
--- a/admin_guide/upgrading_citus.rst
+++ b/admin_guide/upgrading_citus.rst
@@ -112,21 +112,21 @@ On Every Node (Coordinator and workers)
 
   .. code-block:: postgres
 
-    CREATE TABLE public.pg_dist_partition AS
+    CREATE TABLE        public.pg_dist_partition AS
       SELECT * FROM pg_catalog.pg_dist_partition;
-    CREATE TABLE public.pg_dist_shard AS
+    CREATE TABLE        public.pg_dist_shard AS
       SELECT * FROM pg_catalog.pg_dist_shard;
-    CREATE TABLE public.pg_dist_placement AS
+    CREATE TABLE        public.pg_dist_placement AS
       SELECT * FROM pg_catalog.pg_dist_placement;
-    CREATE TABLE public.pg_dist_node_metadata AS
+    CREATE TABLE        public.pg_dist_node_metadata AS
       SELECT * FROM pg_catalog.pg_dist_node_metadata;
-    CREATE TABLE public.pg_dist_node AS
+    CREATE TABLE        public.pg_dist_node AS
       SELECT * FROM pg_catalog.pg_dist_node;
-    CREATE TABLE public.pg_dist_local_group AS
+    CREATE TABLE        public.pg_dist_local_group AS
       SELECT * FROM pg_catalog.pg_dist_local_group;
-    CREATE TABLE public.pg_dist_transaction AS
+    CREATE TABLE        public.pg_dist_transaction AS
       SELECT * FROM pg_catalog.pg_dist_transaction;
-    CREATE TABLE public.pg_dist_colocation AS
+    CREATE TABLE        public.pg_dist_colocation AS
       SELECT * FROM pg_catalog.pg_dist_colocation;
 
 2. Configure the new database instance to use Citus.
--- a/admin_guide/upgrading_citus.rst
+++ b/admin_guide/upgrading_citus.rst
@@ -260,3 +260,8 @@ On Worker Nodes
                                 -d $OLD_PG_DATA -D $NEW_PG_DATA
 
 4. Start the new server.
+5. Add entry to ``pg_dist_local_group``:
+
+  .. code-block:: postgres
+
+    INSERT INTO pg_dist_local_group VALUES (0);
--- a/admin_guide/upgrading_citus.rst
+++ b/admin_guide/upgrading_citus.rst
@@ -105,20 +105,29 @@ New data directory after upgrade
 New PostgreSQL installation path
   :code:`export NEW_PG_PATH=/usr/pgsql-10`
 
-On the Coordinator Node
------------------------
+On Every Node (Coordinator and workers)
+---------------------------------------
 
 1. Back up Citus metadata in the old server.
 
   .. code-block:: postgres
 
-    CREATE TABLE public.pg_dist_partition AS SELECT * FROM pg_catalog.pg_dist_partition;
-    CREATE TABLE public.pg_dist_shard AS SELECT * FROM pg_catalog.pg_dist_shard;
-    CREATE TABLE public.pg_dist_placement AS SELECT * FROM pg_catalog.pg_dist_placement;
-    CREATE TABLE public.pg_dist_node AS SELECT * FROM pg_catalog.pg_dist_node;
-    CREATE TABLE public.pg_dist_local_group AS SELECT * FROM pg_catalog.pg_dist_local_group;
-    CREATE TABLE public.pg_dist_transaction AS SELECT * FROM pg_catalog.pg_dist_transaction;
-    CREATE TABLE public.pg_dist_colocation AS SELECT * FROM pg_catalog.pg_dist_colocation;
+    CREATE TABLE public.pg_dist_partition AS
+      SELECT * FROM pg_catalog.pg_dist_partition;
+    CREATE TABLE public.pg_dist_shard AS
+      SELECT * FROM pg_catalog.pg_dist_shard;
+    CREATE TABLE public.pg_dist_placement AS
+      SELECT * FROM pg_catalog.pg_dist_placement;
+    CREATE TABLE public.pg_dist_node_metadata AS
+      SELECT * FROM pg_catalog.pg_dist_node_metadata;
+    CREATE TABLE public.pg_dist_node AS
+      SELECT * FROM pg_catalog.pg_dist_node;
+    CREATE TABLE public.pg_dist_local_group AS
+      SELECT * FROM pg_catalog.pg_dist_local_group;
+    CREATE TABLE public.pg_dist_transaction AS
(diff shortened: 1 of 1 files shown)


begriffs:
@sumedhpathak so I just tried to reproduce the upgrade problem where `pg_dist_local_group` is supposedly empty, but I can't do it. I tried creating a [single machine cluster](http://docs.citusdata.com/en/v7.5/installation/single_machine_debian.html) except using `postgresql-9.6-citus-7.5`. Then I distributed the table from the real-time use-case guide and added some data, just so it wasn't an empty db, and then followed the [9.6 -> 10](http://docs.citusdata.com/en/v7.5/admin_guide/upgrading_citus.html#upgrading-postgresql-version-from-9-6-to-10-0) upgrade steps in our docs, adjusting the paths as appropriate.

After running this step in the original instructions
```sql
CREATE TABLE public.pg_dist_local_group AS SELECT * FROM pg_catalog.pg_dist_local_group;
```
the `pg_dist_local_group` table has a row:
```sql
TABLE pg_dist_local_group;
/*
 groupid
---------
       0
*/
```

I believe that the extra step I added in this PR is indeed redundant. I had added it on faith based on a support email from @marcocitus that solved a customer issue. There must have been something else going on for the customer since it appears that our existing upgrade steps should be working.

We can probably close this PR.

marcocitus:
I believe the issue is that the `pg_dist_local_group` entry is needed on the workers, while all the other metadata is only needed on the coordinator.

begriffs:
Thanks for the clarification, I'm going through the whole upgrade process now to try to reproduce the problem on the workers.

begriffs:
@sumedhpathak and @marcocitus - OK, I reproduced the error in an upgrade, and confirmed that adding a row to `pg_dist_local_group` **on the workers** fixes it. (In fact, reviewing my original changes in this PR, it said to add the row on the workers, not the coordinator. I lost track of that when looking at the PR after Sumedh's question.)

One note, although this fixes the error about pg_dist_local_group, there is still a warning at the end that says, "WARNING:  could not find any entries in pg_dist_metadata." Maybe there's another step to fix that as well?

___
Log from reproducing the problem and fixing it:

```sh
curl https://install.citusdata.com/community/deb.sh | sudo bash
sudo apt-get install postgresql-9.6-citus-7.5 postgresql-10-citus-7.5

sudo su - postgres

export OLD_PG_PATH=/usr/lib/postgresql/9.6
export NEW_PG_PATH=/usr/lib/postgresql/10

mkdir -p c96/coordinator
mkdir c09/worker{1,2}

mkdir -p c10/coordinator
mkdir c10/worker{1,2}

$OLD_PG_PATH/bin/initdb -D c96/coordinator
$OLD_PG_PATH/bin/initdb -D c96/worker1
$OLD_PG_PATH/bin/initdb -D c96/worker2

$NEW_PG_PATH/bin/initdb -D c10/coordinator
$NEW_PG_PATH/bin/initdb -D c10/worker1
$NEW_PG_PATH/bin/initdb -D c10/worker2

echo "shared_preload_libraries = 'citus'" >> c96/coordinator/postgresql.conf
echo "shared_preload_libraries = 'citus'" >> c96/worker1/postgresql.conf
echo "shared_preload_libraries = 'citus'" >> c96/worker2/postgresql.conf
echo "shared_preload_libraries = 'citus'" >> c10/coordinator/postgresql.conf
echo "shared_preload_libraries = 'citus'" >> c10/worker1/postgresql.conf
echo "shared_preload_libraries = 'citus'" >> c10/worker2/postgresql.conf

$OLD_PG_PATH/bin/pg_ctl -D c96/coordinator -o "-p 9700" -l coordinator_logfile start
$OLD_PG_PATH/bin/pg_ctl -D c96/worker1 -o "-p 9701" -l worker1_logfile start
$OLD_PG_PATH/bin/pg_ctl -D c96/worker2 -o "-p 9702" -l worker2_logfile start

psql -p 9700 -c "CREATE EXTENSION citus;"
psql -p 9701 -c "CREATE EXTENSION citus;"
psql -p 9702 -c "CREATE EXTENSION citus;"

psql -p 9700 -c "SELECT * from master_add_node('localhost', 9701);"
psql -p 9700 -c "SELECT * from master_add_node('localhost', 9702);"

####
# create and fill http_request table from RT use case guide
# ... <commands> ...

$OLD_PG_PATH/bin/pg_ctl -D c96/coordinator stop
$OLD_PG_PATH/bin/pg_ctl -D c96/worker1 stop
$OLD_PG_PATH/bin/pg_ctl -D c96/worker2 stop

$NEW_PG_PATH/bin/pg_upgrade -b $OLD_PG_PATH/bin/ -B $NEW_PG_PATH/bin/ \
                            -d c96/coordinator -D c10/coordinator --check

## looks good

$NEW_PG_PATH/bin/pg_upgrade -b $OLD_PG_PATH/bin/ -B $NEW_PG_PATH/bin/ \
                            -d c96/coordinator -D c10/coordinator

$NEW_PG_PATH/bin/pg_ctl -D c10/coordinator -o "-p 9800" -l coordinator_logfile start

psql -p 9800
# restore metadata tables, drop backups, restart seqs, restore triggers, set deps
# ... <commands> ...

$NEW_PG_PATH/bin/pg_upgrade -b $OLD_PG_PATH/bin/ -B $NEW_PG_PATH/bin/ \
                            -d c96/worker1 -D c10/worker1 --check
$NEW_PG_PATH/bin/pg_upgrade -b $OLD_PG_PATH/bin/ -B $NEW_PG_PATH/bin/ \
                            -d c96/worker2 -D c10/worker2 --check

## looks good

$NEW_PG_PATH/bin/pg_upgrade -b $OLD_PG_PATH/bin/ -B $NEW_PG_PATH/bin/ \
                            -d c96/worker1 -D c10/worker1
$NEW_PG_PATH/bin/pg_upgrade -b $OLD_PG_PATH/bin/ -B $NEW_PG_PATH/bin/ \
                            -d c96/worker2 -D c10/worker2

# start new workers with old ports

$NEW_PG_PATH/bin/pg_ctl -D c10/worker1 -o "-p 9701" -l worker1_logfile start
$NEW_PG_PATH/bin/pg_ctl -D c10/worker2 -o "-p 9702" -l worker2_logfile start

psql -p 9800 -c 'table http_request;'

tail worker1_logfile
```
This reveals the error:
```
2018-08-21 21:45:22.557 UTC [8481] ERROR:  could not find any entries in pg_dist_local_group
2018-08-21 21:45:22.557 UTC [8481] STATEMENT:  SELECT * FROM dump_local_wait_edges()
```
Now fix it and check:
```sh
psql -p 9701 -c 'INSERT INTO pg_dist_local_group VALUES (0);'
psql -p 9702 -c 'INSERT INTO pg_dist_local_group VALUES (0);'
echo "" > worker1_logfile
psql -p 9800 -c 'table http_request;'
tail worker1_logfile
```
Shows
```
2018-08-21 21:49:32.605 UTC [8994] WARNING:  could not find any entries in pg_dist_metadata
```

admin_guide/upgrading_citus.rst
```diff
@@ -260,3 +260,8 @@ On Worker Nodes
                                 -d $OLD_PG_DATA -D $NEW_PG_DATA
 
 4. Start the new server.
+5. Add entry to ``pg_dist_local_group``:
```
sumedhpathak: I think the right approach should be to copy/save the metadata values before the upgrade and copy them back after the upgrade (similar to the master node). This fix is a temporary one which works since we aren't yet using pg_dist_local_group. 
@marcocitus What is the interaction with MX? Should we backup all metadata like we do with the coordinator node?

begriffs:
@sumedhpathak OK, gave it another test with the new instructions (on Ubuntu single-node cluster), and it works for me.

The instructions have the user back up all metadata tables, even though some are unnecessary on the workers. It can't hurt, and is more compatible with MX.

admin_guide/upgrading_citus.rst
```diff
@@ -188,12 +207,12 @@ On the Coordinator Node
...
+
     SELECT setval('pg_catalog.pg_dist_groupid_seq', (SELECT MAX(groupid)+1 AS max_group_id FROM pg_dist_node), false);
 
     SELECT setval('pg_catalog.pg_dist_node_nodeid_seq', (SELECT MAX(nodeid)+1 AS max_node_id FROM pg_dist_node), false);
 
-    SELECT setval('pg_catalog.pg_dist_placement_placementid_seq', (SELECT MAX(placementid)+1 AS max_placement_id FROM pg_dist_placement), false);
-
     SELECT setval('pg_catalog.pg_dist_colocationid_seq', (SELECT MAX(colocationid)+1 AS max_colocation_id FROM pg_dist_colocation), false);
 
 10. Register triggers.
```
sumedhpathak: I was a bit skeptical of steps 10 and 11 for upgrade on the workers. @onderkalaci Any issues if we register truncate triggers and set the dependency on citus extension on the workers for MX? Dependency seems fine, but do we do something special for table truncate?

onderkalaci: Oh, sorry that I missed this comment.

Yes, we need to create the `TRUNCATE` triggers on the workers as well for MX (because we support `TRUNCATE` from the worker nodes as well).

So, it makes sense that this step is under `On Every Node (Coordinator and workers)`.

>  but do we do something special for table truncate?

I don't see anything special for truncate other than creating the triggers.

PR #756, opened 2018-08-14
Common error about could not connect

### LINKED ISSUE #745: Common error: parallel copy running out of connections

Doing a parallel copy uses a lot of connections, and they are in limited supply.

Reproduce the problem myself, and get the error message.

The solution is to reduce the shard count or use fewer copies in parallel

Common errors section.

begriffs:
Fixes #745

### DIFF
--- a/reference/common_errors.rst
+++ b/reference/common_errors.rst
@@ -174,7 +174,7 @@ This allows reusing sockets in TIME_WAIT state for new connections when it is sa
 Could not connect to any active placements
 ------------------------------------------
 
-When all available worker connection slots are in use, further connections will fail. This happens most often when copying data into Citus in parallel:
+When all available worker connection slots are in use, further connections will fail.
 
 ::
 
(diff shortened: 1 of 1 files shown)


reference/common_errors.rst
```diff
@@ -171,6 +171,23 @@ Configure the operating system to re-use TCP sockets. Execute this on the shell
 
 This allows reusing sockets in TIME_WAIT state for new connections when it is safe from a protocol viewpoint. Default value is 0 (disabled).
 
+Could not connect to any active placements
+------------------------------------------
+
+When all available worker connection slots are in use, further connections will fail. This happens most often when copying data into Citus in parallel:
```
sumedhpathak: Could we link to http://docs.citusdata.com/en/v7.5/performance/performance_tuning.html#bulk-copy-250k-2m-s? It would also be good to add some color on why this will happen. i.e., COPY always opens up a connection per shard. So if you run 10 concurrent copies, that will result in X number of connections. So recommendation would be to reduce the number of parallel copies.

PR #792, opened 2018-10-18
Protecting against long-running queries

### LINKED ISSUE #787: How to protect against long-running queries

Long running queries can hold locks, queue up WAL, or just consume a lot of system resources. Users would like an easy way to prevent queries from running on too long.

Show people that the `statement_timeout` GUC automates aborting long queries.

### Corner cases, gotchas

Changing this value on the coordinator doesn't automatically propagate to Citus workers.
Note to self: test what happens if the value is set only on the coordinator.

* https://www.citusdata.com/blog/2017/04/28/postgres-tips-for-rails/
* https://www.postgresql.org/docs/10/static/runtime-config-client.html#GUC-STATEMENT-TIMEOUT

begriffs:
Fixes #787

### DIFF
--- a/admin_guide/cluster_management.rst
+++ b/admin_guide/cluster_management.rst
@@ -418,10 +418,10 @@ Long running queries can hold locks, queue up WAL, or just consume a lot of syst
 
    -- limit queries to five minutes
    ALTER DATABASE citus
-     SET statement_timeout TO 1000;
+     SET statement_timeout TO 300000;
    SELECT run_command_on_workers($cmd$
      ALTER DATABASE citus
-       SET statement_timeout TO 1000;
+       SET statement_timeout TO 300000;
    $cmd$);
 
 The timeout is specified in milliseconds.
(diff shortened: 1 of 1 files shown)


admin_guide/cluster_management.rst
```diff
@@ -406,6 +406,30 @@ Finally, by joining ``citus_stat_statements`` with ``pg_stat_statements`` we can
...
+
+Limiting Long-Running Queries
+-----------------------------
+
+Long running queries can hold locks, queue up WAL, or just consume a lot of system resources, so in a production environment it's good to prevent them from running too long. You can set the `statement_timeout <https://www.postgresql.org/docs/10/static/runtime-config-client.html#GUC-STATEMENT-TIMEOUT>`_ parameter on the coordinator and workers to cancel queries that run too long.
+
+.. code-block:: postgres
+
+   -- limit queries to five minutes
+   ALTER DATABASE citus
```
sumedhpathak: Don't we cancel the worker queries if we set this on the coordinator? Did you observe any different behavior?

begriffs: No, I observed the queries on the workers to keep running (with citus 7.5.0 on Cloud at least)

sumedhpathak:
🚢 . Separately, @marcocitus Wouldn't statement_timeout on the coordinator cause cancellation of the worker queries?

Issue #790, opened 2018-10-17
Update "How Far Can Citus Scale" section -- Reference to Kixer


ozgune:
Could we update this section and replace Kixer with a different customer (who agreed to a case study with us)?

ozgune:
I also realize that some of the references we give about scale is dated in this section. For example, Heap now stores ~900 TB of data. Do we want to update this section too?

sumedhpathak:
@begriffs Can we prioritize this? Also, Mixrank has 1.6PB of data in their cluster. We could also reduce the number of stories we list here, since we also point to our customer stories section on our site.

begriffs:
Yep, I moved it right under the 8.0 release checklist issue in waffle.

PR #796, opened 2018-10-22
Simplify "What is citus" section

### LINKED ISSUE #791: Think about flow in the "What is Citus" page?
I'm reading our documentation's first page, and I'm a bit confused.

https://docs.citusdata.com/en/v7.5/get_started/what_is_citus.html

In particular, I'm confused about two things. First, this section starts with "What is Citus?"

>> Fast-growing multi-tenant apps want to add new customers, deliver great performance, and not have to worry about database infrastructure. Data analysts want sub-second response times for customer-facing analytics dashboards, even with real-time ingestion, very large data sets, complex queries, and lots of concurrent users.

This paragraph doesn't talk about Citus.

Also, I find that our definitions for "a multi-tenant app" and "customer-facing analytics dashboards" require on-boarding. Even people who're familiar with distributed databases need time to digest and align with this terminology.

Then, before the reader understands what Citus is, we jump to benefits specific to use-cases.

>> Citus allows these and other applications to enjoy the power and familiarity of a traditional relational database, but with the capability for massive scale. Applications connect to and use a Citus-enabled PostgreSQL database just like they would a traditional PostgreSQL database. Citus uses the same SQL commands that developers and frameworks already know.

>> Multi-Tenant Advantages | Real-Time Advantages

Overall, this section looks like we're trying to answer the question "Why Citus?" instead of "What is Citus?" **If** that's the goal, could we instead create a separate question? We could then directly link to our product and use-cases sections? (cc @clairegiordano)

I'm looking to share our docs with a technical audience, that's the reason I'm asking.

Second, we also define the same topics around multi-tenant and real-time analytics sections just below (in section When to Use Citus?) In a sense, we're repeating our use cases benefits twice in the first page.

begriffs:
There was a lot of information in there. The first two paragraphs were a little overwhelming.

Now we focus on the *what* first and move some of the *why* down lower.

Fixes #791

### DIFF
--- a/get_started/what_is_citus.rst
+++ b/get_started/what_is_citus.rst
@@ -42,12 +42,6 @@ For more customers and statistics, see our `customer stories <https://www.citusd
 When to Use Citus
 =================
 
-Citus serves many use cases. Two common ones are scaling multi-tenant (B2B) databases and real-time analytics.
-
-Fast-growing `multi-tenant apps <https://www.citusdata.com/use-cases/multi-tenant-apps>`_ want to add new customers, deliver great performance, and not have to worry about database infrastructure. Data analysts want sub-second response times for customer-facing `analytics dashboards <https://www.citusdata.com/use-cases/real-time-analytics>`_, even with real-time ingestion, very large data sets, complex queries, and lots of concurrent users.
-
-Citus allows these and other applications to enjoy the power and familiarity of a traditional relational database, but with the capability for massive scale. Applications connect to and use a Citus-enabled PostgreSQL database just like they would a traditional PostgreSQL database. Citus uses the :ref:`same SQL commands <citus_sql_reference>` that developers and frameworks already know.
-
 .. _mt_blurb:
 
 Multi-Tenant Database
--- a/get_started/what_is_citus.rst
+++ b/get_started/what_is_citus.rst
@@ -22,11 +22,9 @@ Citus scales horizontally by adding worker nodes, vertically by upgrading worker
 
 * `Algolia <https://www.citusdata.com/customers/algolia>`_
     * 5-10B rows ingested per day
-* `Kixer <https://www.citusdata.com/customers/kixer>`_
-    * >1B events ingested per day
 * `Heap <https://www.citusdata.com/customers/heap>`_
     * 500+ billion events
-    * 500TB of data on a 40-node Citus database cluster 
+    * 900TB of data on a 40-node Citus database cluster
 * `Chartbeat <https://www.citusdata.com/customers/chartbeat>`_
     * >2.6B rows of data added per month
 * `Pex <https://www.citusdata.com/customers/pex>`_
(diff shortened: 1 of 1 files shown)


get_started/what_is_citus.rst
```diff
@@ -66,7 +44,31 @@ For more customers and statistics, see our `customer stories <https://www.citusd
...
+Citus serves many use cases. Two common ones are scaling multi-tenant (B2B) databases and real-time analytics.
+
+Fast-growing `multi-tenant apps <https://www.citusdata.com/use-cases/multi-tenant-apps>`_ want to add new customers, deliver great performance, and not have to worry about database infrastructure. Data analysts want sub-second response times for customer-facing `analytics dashboards <https://www.citusdata.com/use-cases/real-time-analytics>`_, even with real-time ingestion, very large data sets, complex queries, and lots of concurrent users.
+
+Citus allows these and other applications to enjoy the power and familiarity of a traditional relational database, but with the capability for massive scale. Applications connect to and use a Citus-enabled PostgreSQL database just like they would a traditional PostgreSQL database. Citus uses the :ref:`same SQL commands <citus_sql_reference>` that developers and frameworks already know.
+
++--------------------------------------------------------+-------------------------------------------------------------+
+| Multi-Tenant Advantages                                | Real-Time Advantages                                        |
```
sumedhpathak: I'm now wondering if we just fold in these advantages as bullet points in the below subsections. I recall I had recommended the table approach, but this seems like it compares one vs the other, which isn't quite what we want. 
Could you make that change @begriffs ?

begriffs: I see what you mean about the comparison feeling -- and that a user doesn't look at the advantages of each in order to determine which to use, but rather matches their modeling needs. I do like how clear the advantages in the table are at a glance, so I'll convert the table rows to bullet points like you suggest.

begriffs:
Fixes #790 as well

get_started/what_is_citus.rst
```diff
@@ -35,7 +33,7 @@ Citus scales horizontally by adding worker nodes, vertically by upgrading worker
     * 2.4TB memory, 1280 cores, and 60TB of data
     * ...with plans to grow to 45 nodes
 * `Mixrank <https://www.citusdata.com/customers/mixrank>`_
-    * 160TB of time series data
+    * 1600TB of time series data
```
sumedhpathak: Minor, but 1.6PB sounds better.

get_started/what_is_citus.rst
```diff
@@ -66,7 +42,11 @@ For more customers and statistics, see our `customer stories <https://www.citusd
 When to Use Citus
 =================
 
-Citus serves many use cases. Two common ones are scaling multi-tenant (B2B) databases and real-time analytics. In addition to the information below, there are examples of Citus `use-cases <https://www.citusdata.com/use-cases>`_ and `customer case studies <https://www.citusdata.com/solutions/case-studies>`_ on our main web site.
+Citus serves many use cases. Two common ones are scaling multi-tenant (B2B) databases and real-time analytics.
```
sumedhpathak: Could we add some verbiage here like:
"Citus provides a lot of benefits for these use-cases, and below are some of the reasons why Citus is the best choice for these use-cases" (An amended version of that).

Another option is the move the below 2 paragraphs into a "Why Citus" right before "When to use".

get_started/what_is_citus.rst
begriffs: (This has been corrected but github still shows the comment perhaps because it's attached to an individual commit)

get_started/what_is_citus.rst
begriffs: I was trying various things moving that content around, and it just felt redundant and long-winded. See how you like it with the paragraphs removed.

PR #797, opened 2018-10-22
use JOIN keyword instead of obsolete implicit joins

wbolster:
Quoting [A history lesson on SQL joins (in Postgres)](https://www.citusdata.com/blog/2018/09/25/history-on-joins-in-postgres/) on the Citus blog:

> and if your intention is not to implement a CROSS JOIN then using a comma separated list of tables in your FROM clause looks deadly suspicious.


wbolster:
you're welcome :kissing_closed_eyes:

PR #800, opened 2018-10-25
New views etc in 8.0

begriffs:
(Opening this PR, but going to also rearrange some sections)

### DIFF
--- a/arch/mx.rst
+++ b/arch/mx.rst
@@ -104,14 +104,19 @@ An important aspect to consider is that horizontally scaling out your processing
 MX Limitations
 --------------
 
-Although MX allows direct reading and writing from worker nodes, it doesn't support all commands on workers.
+Although MX allows direct reading and writing from worker nodes, it doesn't support all commands on workers. The coordinator node is the authoritative source of Citus metadata, so queries that change metadata must happen via the coordinator.
 
-Limitations and unsupported statements for execution on a worker node:
+Supported only via coordinator
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-* :ref:`DDL <ddl>` commands on distributed tables are not executable from workers, and must be done through the coordinator.
-* Workers may not call :ref:`user_defined_functions` that change Citus metadata.
-* ``CREATE VIEW`` is not propagated to other worker nodes.
-* Preparing statements (e.g. ``PREPARE p1 AS query``) is not propagated to other worker nodes.
+* :ref:`DDL <ddl>` commands.
+* :ref:`user_defined_functions` that change Citus metadata.
+* Queries accessing :ref:`append distributed <append_distribution>` tables.
+* ``CREATE VIEW`` is only propagated to other nodes when run from the coordinator.
+* Prepared statements (e.g. ``PREPARE p1 AS query``).
+
+Other query limitations
+~~~~~~~~~~~~~~~~~~~~~~~
+
 * Foreign data wrappers, including ``cstore_fdw``, are not supported with Citus MX.
 * Serial columns must have type "bigserial." Globally in the cluster the sequence values will not be monotonically increasing because the sixteen most significant bits hold the worker node id.
-* Queries from workers cannot access :ref:`append distributed <append_distribution>` tables.
--- a/arch/mx.rst
+++ b/arch/mx.rst
@@ -3,7 +3,7 @@
 Citus MX
 ========
 
-Citus MX is a new version of Citus that adds the ability to use hash-distributed tables from any node in a Citus cluster, which allows you to scale out your query throughput by opening many connections across all the nodes. This is particularly useful for performing small reads and writes at a very high rate in a way that scales horizontally. Citus MX is currently available in private beta on `Citus Cloud <https://www.citusdata.com/product/cloud>`_.
+Citus MX is a new version of Citus that adds the ability to use hash-distributed tables from any node in a Citus cluster, which allows you to scale out your query throughput by opening many connections across all the nodes. This is particularly useful for performing small reads and writes at a very high rate in a way that scales horizontally. Citus MX is currently available in Citus Enterprise Edition and on `Citus Cloud <https://www.citusdata.com/product/cloud>`_.
 
 In the Citus MX architecture, all nodes are PostgreSQL servers running the Citus extension. One node is acting as coordinator and the others as data nodes, each node also has a hot standby that automatically takes over in case of failure. The coordinator is the authoritative source of metadata for the cluster and data nodes store the actual data in shards. Distributed tables can only be created, altered, or dropped via the coordinator, but can be queried from any node. When making changes to a table (e.g. adding a column), the metadata for the distributed tables is propagated to the workers using PostgreSQL’s built-in 2PC mechanism and distributed locks. This ensures that the metadata is always consistent such that every node can run distributed queries in a reliable way.
 
--- a/arch/mx.rst
+++ b/arch/mx.rst
@@ -1,7 +1,7 @@
 .. _mx:
 
-Masterless Mode
-===============
+Citus MX
+========
 
 Citus MX is a new version of Citus that adds the ability to use hash-distributed tables from any node in a Citus cluster, which allows you to scale out your query throughput by opening many connections across all the nodes. This is particularly useful for performing small reads and writes at a very high rate in a way that scales horizontally. Citus MX is currently available in private beta on `Citus Cloud <https://www.citusdata.com/product/cloud>`_.
 
--- a/arch/mx.rst
+++ b/arch/mx.rst
@@ -101,11 +101,17 @@ The ability to perform analytical SQL queries combined with high volume data ing
 
 An important aspect to consider is that horizontally scaling out your processing power ensures that indexes don’t necessarily become an ingestion bottleneck as your application grows. PostgreSQL has very powerful indexing capabilities and with the ability to scale out you can almost always get the desired read- and write-performance.
 
-Limitations Compared to Citus
------------------------------
+MX Limitations
+--------------
 
-All Citus 7.4 features are supported in Citus MX with the following exceptions:
+Although MX allows direct reading and writing from worker nodes, it doesn't support all commands on workers.
 
-Append-distributed tables currently cannot be made available from workers. They can still be used in the traditional way, with queries going through the coordinator. However, append-distributed tables already allowed you to :ref:`bulk_copy`.
+Limitations and unsupported statements for execution on a worker node:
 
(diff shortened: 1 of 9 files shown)


begriffs:
It's nice to have architecture stuff put together in one place. Although the top level TOC is starting to get long again... I wonder if we should compensate by moving another top level thing inside another one. Like put Install under Getting Started, or Use Case Guides under Develop...

arch/mx.rst
```diff
@@ -1,13 +1,10 @@
 .. _mx:
 
-Masterless Mode (beta)
-======================
+Masterless Mode
+===============
 
 Citus MX is a new version of Citus that adds the ability to use hash-distributed tables from any node in a Citus cluster, which allows you to scale out your query throughput by opening many connections across all the nodes. This is particularly useful for performing small reads and writes at a very high rate in a way that scales horizontally. Citus MX is currently available in private beta on `Citus Cloud <https://www.citusdata.com/product/cloud>`_.
 
-Architecture
```
sumedhpathak: Any reason to remove this?

arch/mx.rst
```diff
@@ -1,13 +1,10 @@
 .. _mx:
 
-Masterless Mode (beta)
-======================
+Masterless Mode
```
sumedhpathak: See below (since I can't comment on unchanged lines). In "Limitations compared to Citus" we note 7.4. Can we update that to 8.0? Also, in Onder's email titled "Highlights from Citus 8.0: Citus MX" , see section: "Limitations/Differences of Citus MX compared to single-coordinator Citus". That section is a bit detailed, so we'll need to condense that. But we should update the limitations section with this content.

get_started/concepts.rst
```diff
@@ -1,26 +1,20 @@
 .. _citus_concepts:
 
-
-Concepts
```
sumedhpathak: I would still like to keep the Concepts part. In the left TOC, seeing Architecture --> Nodes seems unclear. We could have just 2 top-level items there for now "Concepts" and the MX architecture.

arch/mx.rst
```diff
@@ -1,13 +1,10 @@
 .. _mx:
 
-Masterless Mode (beta)
```
sumedhpathak: Let's rename this. We've moved away from 'master' to 'coordinator'. What are our options?
- MX Architecture? (Stick with MX branding)
- Citus MX

develop/api_metadata.rst
```diff
@@ -366,3 +333,203 @@ Caveats:
...
+In this example the queries originated on the coordinator, but the view can also list locks between queries originating on workers (executed with Citus MX for instance).
+
+Tables on all Nodes
+-------------------
+
+Citus has other informational tables and views which are accessible on all nodes, not just the coordinator.
+
+.. _pg_dist_authinfo:
+
+Connection Credentials Table
```
sumedhpathak: We also have introduced pg_dist_pol_info. Could you capture that as well? Thanks!

develop/api_metadata.rst
```diff
@@ -366,3 +333,203 @@ Caveats:
 * It's a coordinator node feature, with no :ref:`Citus MX <mx>` support
 * Tracks a limited number of queries, set by the ``pg_stat_statements.max`` GUC (default 5000)
 * To truncate the table, use the ``citus_stat_statements_reset()`` function
+
+Distributed Query Activity
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+With :ref:`mx` users can execute distributed queries from any node. Examining the standard Postgres `pg_stat_activity <https://www.postgresql.org/docs/current/static/monitoring-stats.html#PG-STAT-ACTIVITY-VIEW>`_ view on the coordinator won't include those worker-initiated queries. Also queries might get blocked on row-level locks on one of the shards on a worker node. If that happens then those queries would not show up in `pg_locks <https://www.postgresql.org/docs/current/static/view-pg-locks.html>`_ on the Citus coordinator node.
```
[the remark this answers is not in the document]
begriffs: Yep, the `:ref:` command substitutes with whatever name the destination section has, or can be overridden by

```
:ref:`new name <link>`
```

But in our case updating the title of the mx link does the trick.

arch/mx.rst
begriffs: After the file move (`cloud/additional_mx.rst` → `arch/mx.rst`), the MX section now lives under Architecture in the TOC.

"Architecture > Citus MX > Architecture" looked kind of strange when expanding the TOC.

arch/mx.rst
begriffs: I'll take care of the version updates in the release checklist PR, and the MX limitations in this one.

arch/mx.rst
begriffs: I'll stick with "Citus MX"

get_started/concepts.rst
begriffs: Restored it.

arch/mx.rst
begriffs: Added a concise list of limitations

develop/api_guc.rst
```diff
@@ -99,6 +99,35 @@ Citus honors only a whitelisted subset of the options, namely:
 
 *(† = subject to the runtime presence of optional PostgreSQL features)*
 
+.. _override_table_visibility:
```
aamederen: Currently, the MX docs say `If you connect to the data URL using psql and run \d, then you will see all the distributed tables and some of the shards. ` We should mention this guc there as well.

begriffs: Good point, thanks.

arch/mx.rst
```diff
@@ -104,11 +101,17 @@ The ability to perform analytical SQL queries combined with high volume data ing
...
 
-All Citus 7.4 features are supported in Citus MX with the following exceptions:
+Although MX allows direct reading and writing from worker nodes, it doesn't support all commands on workers.
 
-Append-distributed tables currently cannot be made available from workers. They can still be used in the traditional way, with queries going through the coordinator. However, append-distributed tables already allowed you to :ref:`bulk_copy`.
+Limitations and unsupported statements for execution on a worker node:
 
-When performing writes on a hash-distributed table with a bigserial column via the data URL, sequence numbers are no longer monotonic, but instead have the form <16-bit unique node ID><48-bit local sequence number> to ensure uniqueness. The coordinator node always has node ID 0, meaning it will generate sequence numbers as normal. Serial types smaller than bigserial cannot be used in distributed tables.
+* :ref:`DDL <ddl>` commands on distributed tables are not executable from workers, and must be done through the coordinator.
+* Workers may not call :ref:`user_defined_functions` that change Citus metadata.
```
sumedhpathak: This can be done via the coordinator. Can we add a separate section like:
"Supported only via the Coordinator", and add this and the others which work via the coordinator to that? (DDL and Create-table UDF's) 
We can also clarify in that section that since the Coordinator is the authoritative source of the metadata, changes to it only happen via that node, and are propagated to the worker nodes. (Thus the limitation).

develop/api_metadata.rst
```diff
@@ -366,3 +333,231 @@ Caveats:
...
+.. note::
+
+   :ref:`Shard rebalancing <shard_rebalancing>` is not compatible with connection poolers such as pgbouncer and ignores the settings in ``pg_dist_poolinfo``.
+
+.. _worker_shards:
+
+Shards and Indices on Workers
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Worker nodes store shards as tables that are ordinarily hidden (see :ref:`override_table_visibility`). The easiest way to obtain information about the shards on each worker is to consult that worker's ``citus_shards_on_worker`` view. For instance, here are some shards on a worker for the distributed table ``test_table``:
```
sumedhpathak: Does this view also exist on the coordinator?

develop/api_metadata.rst
```diff
@@ -366,3 +333,231 @@ Caveats:
...
+    (1 row)
+
+Connection Pooling Credentials
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. note::
+
+  This table is a part of Citus Enterprise Edition. Please `contact us <https://www.citusdata.com/about/contact_us>`_ to obtain this functionality.
+
+The ``pg_dist_poolinfo`` table holds the host, port and database name for Citus to use when connecting to a node through a connection pooler. This table is specific to connection pooling rather than direct connections between nodes, and overrides the values in :ref:`pg_dist_node <pg_dist_node>`.
```
sumedhpathak: What do you think of the following reframing: "If you want to use a connection pooler to connect to a node, you can set it up using this metadata table. If pool information is present, Citus will try to use that instead of the direct connection to the information in pg_dist_node. In certain scenarios (e.g. rebalancing) CItus will connect directly since logical replication..."

get_started/concepts.rst
```diff
@@ -1,26 +1,25 @@
...
 .. _distributed_arch:
 
-Distributed Architecture
-------------------------
+Nodes
+=====
 
 Citus is a PostgreSQL `extension <https://www.postgresql.org/docs/current/static/external-extensions.html>`_ that allows commodity database servers (called *nodes*) to coordinate with one another in a "shared nothing" architecture. The nodes form a *cluster* that allows PostgreSQL to hold more data and use more CPU cores than would be possible on a single computer. This architecture also allows the database to scale by simply adding more nodes to the cluster.
 
-Nodes: Coordinator and Workers
```
sumedhpathak: I still prefer to explicitly call out Workers and Coordinators.

performance/performance_tuning.rst
```diff
@@ -411,9 +411,9 @@ From a throughput standpoint, you can expect data ingest ratios of 250K - 2M row
...
 
-Masterless Citus (50k/s-500k/s)
-===============================
+Citus MX (50k/s-500k/s)
+=======================
 
-Masterless Citus (Citus MX) builds on the Citus extension. It gives you the ability to query and write to distributed tables from any node, which allows you to horizontally scale out your write-throughput using PostgreSQL. It also removes the need to interact with a primary node in a Citus cluster for data ingest or queries.
+Citus MX builds on the Citus extension. It gives you the ability to query and write to distributed tables from any node, which allows you to horizontally scale out your write-throughput using PostgreSQL. It also removes the need to interact with a primary node in a Citus cluster for data ingest or queries.
 
 Citus MX is currently available in private beta on Citus Cloud. For more information see :ref:`mx`.
```
sumedhpathak: We probably want to change this to: Citus MX is available in our Enterprise Edition and Citus Cloud. For more information ..."

begriffs: D'oh. Just did a search to find all references to "beta" and fixed that.

develop/api_metadata.rst
begriffs: I like it. Worked it in.

develop/api_metadata.rst
begriffs: It does, and is empty.

However... Now that I test it, it's empty on the workers too even with a table distributed. Asked people to check this out in slack.

onderkalaci: Please see my comment regarding this in the above comments.

begriffs:
@onderkalaci thanks for the explanation. Since I closed this PR too soon, I created another one with your suggestions: https://github.com/citusdata/citus_docs/pull/810

PR #803, opened 2018-10-26
fix broken sphinx install link

hanefi:
Generating HTMLs for the first time. Realized that sphinx installation link was broken

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,4 +14,4 @@ To generate HTML version:
 
 ---
 
-**Sphinx Installation Note:** If you're on OSX you might want to install the Python from homebrew - then a simple `pip install sphinx` does the trick.
+**Sphinx Installation Note:** If you're on OSX you might want to install the Python from homebrew - then a simple `brew install sphinx-doc` or `pip install sphinx` does the trick.
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 To generate HTML version:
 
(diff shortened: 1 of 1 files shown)


hanefi:
As it turns out, sphinx-doc brew formula is keg-only. It is advised to use pip instead. I will remove the brew command

```
sphinx-doc is keg-only, which means it was not symlinked into /usr/local,
because this formula is mainly used internally by other formulae.
Users are advised to use `pip` to install sphinx-doc.

If you need to have sphinx-doc first in your PATH run:
  echo 'export PATH="/usr/local/opt/sphinx-doc/bin:$PATH"' >> ~/.zshrc
```

README.md
```diff
@@ -14,4 +14,4 @@ To generate HTML version:
 
 ---
 
-**Sphinx Installation Note:** If you're on OSX you might want to install the Python from homebrew - then a simple `pip install sphinx` does the trick.
+**Sphinx Installation Note:** If you're on OSX you might want to install the Python package from pip - then a simple `pip install sphinx` does the trick.
```
begriffs: Nice research about the keg-only issue. Maybe we want to provide more context here, like:

```markdown
**Sphinx Installation Note:** on OS X it's better to install sphinx via
[pip](https://pip.pypa.io/en/stable/installing/) rather than Homebrew.
(The brew formula is keg-only and used primarily by other tools.)
Use `pip install sphinx`. 
```

PR #807, opened 2018-11-02
Add warning about master_update_node blocking

### LINKED ISSUE #595: Document master_update_node()

begriffs:
Fixes #595 

@ozgune is this what you wanted to warn people about?

To be honest I don't know exactly the sequence of how to use `master_update_node` -- if someone calls it after already changing a worker node's hostname then I don't see how existing queries could be running on the node and blocking the function. Or is this UDF meant to be called prior to actually changing the hostname?

### DIFF
--- a/develop/api_udf.rst
+++ b/develop/api_udf.rst
@@ -407,7 +407,7 @@ Example
 master_update_node
 $$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$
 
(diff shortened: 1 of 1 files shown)


sumedhpathak:
@begriffs You might want to send @ozgune an email or setup a time to follow up. 
My understanding is that `master_update_node` will lock the placements on that node. So if someone tries to insert to a placement on that node, it will block until `master_update_node` finishes. The use-case is that during a failover, we do: `BEGIN; master_update_node()`. Then we actually process the failover, while in the meantime writes are queued up. After the failover is complete, we `commit`, releasing the locks. The inserts then go to the right new node.

PR #802, opened 2018-10-26
Preparing for 8.0

begriffs:
Fixes #798 

We'll probably want to rebase this on top of the what is citus changes and new 8.0 features pr once they are both merged. That way I can check that there are still no sphinx warnings when it all gets put together.

PR #808, opened 2018-11-05
Bump Citus version to 8.0.0

velioglu:
Bump Citus version to 8.0.0


begriffs:
Thanks, although I'll take care of all version bumps in #802.

PR #810, opened 2018-11-06
Onder's suggestions about MX

begriffs:


begriffs:
@sumedhpathak and @onderkalaci - I'm going to merge this so I can continue with the 8.0 docs release. We can iterate in the new version if there is more to cover.

PR #813, opened 2018-11-08
Warn not to use docker for prod

### LINKED ISSUE #811: Note that Docker is for Development only
We should set the expectation on Docker installation that it is for Development / Testing reasons only. The image isn't primed for production use, and we should discourage using it for the same.

We should add a note/comment on http://docs.citusdata.com/en/v8.0/installation/single_machine_docker.html.

begriffs:
Fixes #811


begriffs:
I'll cherry-pick this to the v8.0 branch

sumedhpathak:
Great, thanks.

PR #814, opened 2018-11-13
Fix migration issues that Sai pointed out

### LINKED ISSUE #561: How to modify continuous integration for an application that uses Citus

People create Rails/Django apps with Citus as the backend and wish to run their integration tests on Circle or Travis. Provide instructions for how to set up a Citus cluster on CI for the application.

Example Circle and Travis config files showing the migration from Postgres to Citus.

New subsection in https://docs.citusdata.com/en/v7.1/migration/transitioning.html#app-migration

### LINKED ISSUE #773: Include distribution column in keys need not be in the Prepare Tables for Migration section
In the [migration](http://docs.citusdata.com/en/v7.5/develop/migration.html) section, **[Prepare Tables for Migration](http://docs.citusdata.com/en/v7.5/develop/migration_mt_schema.html#prepare-tables-for-migration)** is a step that happens on an existing single node production database. There is no need to **[Include distribution column in keys](http://docs.citusdata.com/en/v7.5/develop/migration_mt_schema.html#include-distribution-column-in-keys)** on the production tables.

We can push it to the [next section](http://docs.citusdata.com/en/v7.5/develop/migration_mt_query.html) (Prepare Application for Migration) because we need to add the distribution column to unique/primary/foreign keys in the app that fits citus. Also worth making a note that our rails gem automatically changes the primary keys if you follow the steps.

### LINKED ISSUE #774: Document write-only-mode for rails gem
Similar to [this](https://github.com/citusdata/activerecord-multi-tenant#rolling-out-activerecord-multi-tenant-for-your-application-write-only-mode) section int the ar-multitenant github repo. We can create a separate page and link it in the **Prepare Tables for Migration** or the [Prepare Application for Migration](http://docs.citusdata.com/en/v7.5/develop/migration_mt_query.html) section.

### LINKED ISSUE #775: Document SQL command to perform backfilling periodically
We can add this to the [Backfill newly created columns](http://docs.citusdata.com/en/v7.5/develop/migration_mt_schema.html#backfill-newly-created-columns) section. It would be on the below lines (please check the correctness, it might be wrong):
```
UPDATE table_being_backfilled tb
SET tenant_id=t
FROM table_being_used tu
WHERE AND tb.tuid=tu.id AND tb.tenant_id IS NULL LIMIT 10000;

SELECT pg_cron(every 30 minutes, function with above query and VACUUM)
````

### LINKED ISSUE #777: Make the app stack migration visible
With the new migration page, the libraries/blogs/docs for app stacks are not visible in the index. Worth adding them to make the docs developer friendly.

```
Ruby on Rails migration - uses the activerecord-multi-tenant Ruby gem
Django migration - uses the django-multitenant Python library
ASP.NET migration - uses the 3rd party SAASkit
Java Hibernate - a blog post about scoping queries to tenants
```

### LINKED ISSUE #778: Tell the user to enable ssl to be able to connect to Citus Cloud.
Citus Cloud url by default requires ssl. We have seen quite a number of times that customers who migrate over from other databases services (which don't require ssl) hit this issue that they can't connect to the citus cluster. Once ssl is enabled it just works. Documenting this would be useful. A good location could be [Prepare Application for Migration](https://docs.citusdata.com/en/v7.5/develop/migration_mt_query.html)

We could start with just mentioning to enable ssl in app and then later give examples (as below) on how to do it in some app stacks.

i) [rails example 1](https://gist.github.com/mrrooijen/294b24140a18be98e0d5e0e6bca818b6) 
ii) [rails example 2](https://accidentaltechnologist.com/ruby-on-rails/ruby-rails-supporting-ssl-postgresql-heroku/)

### LINKED ISSUE #793: Improvements in data migration section

* We currently say >200gb etc, but could just say big and small databases to be more general
* After warp starts, there are things customers should be cautious about
  * Want to emphasize this things to watch out for when warping starts (see internal wiki)
  * We do mention that basebackup can start running up disk space, but we should mention that even during the streaming if the destination gets messed up then the replication slot can also cause the same type of problem
  * Sai can share a command that can drop the problematic replication slot so people can try again from the beginning

### LINKED ISSUE #804: Add instructions to change the database cleaner strategy to truncation
We've seen in the past that users have to change their database cleaner strategy to truncation when testing their test suite. (Active Record apps)

Keeping it at transaction sometimes leads to subsequent tests failing.

We should add a section to docs to document this in the rails migration section.

### LINKED ISSUE #805: Google doc rather than downloadable spreadsheet

We currently provide a downloadable spreadsheet to help people plan how to distribute tables. For convenience, turn that into a read-only google document that people can clone.

* Populate it with examples from our multi-tenant use case
* Point out where those tables come from in the doc

http://docs.citusdata.com/en/v7.5/develop/migration_mt_schema.html#identify-types-of-tables

### PR DESCRIPTION
**Pull request by begriffs**
Fixes #778 
Fixes #777 
Fixes #775 
Fixes #773 
Fixes #774 
Fixes #805 
Fixes #561 
Fixes #804 
Fixes #793

### DIFF
--- a/cloud/security.rst
+++ b/cloud/security.rst
@@ -40,7 +40,7 @@ Requiring SSL, plus verifying against a certificate:
      adapter: postgresql
      encoding: unicode
      sslmode: verify-full
-     sslrootcert: citus.crt
+     sslrootcert: /path/to/citus.crt
      url: postgres://user:password@host:port/db
 
 Two-Factor Authentication
--- a/cloud/security.rst
+++ b/cloud/security.rst
@@ -1,6 +1,8 @@
 Security
 ########
 
+.. _cloud_ssl:
+
 Connecting with SSL
 ===================
 
@@ -17,6 +19,29 @@ The string may need to be quoted in your shell to preserve the ampersand.
 .. note::
 
   Database clients must support SSL to connect to Citus Cloud. In particular :code:`psql` needs to be compiled :code:`--with-openssl` if building PostgreSQL from source.
+
+Application frameworks can customize database connection parameters using config files. For example, Ruby on Rails uses a file called `config/database.yml <https://guides.rubyonrails.org/configuring.html#configuring-a-database>`_. This file accepts ``sslmode`` and ``sslrootcert`` options to specify connection security. Here are two examples.
+
+Requiring SSL for the production db connection:
+
+.. code-block:: yaml
+
+   production:
+     adapter: postgresql
+     encoding: unicode
+     sslmode: require
+     url: postgres://user:password@host:port/db
+
+Requiring SSL, plus verifying against a certificate:
+
+.. code-block:: yaml
+
+   production:
+     adapter: postgresql
+     encoding: unicode
+     sslmode: verify-full
+     sslrootcert: citus.crt
+     url: postgres://user:password@host:port/db
 
 Two-Factor Authentication
 -------------------------
--- a/develop/migration_mt_query.rst
+++ b/develop/migration_mt_query.rst
@@ -33,7 +33,7 @@ Include distribution column in keys
 
 Citus :ref:`cannot enforce <non_distribution_uniqueness>` uniqueness constraints unless a unique index or primary key contains the distribution column. Thus we must modify primary and foreign keys in our example to include store_id.
 
-Here are SQL commands to turn the simple keys composite in the development database:
+Some of the libraries listed in the next section are able to help migrate the database schema to include the distribution column in keys. However, here is an example of the underlying SQL commands to turn the simple keys composite in the development database:
 
 .. code-block:: sql
 
--- a/develop/migration_mt_query.rst
+++ b/develop/migration_mt_query.rst
@@ -74,6 +74,11 @@ Finally, when joining tables make sure to filter by tenant id too. For instance 
      AND l.store_id='8c69aa0d-3f13-4440-86ca-443566c1fc75'
      AND p.store_id='8c69aa0d-3f13-4440-86ca-443566c1fc75'
 
+Enable Secure Connections
+-------------------------
+
+Clients should connect to Citus with SSL to protect information and prevent man-in-the-middle attacks. Citus Cloud, in fact, rejects unencrypted connections. To learn how to make an SSL connection see :ref:`cloud_ssl`.
+
 Check for cross-node traffic
 ----------------------------
 
--- a/develop/migration_mt_query.rst
+++ b/develop/migration_mt_query.rst
@@ -3,12 +3,12 @@
 Prepare Application for Citus
 =============================
 
-Once the distribution key is present on all appropriate tables, the application needs to include it in queries.
+Once the distribution key is present on all appropriate tables, the application needs to include it in queries. The following steps should be done using a copy of the application running in a development environment, and testing against a Citus back-end. After the application is working with Citus we'll see how to migrate production data from the source database into a real Citus cluster.
 
 Add distribution key to queries
 -------------------------------
 
-To execute queries efficiently for a specific tenant, Citus needs to route them to the appropriate node and run them there. Thus every query must identify which tenant it involves. For simple select, update, and delete queries this means that the *where* clause must filter by tenant id.
+To execute queries efficiently for a specific tenant, Citus needs to route them to the appropriate node and run them there. Thus **every query must identify which tenant it involves**. For simple select, update, and delete queries this means that the *where* clause must filter by tenant id.
 
 * Application code and any other ingestion processes that write to the tables should be updated to include the new columns.
 * Running the application test suite against the modified schema on Citus is a good way to determine which areas of the code need to be modified.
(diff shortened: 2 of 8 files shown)

### CODE REVIEW

File: develop/migration_mt_schema.rst [hunk 1]
```diff
@@ -112,4 +112,39 @@ We backfill the table by obtaining the missing values from a join query with ord
...
+
+.. code-block:: postgresql
+
+   -- the a function to backfill up to
+   -- one thousand rows from line_items
+
+   CREATE FUNCTION backfill_batch()
+   RETURNS void LANGUAGE sql AS $$
+     WITH batch AS (
+       SELECT *
```
begriffs: select less here

develop/migration_mt_schema.rst
```diff
@@ -18,7 +18,7 @@ We are happy to help review your environment to be sure that the ideal distribut
 Identify types of tables
 ------------------------
 
-Once a distribution key is identified, review the schema to identify how each table will be handled and whether any modifications to table layouts will be required. We typically advise tracking this with a spreadsheet, and have created a `template <https://examples.citusdata.com/citus-migration-plan.xlsx>`_ you can use.
+Once a distribution key is identified, review the schema to identify how each table will be handled and whether any modifications to table layouts will be required. We typically advise tracking this with a spreadsheet, and have created a `template <https://docs.google.com/spreadsheets/d/1jYlc22lHdP91pTrb6s35QfrN9nTE1BkVJnCSZeQ7ZmI/edit>`_ you can use.
```
begriffs: @saicitus or @colton-citus : do you have some nice example data for this spreadsheet? Or should I use the same tables that we have in the migration docs?

begriffs:
I pushed a change for #773 in particular. It does make other sections more complicated, because the production citus cluster to be should pull the schema from the development database now rather than the old production database. The dev db had the additional step of making its keys composite to include the distribution column. Better review my changes carefully.

cloud/security.rst
```diff
@@ -18,6 +20,29 @@ The string may need to be quoted in your shell to preserve the ampersand.
...
+
+Requiring SSL, plus verifying against a certificate:
+
+.. code-block:: yaml
+
+   production:
+     adapter: postgresql
+     encoding: unicode
+     sslmode: verify-full
+     sslrootcert: citus.crt
```
colton-citus: Might want to make this `/path/to/citus.crt` to make formatting clear

develop/migration_mt_query.rst
```diff
@@ -1,33 +1,101 @@
...
-Prepare Application for Migration
-=================================
+Prepare Application for Citus
+=============================
 
-Once the distribution key is present on all appropriate tables, the application needs to include it in queries.
+Set up Development Citus Cluster
+--------------------------------
+
+When modifying the application to work with Citus, you'll need a database to test against. Follow the instructions to set up a :ref:`development` of your choice.
```
colton-citus: This should probably say "development environment of your choice"

develop/migration_mt_query.rst
```diff
@@ -1,33 +1,101 @@
...
+
+Be sure to modify data flows to add keys to incoming data.
 
 Add distribution key to queries
 -------------------------------
 
-To execute queries efficiently for a specific tenant, Citus needs to route them to the appropriate node and run them there. Thus every query must identify which tenant it involves. For simple select, update, and delete queries this means that the *where* clause must filter by tenant id.
+Once the distribution key is present on all appropriate tables, the application needs to include it in queries. The following steps should be done using a copy of the application running in a development environment, and testing against a Citus back-end. After the application is working with Citus we'll see how to migrate production data from the source database into a real Citus cluster.
+
+To execute queries efficiently for a specific tenant, Citus needs to route them to the appropriate node and run them there. Thus **every query must identify which tenant it involves**. For simple select, update, and delete queries this means that the *where* clause must filter by tenant id.
```
colton-citus: We should probably make it more clear that it's _possible_ to do cross-tenant queries, but that we should add the tenant ID where practical. I don't want them to miss out on entire workloads.

develop/migration_mt_schema.rst
colton-citus: I'm not sure if we want to pre-fill the spreadsheet, but if you're sure that we do, just use the same tables you already use as examples.

develop/migration_mt_query.rst
begriffs: That word is an internal section identifier that gets replaced by the title of the section in the user-visible output. So in this case it would say "to setup a Single-Machine Cluster of your choice."

An alternate form in RestructuredText is to say
```rst
:ref:`new name <section_id>`
```
to override the name.

develop/migration_mt_schema.rst
```diff
@@ -112,4 +71,40 @@ We backfill the table by obtaining the missing values from a join query with ord
...
+   -- the a function to backfill up to
+   -- one thousand rows from line_items
+
+   CREATE FUNCTION backfill_batch()
+   RETURNS void LANGUAGE sql AS $$
+     WITH batch AS (
+       SELECT line_items_id, order_id
+         FROM line_items
+        WHERE store_id IS NULL
+        LIMIT 1000
```
saicitus: make this 10000

develop/migration_mt_schema.rst
```diff
@@ -112,4 +71,40 @@ We backfill the table by obtaining the missing values from a join query with ord
...
+     )
+     UPDATE line_items AS li
+        SET store_id = orders.store_id
+       FROM batch, orders
+      WHERE batch.line_item_id = li.line_item_id
+        AND batch.order_id = orders.order_id;
+   $$;
+
+   -- run the function every half hour
+   SELECT cron.schedule('*/30 * * * *', 'SELECT backfill_batch()');
```
saicitus: make this 15 mins

begriffs:
Sai: I tried moving the framework specific pages to an "ORM Integration" section under Develop but it just didn't flow well. Those sections are pretty specific to migration and don't seem appropriate as standalone material for how to using the framework libraries.

PR #819, opened 2018-11-20
More AZs supported in Cloud

begriffs:


begriffs:
@lfittl you move a ninja speed.

PR #820, opened 2018-11-21
Update SQL support section

### LINKED ISSUE #818: SQL Support and Workarounds are not up to date
This links should be updated:

http://docs.citusdata.com/en/v8.0/develop/reference_workarounds.html#sql-support-and-workarounds

## Supported Correlated Subqueries
Citus supports correlated subqueries as long as the correlation is on the distribution key and the subqueries conform to subquery pushdown rules (e.g., GROUP BYs on the distribution column, no LIMIT/LIMIT OFFSET in the subquery, DISTINCT/WINDOW functions on the distribution column). You can see some examples here: [subqueries in where clause](https://github.com/citusdata/citus/blob/master/src/test/regress/sql/multi_subquery_in_where_clause.sql#L30), [lateral joins](https://github.com/citusdata/citus/blob/master/src/test/regress/sql/multi_subquery_behavioral_analytics.sql#L433), [EXISTS-NOT EXISTS subqueries](https://github.com/citusdata/citus/blob/master/src/test/regress/sql/multi_subquery_behavioral_analytics.sql#L486-L507).

These types of subqueries could be very useful for real-time analytics.

## Modifying CTEs

Citus supports modifying CTEs for several releases now. The example [here](https://www.postgresql.org/docs/current/queries-with.html#QUERIES-WITH-MODIFYING) can be executed and many more. See the relevant [PR](https://github.com/citusdata/citus/pull/2006/files) for examples.

## SELECT FOR UPDATE

Citus supports `SELECT FOR UPDATE` when the query is a router query. So, multi-tenant use-case is covered, which was the mostly wanted use-case for `SELECT FOR UPDATE`. The relevant [PR](https://github.com/citusdata/citus/pull/2197).

begriffs:
Fixes #818 

@onderkalaci does this change look accurate to you? I broke the list into "not yet supported" and "partially supported."

### DIFF
--- a/develop/reference_workarounds.rst
+++ b/develop/reference_workarounds.rst
@@ -5,20 +5,23 @@ SQL Support and Workarounds
 
 As Citus provides distributed functionality by extending PostgreSQL, it is compatible with PostgreSQL constructs. This means that users can use the tools and features that come with the rich and extensible PostgreSQL ecosystem for distributed tables created with Citus.
 
-Citus supports all SQL queries on distributed tables, with only these exceptions:
+Citus supports all SQL queries on distributed tables, with a few exceptions.
 
-* Correlated subqueries
-* `Recursive <https://www.postgresql.org/docs/current/static/queries-with.html#idm46428713247840>`_/`modifying <https://www.postgresql.org/docs/current/static/queries-with.html#QUERIES-WITH-MODIFYING>`_ CTEs
+Not yet supported:
+
+* `Recursive CTEs <https://www.postgresql.org/docs/current/static/queries-with.html#idm46428713247840>`_
 * `TABLESAMPLE <https://www.postgresql.org/docs/current/static/sql-select.html#SQL-FROM>`_
-* `SELECT … FOR UPDATE <https://www.postgresql.org/docs/current/static/sql-select.html#SQL-FOR-UPDATE-SHARE>`_
 * `Grouping sets <https://www.postgresql.org/docs/current/static/queries-table-expressions.html#QUERIES-GROUPING-SETS>`_
-* `Window functions <https://www.postgresql.org/docs/current/static/tutorial-window.html>`_ that do not include the distribution column in PARTITION BY
+
+Partially supported:
+
+* `SELECT … FOR UPDATE <https://www.postgresql.org/docs/current/static/sql-select.html#SQL-FOR-UPDATE-SHARE>`_ are supported for :ref:`router_executor` queries.
+* `Window functions <https://www.postgresql.org/docs/current/static/tutorial-window.html>`_ are supported when they include the distribution column in PARTITION BY.
+* Correlated subqueries are supported when the correlation is on the :ref:`dist_column` and the subqueries conform to subquery pushdown rules (e.g., grouping by the distribution column, with no LIMIT or LIMIT OFFSET clause).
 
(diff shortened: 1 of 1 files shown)


develop/reference_workarounds.rst
```diff
@@ -5,20 +5,23 @@ SQL Support and Workarounds
...
 
-Citus supports all SQL queries on distributed tables, with only these exceptions:
+Citus supports all SQL queries on distributed tables, with a few exceptions.
 
-* Correlated subqueries
-* `Recursive <https://www.postgresql.org/docs/current/static/queries-with.html#idm46428713247840>`_/`modifying <https://www.postgresql.org/docs/current/static/queries-with.html#QUERIES-WITH-MODIFYING>`_ CTEs
+Not yet supported:
+
+* `Recursive CTEs <https://www.postgresql.org/docs/current/static/queries-with.html#idm46428713247840>`_
 * `TABLESAMPLE <https://www.postgresql.org/docs/current/static/sql-select.html#SQL-FROM>`_
```
onderkalaci: Ops, I forgot to mention. We also support `TABLESAMPLE` for router queries, see https://github.com/citusdata/citus/pull/2332. So, it should also be moved down to `Partially supported section`

develop/reference_workarounds.rst
```diff
@@ -5,20 +5,23 @@ SQL Support and Workarounds
...
 
-Furthermore, in :ref:`mt_use_case` when queries are filtered by table :ref:`dist_column` to a single tenant then all SQL features work, including the ones above.
+Partially supported:
 
-To learn more about PostgreSQL and its features, you can visit the `PostgreSQL documentation <http://www.postgresql.org/docs/current/static/index.html>`_.
+* `SELECT … FOR UPDATE <https://www.postgresql.org/docs/current/static/sql-select.html#SQL-FOR-UPDATE-SHARE>`_ are supported for :ref:`router_executor` queries.
+* `Window functions <https://www.postgresql.org/docs/current/static/tutorial-window.html>`_ are supported when they include the distribution column in PARTITION BY.
+* Correlated subqueries are supported when the correlation is on the :ref:`dist_column` and the subqueries conform to subquery pushdown rules (e.g., grouping by the distribution column, with no LIMIT or LIMIT OFFSET clause).
+
+Furthermore, in :ref:`mt_use_case` when queries are filtered by table :ref:`dist_column` to a single tenant then all SQL features work, including the ones above.
```
onderkalaci: The following comment could be a personal opinion. But, I think instead of giving the following sentence in the end of the section:
```
Furthermore, in :ref:`mt_use_case` when queries are filtered by table :ref:`dist_column` to a single tenant then all SQL features work, including the ones above.
```

It might be more powerful to move the following even before: `Citus supports all SQL queries on distributed tables, with a few exceptions`. Such as:

```
To start with, in :ref:`mt_use_case` when queries are filtered by table :ref:`dist_column` to a single tenant then all SQL features work, including the ones below.
``` 

(You can consider changing `To start with` phrase to something more informal maybe?)

And, if we do the above, I'd suggest changing the title `Partially Supported ` to `Limitations for Parallel SQL Queries ` and list everything under that list.

That seems more accurate, because we don't have anything that should go under `Not yet supported:`, since now everything works for router queries. In the past, that probably made sense because some of things were not working for router queries such as tablesample and modifying ctes.

begriffs: Good point. I added a commit to structure it like you suggested. See what you think.

PR #831, opened 2018-12-13
ON CONFLICT is supported via the coordinator

begriffs:


develop/reference_dml.rst
```diff
@@ -45,11 +45,13 @@ Sometimes it's convenient to put multiple insert statements together into a sing
...
 
-The second way of executing an ``INSERT … SELECT`` statement is selecting the results from worker nodes, pulling the data up to the coordinator node, and then issuing an INSERT statement from the coordinator with the data. Citus is forced to use this approach when the source and destination tables are not colocated. This method does not support ``ON CONFLICT``.
+The second way of executing an ``INSERT … SELECT`` statement is selecting the results from worker nodes, pulling the data up to the coordinator node, and then issuing an INSERT statement from the coordinator with the data. Citus is forced to use this approach when the source and destination tables are not colocated. Because of the network overhead, this method is not as efficient.
+
+If upserts are an important operation in your application, the ideal solution is to model the data so that the source and destination tables are colocated, and so that the distribution column can be part of the GROUP BY clause in the upsert statement (if aggregating).
```
sumedhpathak: Maybe just add that it would be ideal because everything would be parallelized across the worker nodes and will run fast.

PR #833, opened 2018-12-13
Demonstrate support for NOT VALID constraints

begriffs:
...maybe it would have been enough to say we support NOT VALID and VALIDATE CONSTRAINT, and point people to the postgres docs. However I put an example in here. See what you think.

### DIFF
--- a/develop/reference_ddl.rst
+++ b/develop/reference_ddl.rst
@@ -294,6 +294,60 @@ Not-null constraints can be applied to any column (distribution or not) because 
 
   ALTER TABLE ads ALTER COLUMN image_url SET NOT NULL;
 
+Using NOT VALID Constraints
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+In some situations it can be useful to enforce constraints for new rows, while allowing existing non-conforming rows to remain unchanged. Citus supports this feature for CHECK constraints and foreign keys, using PostgreSQL's "NOT VALID" constraint designation.
+
+For example, consider an application which stores user profiles in a :ref:`reference table <reference_tables>`.
+
+.. code-block:: postgres
+
+   -- we'll store user emails case insensitively, so enable
+   -- the "citext" data type on the coordinator and workers
+   CREATE EXTENSION citext;
+   SELECT run_command_on_workers('CREATE EXTENSION citext;');
+
+   CREATE TABLE users ( email citext PRIMARY KEY );
+   SELECT create_reference_table('users');
+
+In the course of time imagine that a few non-addresses get into the table.
+
+.. code-block:: postgres
+
+   INSERT INTO users VALUES
+      ('foo@example.com'), ('hacker12@aol.com'), ('lol');
+
+We would like to validate the addresses, but PostgreSQL does not ordinarily allow us to add a CHECK constraint that fails for existing rows. However it *does* allow a constraint marked not valid:
+
+.. code-block:: postgres
+
+   ALTER TABLE users
+   ADD CONSTRAINT syntactic_email
+   CHECK (email ~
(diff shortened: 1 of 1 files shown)


develop/reference_ddl.rst
```diff
@@ -294,6 +294,60 @@ Not-null constraints can be applied to any column (distribution or not) because
...
+      '^[a-zA-Z0-9.!#$%&''*+/=?^_`{|}~-]+@[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?(?:\.[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?)*$'
+   ) NOT VALID;
+
+This succeeds, and new rows are protected.
+
+.. code-block:: postgres
+
+   INSERT INTO users VALUES ('fake');
+
+   /*
```
sumedhpathak: Is this supposed to be here? The below '*' isn't closed.

develop/reference_ddl.rst
```diff
@@ -294,6 +294,60 @@ Not-null constraints can be applied to any column (distribution or not) because
...
+
+In some situations it can be useful to enforce constraints for new rows, while allowing existing non-conforming rows to remain unchanged. Citus supports this feature for CHECK constraints and foreign keys, using PostgreSQL's "NOT VALID" constraint designation.
+
+For example, consider an application which stores user profiles in a :ref:`reference table <reference_tables>`.
+
+.. code-block:: postgres
+
+   -- we'll store user emails case insensitively, so enable
+   -- the "citext" data type on the coordinator and workers
+   CREATE EXTENSION citext;
```
sumedhpathak: Do we need to use citext for this example?

develop/reference_ddl.rst
```diff
@@ -294,6 +294,60 @@ Not-null constraints can be applied to any column (distribution or not) because
...
+
+.. code-block:: postgres
+
+   INSERT INTO users VALUES ('fake');
+
+   /*
+   ERROR:  new row for relation "users_102010" violates
+           check constraint "syntactic_email_102010"
+   DETAIL:  Failing row contains (fake).
+   *
```
sumedhpathak: Dangling *?

develop/reference_ddl.rst
```diff
@@ -294,6 +294,60 @@ Not-null constraints can be applied to any column (distribution or not) because
 
   ALTER TABLE ads ALTER COLUMN image_url SET NOT NULL;
 
+Using NOT VALID Constraints
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+In some situations it can be useful to enforce constraints for new rows, while allowing existing non-conforming rows to remain unchanged. Citus supports this feature for CHECK constraints and foreign keys, using PostgreSQL's "NOT VALID" constraint designation.
```
sumedhpathak: Maybe add a link to https://www.postgresql.org/docs/11/sql-altertable.html somewhere for further exploration of 'not valid'?

develop/reference_ddl.rst
begriffs: No, we could also use text. I just thought it's a nice practice to encourage treating email addresses case-insensitively (citext indexes them that way and everything). CItext is included with postgres as one of those [additional supplied modules](https://www.postgresql.org/docs/11/contrib.html).

But if you think it's too distracting from the core lesson then I could remove it.

develop/reference_ddl.rst
begriffs: Thanks, I'll check for unbalanced comments. I wrapped the output in comments so the sphinx lexer would be able to apply syntax highlighting.

begriffs:
@sumedhpathak I updated with your suggestions.

PR #838, opened 2019-01-03
Note the SSL changes in Citus 8.1

### LINKED ISSUE #827: Coordinator->worker connection needs adjustment for 8.1 upgrade

The new version changes `citus.node_conninfo` from sslmode=prefer to sslmode=require when citus is compiled against a postgres with ssl compiled in.

This impacts new users onboarding to citus as well as existing users upgrading.

Installation section: no change needed for pg10,11 but on pg9.6 the user is required to restart postgres as it can only enable ssl during boot.

Upgrade section: review the situation in linked PR that would cause people problems on upgrades. Reproduce the problem, verify that I can fix it, then update the docs.

https://github.com/citusdata/citus/pull/2503

begriffs:
Fixes #827

### DIFF
--- a/admin_guide/cluster_management.rst
+++ b/admin_guide/cluster_management.rst
@@ -77,6 +77,22 @@ To add a new node to the cluster, you first need to add the DNS name or IP addre
 
 The new node is available for shards of new distributed tables. Existing shards will stay where they are unless redistributed, so adding a new worker may not help performance without further steps.
 
+.. note::
+
+   As of Citus 8.1, workers use encrypted communication by default. A new node running version 8.1 or greater will refuse to talk with other workers who do not have SSL enabled. When adding a node to an unencrypted cluster, you must reconfigure the new node.
+
+   First, from the coordinator node check whether the other workers use SSL:
+
+   .. code-block:: sql
+
+      SELECT run_command_on_workers('show ssl');
+
+   If they do not, then connect to the new node and permit it to communicate over plaintext if necessary:
+
+   .. code-block:: sql
+
+      ALTER SYSTEM SET 'citus.node_conninfo' TO 'sslmode=prefer';
+
 .. _shard_rebalancing:
 
 Rebalance Shards without Downtime
@@ -464,7 +480,7 @@ To set non-sensitive libpq connection parameters to be used for all node connect
   SET citus.node_conninfo =
     'sslrootcert=/path/to/citus.crt sslmode=verify-full';
 
-There is a whitelist of parameters that the GUC accepts. See the :ref:`node_conninfo <node_conninfo>` reference for details.
+There is a whitelist of parameters that the GUC accepts. See the :ref:`node_conninfo <node_conninfo>` reference for details. By default, nodes encrypt communication with SSL (using sslmode=require).
 
 Citus Enterprise Edition includes an extra table used to set sensitive connection credentials. This is fully configurable per host/user. It's easier than managing ``.pgpass`` files through the cluster and additionally supports certificate authentication.
 
--- a/admin_guide/cluster_management.rst
+++ b/admin_guide/cluster_management.rst
@@ -79,7 +79,7 @@ The new node is available for shards of new distributed tables. Existing shards 
 
 .. note::
 
-   As of Citus 8.1, workers use encrypted communication by default. A new node running version 8.1 or greater will refuse to talk with other workers who do not have SSL enabled. When adding a node to an unencrypted cluster, you must reconfigure the new node.
+   As of Citus 8.1, workers use encrypted communication by default. A new node running version 8.1 or greater will refuse to talk with other workers who do not have SSL enabled. When adding a node to a cluster without encrypted communication, you must reconfigure the new node before creating the Citus extension.
 
    First, from the coordinator node check whether the other workers use SSL:
 
@@ -92,6 +92,7 @@ The new node is available for shards of new distributed tables. Existing shards 
    .. code-block:: sql
 
       ALTER SYSTEM SET 'citus.node_conninfo' TO 'sslmode=prefer';
+      SELECT pg_reload_conf();
 
 .. _shard_rebalancing:
 
@@ -480,7 +481,7 @@ To set non-sensitive libpq connection parameters to be used for all node connect
   SET citus.node_conninfo =
     'sslrootcert=/path/to/citus.crt sslmode=verify-full';
 
-There is a whitelist of parameters that the GUC accepts. See the :ref:`node_conninfo <node_conninfo>` reference for details. By default, nodes encrypt communication with SSL (using sslmode=require).
+There is a whitelist of parameters that the GUC accepts. See the :ref:`node_conninfo <node_conninfo>` reference for details. By default, nodes require connections to be encrypted.
 
 Citus Enterprise Edition includes an extra table used to set sensitive connection credentials. This is fully configurable per host/user. It's easier than managing ``.pgpass`` files through the cluster and additionally supports certificate authentication.
 
--- a/admin_guide/cluster_management.rst
+++ b/admin_guide/cluster_management.rst
@@ -481,7 +481,7 @@ To set non-sensitive libpq connection parameters to be used for all node connect
   SET citus.node_conninfo =
     'sslrootcert=/path/to/citus.crt sslmode=verify-full';
 
-There is a whitelist of parameters that the GUC accepts. See the :ref:`node_conninfo <node_conninfo>` reference for details. By default, nodes require connections to be encrypted.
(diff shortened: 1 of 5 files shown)


admin_guide/cluster_management.rst
```diff
@@ -77,6 +77,22 @@ To add a new node to the cluster, you first need to add the DNS name or IP addre
 
 The new node is available for shards of new distributed tables. Existing shards will stay where they are unless redistributed, so adding a new worker may not help performance without further steps.
 
+.. note::
+
+   As of Citus 8.1, workers use encrypted communication by default. A new node running version 8.1 or greater will refuse to talk with other workers who do not have SSL enabled. When adding a node to an unencrypted cluster, you must reconfigure the new node.
```
begriffs: Maybe "unencrypted cluster" is sloppy terminology, but I wanted to keep it short.

admin_guide/upgrading_citus.rst
```diff
@@ -39,6 +39,10 @@ Major and Minor Version Upgrades
 
 Major and minor version upgrades follow the same steps, but be careful: major upgrades can make backward-incompatible changes in the Citus API. It is best to review the Citus `changelog <https://github.com/citusdata/citus/blob/master/CHANGELOG.md>`_ before a major upgrade and look for any changes which may cause problems for your application.
 
+.. note::
+
+   Starting at version 8.1, Citus encrypts inter-node communication by default. SSL settings are configured automatically during an upgrade, with no action needed on your part. Just be careful when adding a new node to a Citus cluster that does not yet use SSL. The :ref:`adding a worker <adding_worker_node>` section covers that situation.
```
begriffs: I didn't talk about how if `sslmode=prefer` at the time of upgrade then Citus won't make the policy more strict. It doesn't seem relevant to the main warning, which is adding a new node to a pre-8.1 cluster.

begriffs: Another thing to think about is when someone upgrades all nodes of an 8.0 cluster to 8.1, then during the process the upgraded workers will be unable to talk to the un-upgraded ones. There will be a weird situation temporarily. Is this something we should cover, or is it as big of a problem as I am imagining?

admin_guide/cluster_management.rst
```diff
@@ -77,6 +77,22 @@ To add a new node to the cluster, you first need to add the DNS name or IP addre
...
+
+   .. code-block:: sql
+
+      SELECT run_command_on_workers('show ssl');
+
+   If they do not, then connect to the new node and permit it to communicate over plaintext if necessary:
+
+   .. code-block:: sql
+
+      ALTER SYSTEM SET 'citus.node_conninfo' TO 'sslmode=prefer';
```
thanodnl: a second command needs to be run before the setting is changed:

```sql
SELECT pg_reload_conf();
```

admin_guide/cluster_management.rst
thanodnl: yeah, first time I read over this, but now it started nagging that unencrypted cluster is easily interpreted as encryption at rest. How about;

`When adding a node to a cluster without encrypted communication, you must reconfigure the new node before creating the citus extension.`

The last part I added is an important part that is missing currently.

admin_guide/cluster_management.rst
```diff
@@ -464,7 +480,7 @@ To set non-sensitive libpq connection parameters to be used for all node connect
   SET citus.node_conninfo =
     'sslrootcert=/path/to/citus.crt sslmode=verify-full';
 
-There is a whitelist of parameters that the GUC accepts. See the :ref:`node_conninfo <node_conninfo>` reference for details.
+There is a whitelist of parameters that the GUC accepts. See the :ref:`node_conninfo <node_conninfo>` reference for details. By default, nodes encrypt communication with SSL (using sslmode=require).
```
thanodnl: `sslmode=requrie` doesn't encrypt the traffic on its own, also `ssl` has to be set to `on` at the other end etc.
I think rewording this to sth like;

`By default, nodes require connections to be encrypted`.

admin_guide/upgrading_citus.rst
thanodnl: This statement is incorrect.

During an upgrade of an older cluster to 8.1 it specifically doesn't make changes to ssl. Only on new installations it enables internode encryption.

marcocitus:
Probably worth noting (we missed this during design): When a root CA file exists (typically in `~/.postgresql/root.crt`) then the certificate will be checked against that CA per the blue Note on https://www.postgresql.org/docs/current/libpq-ssl.html

If Citus generated the certificate during `CREATE EXTENSION citus` then that check will fail with `SSL error: certificate verify failed`. Possible solutionare to sign the certificate, turn off ssl, or remove the root certificate.

cloud/security.rst
```diff
@@ -4,7 +4,9 @@ Security
 Connecting with SSL
 ===================
 
-For security Citus Cloud accepts only SSL connections, which is why the URL contains the :code:`?sslmode=require` parameter. To avoid a man-in-the-middle attack, you can also verify that the server certificate is correct. Download the official `Citus Cloud certificate <https://console.citusdata.com/citus.crt>`_ and refer to it in connection string parameters:
+For security Citus Cloud accepts only SSL connections, which is why the URL contains the :code:`?sslmode=require` parameter. Within the cluster, nodes also encrypt their communication with one another using SSL.
```
sumedhpathak: As a note, this was always true in Cloud. It's now also true for on-prem (new) installations.

admin_guide/upgrading_citus.rst
sumedhpathak: Nils, specifically its this statement which is incorrect?: "SSL settings are configured automatically during an upgrade, with no action needed on your part"? 
The complexity only comes in if you add a *new* node to an existing cluster?

admin_guide/cluster_management.rst
```diff
@@ -464,7 +481,7 @@ To set non-sensitive libpq connection parameters to be used for all node connect
   SET citus.node_conninfo =
     'sslrootcert=/path/to/citus.crt sslmode=verify-full';
 
-There is a whitelist of parameters that the GUC accepts. See the :ref:`node_conninfo <node_conninfo>` reference for details.
+There is a whitelist of parameters that the GUC accepts. See the :ref:`node_conninfo <node_conninfo>` reference for details. By default, nodes require connections to be encrypted.
```
sumedhpathak: @begriffs I think we can expand on this. We require it, but more importantly enforce it by defaulting to 'require' for ssl_mode.

admin_guide/upgrading_citus.rst
begriffs: I see what Nils means about an upgrade of an older cluster to 8.1 not making changes to ssl.

For the upgrade path, backend/distributed/citus--8.0-12--8.0-13.sql only calls citus_check_defaults_for_sslmode() when current_setting('ssl') is false. Furthermore, this function runs `ALTER SYSTEM SET citus.node_conninfo TO 'sslmode=prefer'` only if citus.node_conninfo is set to sslmode=require.

Just out of curiosity, under what situations will someone be doing an upgrade where ssl is false and sslmode=require? The name of citus_check_defaults_for_sslmode (and its comments) suggest that the default for citus.node_conninfo is sslmode=require, but in an upgrade that might not be true? The citus.node_conninfo GUC was first defined in v7.5, and in that version and in v8.0 it had a default value of sslmode=prefer. Starting in v8.1 it gets the default value sslmode=require when Citus is compiled with USE_SSL defined.

In any case, it looks like I should remove the statement about citus adjusting SSL during upgrade, but leave the warning link about adding a new 8.1 node to an upgraded cluster because the cluster may not have SSL enabled.

begriffs: Also odd that many `citus--8.0-*` scripts don't exist in v8.0.2 but appear in v8.1.0. I know pg doesn't care what the names of those files are, but it was confusing to see what I'm assuming is 8.1 functionality in an upgrade script named 8.0

cloud/security.rst
begriffs: Thought I'd just mention it here, as long as I was updating sections that talk about SSL.

begriffs:
@marcocitus I added a section about the certificate verify failed error. Should we include a code snippet for signing the certificate in the shell? Do you have such a snippet handy that you can share?

begriffs:
Discussed with Sumedh, and now that the statement in the upgrade section is adjusted we're going to merge this to proceed with the 8.1 docs release.

We can continue to improve the SSL documentation after the release.

admin_guide/upgrading_citus.rst
thanodnl: Sorry for the late reply

> Just out of curiosity, under what situations will someone be doing an upgrade where ssl is false and sslmode=require?

This happens when a user didn't change `citus.node_conninfo` on their deployment. In that case it is not store in any config file that gets loaded which causes postgres/citus to use the default value as hard coded in the application. Since we changed the hardcoded value it will change without the user noticing.

As you correctly noticed we change the value to the old default in this case, this will result in the user not observing a change to this guc when upgrading.

PR #839, opened 2019-01-04
Necessity of granting usage on other schemas

### LINKED ISSUE #765: Note "grant usage" required for non-public schemas

In the Cloud section we talk about adding users and granting them access to tables. We should also add a note that "grant usage on ..." is required for schemas other than public.

http://docs.citusdata.com/en/v7.5/cloud/security.html#users-and-permissions

begriffs:
Fixes #765 

Is this what you had in mind, Colton?


begriffs:
Reviewed with Colton via a chat and got his 🚢

PR #846, opened 2019-01-15
A reassuring message about maintenance windows

### LINKED ISSUE #782: Clarify the maintenance window on Cloud

Reassure people that this isn't some automatically reoccurring event, but just the preferred time to carry out tasks that the user initiates.

## How does this work? (devs)

Happens for
* software upgrades
* node resizes
* ... find what else

begriffs:
Repeat the message in two places where someone might become worried
that the window is some kind of recurring event that uses resources,
rather than a period of time in which they can schedule their own
events manually.

Fixes #782

### DIFF
--- a/cloud/performance.rst
+++ b/cloud/performance.rst
@@ -21,7 +21,7 @@ After you adjust the slider and/or buttons and accept the changes, Citus Cloud b
 
 .. image:: ../images/cloud-maintenance-window.png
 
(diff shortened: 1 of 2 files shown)


begriffs:
@lfittl can I get a quick review on this? Hoping this little change will help avoid confusion about maintenance windows.

begriffs:
Thanks for the explanation @lfittl - I incorporated it. Look good now?

lfittl:
@begriffs Looks good!

PR #850, opened 2019-02-05
Examples of temp table for window function

### LINKED ISSUE #830: reference workarounds section needs revisiting
at http://docs.citusdata.com/en/latest/develop/reference_workarounds.html#temp-tables-the-workaround-of-last-resort

We say

> Suppose we have a table called github_events, distributed by the column user_id. Then the following window function will not work:

But `github_events` table does not have a `user_id` column. We should perhaps change the sample instead.

begriffs:
Fixes #830

### DIFF
--- a/develop/reference_workarounds.rst
+++ b/develop/reference_workarounds.rst
@@ -72,7 +72,7 @@ Temp Tables: the Workaround of Last Resort
 
 There are still a few queries that are :ref:`unsupported <unsupported>` even with the use of push-pull execution via subqueries. One of them is running window functions that partition by a non-distribution column.
 
-Suppose we have a table called :code:`github_events`, distributed by the column :code:`user_id`. Then the following window function will not work:
+In our :ref:`real-time analytics tutorial <real_time_analytics_tutorial>` we created a table called :code:`github_events`, distributed by the column :code:`user_id`. If we want to query it and count the events for a preselected of repos, one way to do it in SQL is with a window function. However, partitioning on a non-distribution column does not work:
 
 .. code-block:: sql
 
@@ -83,7 +83,7 @@ Suppose we have a table called :code:`github_events`, distributed by the column 
     FROM github_events
    WHERE repo_id IN (8514, 15435, 19438, 21692);
 
-There is another trick though. We can pull the relevant information to the coordinator as a temporary table:
+There is a trick, though. We can pull the relevant information to the coordinator as a temporary table:
 
 .. code-block:: sql
 
@@ -102,4 +102,3 @@ There is another trick though. We can pull the relevant information to the coord
     FROM results;
 
 Creating a temporary table on the coordinator is a last resort. It is limited by the disk size and CPU of the node.
-


develop/reference_workarounds.rst
```diff
@@ -72,34 +72,82 @@ Temp Tables: the Workaround of Last Resort
...
+
+  SELECT campaign_id, name, impressions_count,
+         avg(impressions_count) OVER (PARTITION BY company_id, campaign_id)
+    FROM ads;
+
+That example had a nice solution, but sometimes we're not able to add the distribution column to a partition clause. In our :ref:`real-time analytics tutorial <real_time_analytics_tutorial>` we deal with Github events (which are distributed by user id). Suppose we want to query for which type of action happens most frequently per organization:
+
+.. code-block:: sql
+
+  -- cannot partition by org_name
```
begriffs: NOTE: this is not true! Citus is able to run the query below. It must pull the `counted` subquery up to the coordinator where it's fine to run rank() on it.

Is there an alternate way to write this query to make it fail, or another query entirely which Citus is not able to execute without us making an explicit temp table?

develop/reference_workarounds.rst
```diff
@@ -72,34 +72,82 @@ Temp Tables: the Workaround of Last Resort
...
+        SELECT org->>'login' as org_name,
+               payload->>'action' AS action, count(*) AS n
+          FROM github_events
+         WHERE payload->>'action' <> ''
+         GROUP by 1, 2
+      ) AS counted
+    ) AS ranked
+   WHERE rank = 1;
+
+That causes an error. As a workaround, we can extract the results of the inner query to a local temporary table on the coordinator node, and perform the window function there.
```
begriffs: > That causes an error

As I mentioned above, no it doesn't. But I want to find a query that does.

mtuncer:
Hi @begriffs 
Following query is not supported since it has a partition by window clause on a non-distribution column. Docs say say that `github_events` table is distributed along `repo_id` column query needed to be modified. So I added `event_type` column to the query. This is not the best looking query, feel free to use a different column. Make sure that window function's partition by clause uses a non-distribution column like below.

```sql
SELECT event_type, repo_id, org->'id' as org_id, count(*)
  OVER (PARTITION BY event_type) -- event_type is not distribution column
  FROM github_events
 WHERE repo_id IN (8514, 15435, 19438, 21692);
```

This query will result in
```
ERROR:  could not run distributed query because the window function that is used cannot be pushed down
HINT:  Window functions are supported in two ways. Either add an equality filter on the distributed tables' partition column or use the window functions with a PARTITION BY clause containing the distribution column
```


It could be executed by using a temp table.

```sql

 CREATE TEMP TABLE results AS (
  SELECT event_type, repo_id, org->'id' as org_id
    FROM github_events
   WHERE repo_id IN (8514, 15435, 19438, 21692)
);

-- now run the aggregate locally
SELECT event_type, repo_id, org_id, count(*)
  OVER (PARTITION BY event_type)
  FROM results;
```

As I said below, feel free to change the query as you like key point is that the part `OVER (PARTITION BY <column>)` must include a non-distribution column.

jonels-msft:
> Docs say say that github_events table is distributed along repo_id

@mtuncer interesting, I see that in some places in the docs we distribute github_events by repo_id, and in other places (like the realtime tutorial) we do it by user_id.

I provided a link in this workarounds section to the realtime tutorial definition so that the examples work. Perhaps we want to standardize the example across the docs later though...

PR #852, opened 2019-02-07
Fix queries for (non-)qualified names

### LINKED ISSUE #725: Running on all shards example query errors out in default schema table
In this section https://docs.citusdata.com/en/latest/develop/reference_propagation.html#running-on-all-shards
```SQL
-- create a table in default schema
CREATE TABLE test_1(i int PRIMARY KEY);
-- distribute it
SELECT create_distributed_table('test_1', 'i');
-- run the example query in the section
SELECT sum(result::bigint) AS estimated_count
FROM run_command_on_shards(
  'test_1',
  $cmd$
    SELECT reltuples
      FROM pg_class c
      JOIN pg_catalog.pg_namespace n on n.oid=c.relnamespace
     WHERE n.nspname||'.'||relname = '%s';
  $cmd$
);
```
I get the following error:
```SQL
ERROR:  invalid input syntax for integer: ""
```
Apparently the reason is that `%s` doesn't include the schemaname, and so there is no valid input to sum. Removing `n.nspname` from the `WHERE` clause works fine.
```SQL
-- run the example query in the section after removing n.nspname
SELECT sum(result::bigint) AS estimated_count
FROM run_command_on_shards(
  'test_1',
  $cmd$
    SELECT reltuples
      FROM pg_class c
      JOIN pg_catalog.pg_namespace n on n.oid=c.relnamespace
     WHERE relname = '%s';
  $cmd$
);
```
with the result:
```SQL
estimated_count 
-----------------
               0
(1 row)
```
On the other hand, the original query works fine with a specific schema. For example:
```SQL
-- create a new schema
CREATE SCHEMA try;
-- create a table in new schema
CREATE TABLE try.test_1(i int PRIMARY KEY);
-- distribute it
SELECT create_distributed_table('try.test_1', 'i');
-- run the original query in the section
SELECT sum(result::bigint) AS estimated_count
FROM run_command_on_shards(
  'try.test_1',
  $cmd$
    SELECT reltuples
      FROM pg_class c
      JOIN pg_catalog.pg_namespace n on n.oid=c.relnamespace
     WHERE n.nspname||'.'||relname = '%s';
  $cmd$
);
```
with the result:
```SQL
estimated_count 
-----------------
               0
(1 row)
```
Obviously, the query with the removed part errors out in the latter case.

begriffs:
Run_command_on_shards() removes 'public.' from its %s substitution,
but allows other schema names to remain. We convert to regclass to
resolve names to the proper relations before doing a comparison.

Fixes #725

### DIFF
--- a/develop/reference_propagation.rst
+++ b/develop/reference_propagation.rst
@@ -40,15 +40,17 @@ The :code:`run_command_on_shards` function applies a SQL command to each shard, 
   -- Get the estimated row count for a distributed table by summing the
   -- estimated counts of rows for each shard.
   SELECT sum(result::bigint) AS estimated_count
-  FROM run_command_on_shards(
-    'my_distributed_table',
-    $cmd$
-      SELECT reltuples
-        FROM pg_class c
-        JOIN pg_catalog.pg_namespace n on n.oid=c.relnamespace
-       WHERE n.nspname||'.'||relname = '%s';
-    $cmd$
-  );
+    FROM run_command_on_shards(
+      'my_distributed_table',
+      $cmd$
+        SELECT reltuples
+          FROM pg_class c
+          JOIN pg_catalog.pg_namespace n on n.oid=c.relnamespace
+         WHERE (n.nspname || '.' || relname)::regclass = '%s'::regclass
+           AND n.nspname NOT IN ('citus', 'pg_toast')
+      $cmd$
+    );
+
 
 Running on all Placements
 -------------------------
--- a/develop/reference_propagation.rst
+++ b/develop/reference_propagation.rst
@@ -47,7 +47,7 @@ The :code:`run_command_on_shards` function applies a SQL command to each shard, 
           FROM pg_class c
           JOIN pg_catalog.pg_namespace n on n.oid=c.relnamespace
          WHERE (n.nspname || '.' || relname)::regclass = '%s'::regclass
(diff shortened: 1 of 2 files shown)


develop/reference_propagation.rst
```diff
@@ -40,15 +40,17 @@ The :code:`run_command_on_shards` function applies a SQL command to each shard,
...
-    $cmd$
-  );
+    FROM run_command_on_shards(
+      'my_distributed_table',
+      $cmd$
+        SELECT reltuples
+          FROM pg_class c
+          JOIN pg_catalog.pg_namespace n on n.oid=c.relnamespace
+         WHERE (n.nspname || '.' || relname)::regclass = '%s'::regclass
+           AND n.nspname NOT IN ('citus', 'pg_toast')
```
begriffs: Ran into a permissions error on Cloud when attempting to resolve names using `::regclass` from either of these schemas. Thus I just filter them out, which seems like a bit of a hack.

begriffs:
Lukas, thought I'd assign you as reviewer in case you know a trick to run this query on Cloud without needing to blacklist specific schemas.

lfittl:
@begriffs I think thats because those internal schemas are owned by the postgres user, not the citus user. It seems reasonable to ignore them. You could also consider including `pg_catalog` in the list.

PR #854, opened 2019-03-18
change the django documentation to introduce mixins + details on fkey

louiseGrandjonc:


louiseGrandjonc:
This page is missing `TenantOneToOneField`

louiseGrandjonc:
Also how to handle `ManyToMany` unique constraints

develop/migration_mt_django.rst
```diff
@@ -5,76 +5,185 @@
...
+
+
+  class Task(models.Model):
+      name = models.CharField(max_length=255)
+      project = models.ForeignKey(Project, on_delete=models.CASCADE,
+                                  related_name='tasks')
+
+
+
+The tricky thing with this pattern is that in order to find all tasks for an account, you'll have to query for all of a account's project first. This becomes a problem once you start sharding data, and in particular when you run UPDATE or DELETE queries on nested models like task in this example.
```
jonels-msft: in order to find all tasks for an account, you'll have to query for all of a*n* account's project*s* first

develop/migration_mt_django.rst
```diff
@@ -5,76 +5,185 @@
...
+**1.1 Introducting the column to models belonging to an account**
+
+In order to scale out a multi-tenant model, it's essential that you can locate all records that belong to an account quickly.
+The reason is that we want that an ORM call like
+
+.. code-block:: python
+
+  Project.objects.filter(account_id=1).prefetch_related('tasks')
+
+Generating the following queries
```
jonels-msft: I think "we want … to generate" sounds more natural than "we want … generating"

jonels-msft: The before and after SQL kind of confused me the way it was phrased. What about this.

```rst
In order to scale out a multi-tenant model, it’s essential for queries to quickly
locate all records that belong to an account. Consider an ORM call such as:

.. code-block:: python

  Project.objects.filter(account_id=1).prefetch_related('tasks')

It generates these underlying SQL queries:

.. code-block:: sql

  SELECT *
    FROM myapp_project
   WHERE account_id = 1;
  
  SELECT *
    FROM myapp_task
   WHERE project_id IN (1, 2, 3);

However the second query would go faster with an extra filter:

.. code-block:: diff

  + -- the AND clause identifies the tenant

    SELECT *
      FROM myapp_task
     WHERE project_id IN (1, 2, 3)
  +    AND account_id = 1;
```

develop/migration_mt_django.rst
```diff
@@ -5,76 +5,185 @@
...
-    name = models.CharField(max_length=255)
-    url = models.URLField()
+  class Country(models.Model):
+      name = models.CharField(max_length=255)
+
+  class Account(models.Model):
+      name = models.CharField(max_length=255)
+      domain = models.CharField(max_length=255)
+      subdomain = models.CharField(max_length=255)
+      country = models.ForeignKey(Country, on_delete=models.CASCADE)
```
jonels-msft: This isn't really related to what you're documenting, but... I wonder whether "restrict" would be a safer choice here than "cascade," since otherwise accidentally deleting a country could blow away a bunch of accounts.

develop/migration_mt_django.rst
```diff
@@ -5,76 +5,185 @@
...
+
+  Project.objects.filter(account_id=1).prefetch_related('managers')
+
+
+Can include in their :code:`WHERE` clause the :code:`account_id` like this:
+
+.. code-block:: postgresql
+
+  SELECT *
+  FROM "myapp_project" WHERE "myapp_project"."account_id" = 1;
```
jonels-msft: I'll leave it up to you whether you prefer the query style generated by the ORM such as

```sql
SELECT *
FROM "myapp_project" WHERE "myapp_project"."account_id" = 1;
```

vs a more friendly style like

```sql
SELECT *
  FROM myapp_project
 WHERE account_id = 1;
```

The former has the advantage that it matches what's actually happening, but the latter I find easier to understand.

develop/migration_mt_django.rst
```diff
@@ -5,76 +5,185 @@
...
 
-The tricky thing with this pattern is that in order to find all purchases for a store, you'll have to query for all of a store's products first. This becomes a problem once you start sharding data, and in particular when you run UPDATE or DELETE queries on nested models like purchases in this example.
+  SELECT *
+  FROM "myapp_manager"
+  INNER JOIN "myapp_projectmanager"
+  ON ("myapp_manager"."id" = "myapp_projectmanager"."manager_id"
+  AND  myapp_projectmanager"."account_id" = "myapp_manager"."account_id")
```
jonels-msft: Missing double quote?

Which makes me think, maybe you want to run these commands in a db just to double check them?

develop/migration_mt_django.rst
```diff
@@ -88,84 +197,322 @@ Django automatically creates a simple "id" primary key on models, so we will nee
...
+Some ``UNIQUE`` constraints are created by the ORM and you will need to explicitily drop them.
+This is the case for ``OneToOneField`` and ``ManyToMany`` fields.
+
+
+For these cases you will need to:
+1. Find the constraints
+2. Do a migration to drop them
+3. Re-create constraints including the account\_id field
+
+To find the constraints, connect to your database and run ``\d+ myapp_projectmanager``
```
jonels-msft: Might want to note that we're asking them to connect with psql in particular. (Django has its own command line db console.)

develop/migration_mt_django.rst
```diff
@@ -88,84 +197,322 @@ Django automatically creates a simple "id" primary key on models, so we will nee
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Next, we'll use the `django-multitenant <https://github.com/citusdata/django-multitenant>`_ library to add account_id to foreign keys, and make application queries easier later on.
 
 In requirements.txt for your Django application, add
 
 ::
 
-  django_multitenant>=1.1.0
+  django_multitenant>=2.0.0
```
jonels-msft: Perhaps for safety in case of a future breaking change, could do

```
django_multitenant >=2, <3
```

develop/migration_mt_django.rst
```diff
@@ -88,84 +197,322 @@ Django automatically creates a simple "id" primary key on models, so we will nee
...
 
 .. code-block:: python
 
-  from django_multitenant.models import *
-  from django_multitenant.fields import *
+  from django_multitenant.mixins import *
+
 
-Change all the models to inherit from ``TenantModel`` rather than ``Model``, set the tenant\_id on each model, and use ``TenantForeignKey`` rather than ``ForeignKey`` for any foreign key which does not already contain the tenant\_id.
+Then your models should be changed to have the new manager and inheritance. As we are using Mixins, your models and managers can also inherits from other mixins you might have been using previously. It is for example compatible with using :code:`django.contrib.gis.db` models.
```
jonels-msft: This paragraph was kind of confusing for me. How about something like

```rst
Previously our example models inherited from just models.Model, but now we need
to change them to also inherit from TenantModelMixin. The models in real
projects may inherit from other mixins too like ``django.contrib.gis.db``,
which is fine.

You will also, at this point, introduce the tenant_id to define which column is
the distribution column.
```

develop/migration_mt_django.rst
```diff
@@ -88,84 +197,322 @@ Django automatically creates a simple "id" primary key on models, so we will nee
...
+
+You will also, at this point introduce the tenant_id, to define which column is the distribution column.
 
 .. code-block:: python
 
-  class Store(TenantModel):
-    name = models.CharField(max_length=255)
-    url = models.URLField()
-    tenant_id = "id"
+  class TenantManager(TenantManagerMixin, models.Manager):
```
jonels-msft: What does TenantManager do? Is this what you meant above when you said, "your models should be changed to have the new manager" ?

develop/migration_mt_django.rst
```diff
@@ -178,47 +525,45 @@ We need one final migration to tell Citus to mark tables for distribution. Creat
...
+        "SELECT create_distributed_table('myapp_projectmanager','account_id')"
+      ),
+      migrations.RunSQL(
+        "SELECT create_distributed_table('myapp_task','account_id')"
       ),
-    ]
-
-With all the migrations created from the steps so far, apply them to the database with ``./manage.py migrate``.
-
-There's one more little detail. Server-side cursors do not work well with Citus. Go back to the database configuration in `settings.py` and include the following parameter:
```
jonels-msft: Does Citus now support cursors? We added this note about DISABLE_SERVER_SIDE_CURSORS because someone opened a github issue about it.

develop/migration_mt_django.rst
```diff
@@ -178,47 +525,45 @@ We need one final migration to tell Citus to mark tables for distribution. Creat
...
 The django-multitenant library discussed in the previous section is not only useful for migrations, but for simplifying application queries. The library allows application code to easily scope queries to a single tenant. It automatically adds the correct SQL filters to all statements, including fetching objects through relations.
 
 For instance, in a controller simply ``set_current_tenant`` and all the queries or joins afterward will include a filter to scope results to a single tenant.
 
 .. code-block:: python
 
-  # set the current tenant to the first store
-  s = Store.objects.all()[0]
+  # set the current tenant to the first account
+  s = Account.objects.first()
```
jonels-msft: Nice. :)

develop/migration_mt_django.rst
louiseGrandjonc: I actually copy-pasted some ORM queries, but I simplified the parenthesis, so I must have made a mistake in this one. I think that I will remove the quotes, it will be more readable

develop/migration_mt_django.rst
louiseGrandjonc: The `TenantManager` uses the `TenantManagerMixin` it's basically what handles the changes in the ORM calls (like adding filters to joins and all). It's necessary for the models to do the right queries, and also it means that if they have their own manager, let's say they have 

```
class ProjectManager(models.Manager):
    def get_queryset(self):
        # do something
```

They can change it to be   


```
class ProjectManager(TenantManagerMixin, models.Manager):
    def get_queryset(self):
        # do something
```

develop/migration_mt_django.rst
louiseGrandjonc: Ah, not sure, I deleted this part because in my tests (in the unit tests of the django-multitenant library) I didn't have to change that.

PR #856, opened 2019-03-29
Update Citus version to 8.2

velioglu:


jonels-msft:
Thanks!

I'll have to change the versions in other places in the docs too prior to merging, and run through my upgrade testing checklist.

jonels-msft:
Merged the other PR to do the version bump

PR #858, opened 2019-04-04
single node psql connection option for tutorials

### LINKED ISSUE #855: Tutorial issue
After running the installation tutorial:

http://docs.citusdata.com/en/v8.1/installation/single_machine_debian.html

You move on to the next step:

http://docs.citusdata.com/en/v8.1/get_started/tutorial_multi_tenant.html#multi-tenant-tutorial

But here things get a bit unclear. the tutorial talks about Citus Cloud, Docker but does not mention the basic psql client.

Now, you may fall in the trap that you run: 

```
psql
```

And start with following the steps of creating the tables, index and ...

```
SELECT create_distributed_table('companies', 'id');
```

Boem ...

```
postgres=# SELECT create_distributed_table('companies', 'id');
ERROR:  function create_distributed_table(unknown, unknown) does not exist
LINE 1: SELECT create_distributed_table('companies', 'id');
```

You run into this issue and start wondering what is wrong. A quick \dx shows you that the Citus extension is not loaded but you are still confused as to why ...

To make a story short, the correct command is ofcourse:

```
psql -p 9700
```

As you need to connect to the citus postgresql server and not the default... This is confusing and very error sensitive as its totally not written down. So the documentation needs a update on:

http://docs.citusdata.com/en/v8.1/get_started/tutorial_multi_tenant.html#multi-tenant-tutorial

jonels-msft:
Fixes #855 

@Indribell thanks for reporting how this section can be confusing. Does my change in this PR clear it up?

### DIFF
--- a/get_started/tutorial_multi_tenant.rst
+++ b/get_started/tutorial_multi_tenant.rst
@@ -40,9 +40,15 @@ Creating tables
 
 To start, you can first connect to the Citus coordinator using psql.
 
+**If you are using native Postgres**, as installed in our :ref:`development` guide, the coordinator node will be running on port 9700.
+
+.. code-block:: bash
+
+   psql -p 9700
+
 **If you are using Docker**, you can connect by running psql with the docker exec command:
 
-::
+.. code-block:: bash
 
     docker exec -it citus_master psql -U postgres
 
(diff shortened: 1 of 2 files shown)


jonels-msft:
Look good, @Indribell ?

PR #859, opened 2019-04-04
Bump version to 8.2

jonels-msft:
@sumedhpathak I started with the version upgrade. Have not tested any of these changes yet.

Will we be updating the cloudformation template?

I'll also open a PR for the download links on our main site.


jonels-msft:
Closes #856

admin_guide/upgrading_citus.rst
```diff
@@ -53,15 +53,15 @@ Step 1. Update Citus Package
 .. code-block:: bash
 
   sudo apt-get update
-  sudo apt-get install postgresql-10-citus-8.1
+  sudo apt-get install postgresql-10-citus-8.2
```
sumedhpathak: We should update this to PG11? Seems like I missed this during the 8.1 changes, but these instructions for Citus version upgrades should assume the customer is on PG11.

admin_guide/upgrading_citus.rst
```diff
@@ -53,15 +53,15 @@ Step 1. Update Citus Package
...
   sudo service postgresql restart
 
 **Fedora, CentOS, or Red Hat**
 
 .. code-block:: bash
 
   # Fedora, CentOS, or Red Hat
-  sudo yum swap citus75_10 citus81_10
+  sudo yum swap citus75_10 citus82_10
```
sumedhpathak: Same as above.

admin_guide/upgrading_citus.rst
jonels-msft: This is part of an example upgrading from 7.5 to 8.2. Since 7.x isn't compatible with pg 11, we don't attempt to upgrade both the Citus and pg versions at the same time.

We have a section later on for "Upgrading PostgreSQL version from 10 to 11" but I don't see where it tells people to install pg 11 in there, it kind of assumes that is already installed. Might be worthwhile for me to run through these instructions and actually try to upgrade from 7.5/pg10 -> 8.2/pg10 -> 8.2/pg11 and update whatever is missing.

hanefi:
I found one important change in 8.2 that breaks one of our queries in the Real-Time Dashboard Use-Case.

We will need to change the loop that simulates the data ingestion

hanefi:
I suggest we insert a `COMMIT;` line in the loop after the `INSERT`

https://github.com/citusdata/citus_docs/blob/1ce6cfa2d64a548c4bc7edc48c37e525bdec8bc2//use_cases/realtime_analytics.rst#L72-L92

admin_guide/upgrading_citus.rst
jonels-msft: @sumedhpathak I updated the section to use pg11, and added a note about major upgrades. Could you take a look and see if that looks clear?

jonels-msft:
@hanefi I'm actually glad that loop no longer works. It used to be a transaction isolation violation (allowing dirty reads) and I felt weird showcasing it.

I'll test it on pg 11 next week to verify that adding the commit in the loop makes it work.

jonels-msft:
@hanefi I confirmed that the inserted rows were not visible to another transaction with pg 11 and Citus 8.2, and also that adding a commit statement in the loop fixes it.

@sumedhpathak this PR is ready for final review. If you would also like to merge it and practice doing the version upgrade I can walk you through that.

PR #864, opened 2019-05-21
Add Azure to managed deployments

jonels-msft:
I updated several pages that talked about Cloud to have them also mention Hyperscale.

One area I didn't change are the "portal" pages like https://docs.citusdata.com/en/v8.2/portals/citus_cloud.html

That would be take a little more work, but I can do it if desired.


craigkerstiens:
@jonels-msft typo on getting_started. Otherwise looks good to :shipit:  and then we should come back and try to make a pass on the portal pages as well.

PR #876, opened 2019-09-04
Updates for the adaptive executor

jonels-msft:
The new executor affects many areas of the docs. This is a first pass, and I've added questions below.

### DIFF
--- a/admin_guide/cluster_management.rst
+++ b/admin_guide/cluster_management.rst
@@ -354,37 +354,53 @@ Results:
 
 ::
 
-  ┌────────────┬────────┬───────┬───────────────────────────────────────────────┬───────────────┬───────────────┬───────┐
-  │  queryid   │ userid │ dbid  │                     query                     │   executor    │ partition_key │ calls │
-  ├────────────┼────────┼───────┼───────────────────────────────────────────────┼───────────────┼───────────────┼───────┤
-  │ 1496051219 │  16384 │ 16385 │ select count(*) from foo;                     │ real-time     │ NULL          │     1 │
-  │ 2530480378 │  16384 │ 16385 │ select * from foo where id = $1               │ router        │ 42            │     1 │
-  │ 3233520930 │  16384 │ 16385 │ insert into foo select generate_series($1,$2) │ insert-select │ NULL          │     1 │
-  └────────────┴────────┴───────┴───────────────────────────────────────────────┴───────────────┴───────────────┴───────┘
+  -[ RECORD 1 ]-+----------------------------------------------
+  queryid       | -6844578505338488014
+  userid        | 10
+  dbid          | 13340
+  query         | SELECT count(*) FROM foo;
+  executor      | adaptive
+  partition_key |
+  calls         | 1
+  -[ RECORD 2 ]-+----------------------------------------------
+  queryid       | 185453597994293667
+  userid        | 10
+  dbid          | 13340
+  query         | insert into foo select generate_series($1,$2)
+  executor      | insert-select
+  partition_key |
+  calls         | 1
+  -[ RECORD 3 ]-+----------------------------------------------
+  queryid       | 1301170733886649828
+  userid        | 10
+  dbid          | 13340
+  query         | SELECT * FROM foo WHERE id = $1
+  executor      | adaptive
+  partition_key | 42
+  calls         | 1
 
-We can see that Citus used the real-time executor to run the count query. This executor fragments the query into constituent queries to run on each node and combines the results on the coordinator node. In the case of the second query (filtering by the distribution column ``id = $1``), Citus determined that it needed the data from just one node. Thus Citus chose the router executor to send the whole query to that node for execution. Lastly, we can see that the ``insert into foo select…`` statement ran with the insert-select executor which provides flexibility to run these kind of queries.  The insert-select executor must sometimes pull results to the coordinator and push them back down to workers.
+We can see that Citus uses the adaptive executor most commonly to run queries. This executor fragments the query into constituent queries to run on relevant nodes, and combines the results on the coordinator node. In the case of the second query (filtering by the distribution column ``id = $1``), Citus determined that it needed the data from just one node. Lastly, we can see that the ``insert into foo select…`` statement ran with the insert-select executor which provides flexibility to run these kind of queries.  The insert-select executor must sometimes pull results to the coordinator and push them back down to workers.
 
-So far the information in this view doesn't give us anything we couldn't already learn by running the ``EXPLAIN`` command for a given query. However in addition to getting information about individual queries, the ``citus_stat_statements`` view allows us to answer questions such as "what percentage of queries in the cluster are run by which executors?"
+So far the information in this view doesn't give us anything we couldn't already learn by running the ``EXPLAIN`` command for a given query. However in addition to getting information about individual queries, the ``citus_stat_statements`` view allows us to answer questions such as "what percentage of queries in the cluster are scoped to a single tenant?"
 
 .. code-block:: postgresql
 
-  SELECT executor, sum(calls)
+  SELECT sum(calls),
+         partition_key IS NOT NULL AS single_tenant
   FROM citus_stat_statements
-  GROUP BY executor;
+  GROUP BY 2;
 
 ::
 
-  ┌───────────────┬─────┐
-  │   executor    │ sum │
-  ├───────────────┼─────┤
-  │ insert-select │   1 │
-  │ real-time     │   1 │
-  │ router        │   1 │
-  └───────────────┴─────┘
+  .
+   sum | single_tenant
+  -----+---------------
+     2 | f
+     1 | t
 
-In a multi-tenant database, for instance, we would expect to see "router" for the vast majority of queries, because the queries should route to individual tenants. Seeing too many real-time queries may indicate that queries do not have the proper filters to match a tenant, and are using unnecessary resources.
+In a multi-tenant database, for instance, we would expect the vast majority of queries to be single tenant. Seeing too many multi-tenant queries may indicate that queries do not have the proper filters to match a tenant, and are using unnecessary resources.
 
-We can also find which partition_ids are the most frequent targets of router execution. In a multi-tenant application these would be the busiest tenants.
+We can also find which partition_ids are the most frequent targets. In a multi-tenant application these would be the busiest tenants.
 
 .. code-block:: sql
 
@@ -402,26 +418,6 @@ We can also find which partition_ids are the most frequent targets of router exe
   ├───────────────┼───────────────┤
   │ 42            │             1 │
   └───────────────┴───────────────┘
-
-
-Finally, by joining ``citus_stat_statements`` with ``pg_stat_statements`` we can gain a better view into not only how many queries use different executor types, but how much time each executor spends running queries.
-
-.. code-block:: sql
-
-  SELECT executor, sum(css.calls), sum(pss.total_time)
-  FROM citus_stat_statements css
-  JOIN pg_stat_statements pss USING (queryid)
-  GROUP BY executor;
-
-::
-
-  ┌───────────────┬─────┬───────────┐
-  │   executor    │ sum │    sum    │
-  ├───────────────┼─────┼───────────┤
-  │ insert-select │   1 │ 18.393312 │
-  │ real-time     │   1 │  3.537155 │
-  │ router        │   1 │  0.392590 │
-  └───────────────┴─────┴───────────┘
 
 Statistics Expiration
 ---------------------
--- a/admin_guide/cluster_management.rst
+++ b/admin_guide/cluster_management.rst
@@ -458,9 +458,16 @@ Long running queries can hold locks, queue up WAL, or just consume a lot of syst
 
 The timeout is specified in milliseconds.
 
-.. note::
+To customize the timeout per query, use ``SET LOCAL`` in a transaction:
 
-   It's not possible to make an exception to the timeout for individual queries. Citus does not yet propagate ``SET LOCAL`` to the workers, so the statement_timeout GUC cannot be adjusted within a session.
+.. code-block:: postgres
+
+   BEGIN;
+   -- this limit applies to just the current transaction
+   SET LOCAL statement_timeout TO 300000;
+
+   -- ...
+   COMMIT;
 
 Security
 ========
--- a/admin_guide/cluster_management.rst
+++ b/admin_guide/cluster_management.rst
@@ -379,7 +379,7 @@ Results:
   partition_key | 42
   calls         | 1
 
-We can see that Citus uses the adaptive executor most commonly to run queries. This executor fragments the query into constituent queries to run on relevant nodes, and combines the results on the coordinator node. In the case of the second query (filtering by the distribution column ``id = $1``), Citus determined that it needed the data from just one node. Lastly, we can see that the ``insert into foo select…`` statement ran with the insert-select executor which provides flexibility to run these kind of queries.  The insert-select executor must sometimes pull results to the coordinator and push them back down to workers.
+We can see that Citus uses the adaptive executor most commonly to run queries. This executor fragments the query into constituent queries to run on relevant nodes, and combines the results on the coordinator node. In the case of the second query (filtering by the distribution column ``id = $1``), Citus determined that it needed the data from just one node. Lastly, we can see that the ``insert into foo select…`` statement ran with the insert-select executor which provides flexibility to run these kind of queries.
 
 So far the information in this view doesn't give us anything we couldn't already learn by running the ``EXPLAIN`` command for a given query. However in addition to getting information about individual queries, the ``citus_stat_statements`` view allows us to answer questions such as "what percentage of queries in the cluster are scoped to a single tenant?"
 
--- a/develop/api_guc.rst
+++ b/develop/api_guc.rst
@@ -380,7 +380,7 @@ manually. In the future the adaptive executor may adjust this GUC on the fly,
 shrinking the pool when workers are busy, and growing it when they are idle.
 
 This is a session-level GUC, and thus can be used to limit the resource usage
-for individual sessions. Lowering the the value can limit the queries' use of
+for individual sessions. Lowering the value can limit the queries' use of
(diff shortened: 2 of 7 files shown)


develop/api_guc.rst
```diff
@@ -395,6 +450,31 @@ citus.partition_buffer_size (integer)
 
 Sets the buffer size to use for partition operations and defaults to 8MB. Citus allows for table data to be re-partitioned into multiple files when two large tables are being joined. After this partition buffer fills up, the repartitioned data is flushed into files on disk. This configuration entry can be set at run-time and is effective on the workers.
 
+Real-time executor configuration (deprecated)
```
jonels-msft: Is there still a use for the real-time executor? Assuming that if people want that behavior they can use the adaptive executor with `force_max_query_parallelization`

develop/reference_processing.rst
```diff
@@ -34,31 +34,18 @@ Distributed Query Executor
...
 
-Citus has three basic executor types: real time, router, and task tracker. It chooses which to use dynamically, depending on the structure of each query, and can use more than one at once for a single query, assigning different executors to different subqueries/CTEs as needed to support the SQL functionality. This process is recursive: if Citus cannot determine how to run a subquery then it examines sub-subqueries.
+Citus has three executor types: adaptive, push-pull, and task tracker. It chooses which to use dynamically, depending on the structure of each query, and can use more than one at once for a single query, assigning different executors to different subqueries/CTEs as needed to support the SQL functionality. This process is recursive: if Citus cannot determine how to run a subquery then it examines sub-subqueries.
```
jonels-msft: Would appreciate suggestions for this section. I cut out the real-time and router executors since they appear to be special cases of how the adaptive one works now.

develop/reference_workarounds.rst
```diff
@@ -5,18 +5,18 @@ SQL Support and Workarounds
 
 As Citus provides distributed functionality by extending PostgreSQL, it is compatible with PostgreSQL constructs. This means that users can use the tools and features that come with the rich and extensible PostgreSQL ecosystem for distributed tables created with Citus.
 
-Citus has 100% SQL coverage for any queries it is able to execute on a single worker node. These kind of queries are called :ref:`router executable <router_executor>` and are common in :ref:`mt_use_case` when accessing information about a single tenant.
+Citus has 100% SQL coverage for any queries it is able to execute on a single worker node. These kind of queries are common in :ref:`mt_use_case` when accessing information about a single tenant.
```
jonels-msft: Calling these "single-shard" rather than router executable. Is this the terminology you'd advise?

performance/performance_tuning.rst
```diff
@@ -44,10 +44,10 @@ Here is an example of explaining the plan for a particular example query.
 ::
 
   Sort  (cost=0.00..0.00 rows=0 width=0)
-    Sort Key: minute
+    Sort Key: remote_scan.minute
```
jonels-msft: These are fairly mechanical changes, but I'd like your suggestions about what broader advice to give. The "Advanced" section is especially lacking.

admin_guide/cluster_management.rst
```diff
@@ -354,37 +354,53 @@ Results:
...
+  partition_key | 42
+  calls         | 1
+
+We can see that Citus uses the adaptive executor most commonly to run queries. This executor fragments the query into constituent queries to run on relevant nodes, and combines the results on the coordinator node. In the case of the second query (filtering by the distribution column ``id = $1``), Citus determined that it needed the data from just one node. Lastly, we can see that the ``insert into foo select…`` statement ran with the insert-select executor which provides flexibility to run these kind of queries.  The insert-select executor must sometimes pull results to the coordinator and push them back down to workers.
+
+So far the information in this view doesn't give us anything we couldn't already learn by running the ``EXPLAIN`` command for a given query. However in addition to getting information about individual queries, the ``citus_stat_statements`` view allows us to answer questions such as "what percentage of queries in the cluster are scoped to a single tenant?"
 
 .. code-block:: postgresql
 
-  SELECT executor, sum(calls)
```
jonels-msft: The type of executor no longer seems particularly interesting. What people want to know is whether the query is single-sharded.

admin_guide/cluster_management.rst
```diff
@@ -403,26 +419,6 @@ We can also find which partition_ids are the most frequent targets of router exe
   │ 42            │             1 │
   └───────────────┴───────────────┘
 
-
-Finally, by joining ``citus_stat_statements`` with ``pg_stat_statements`` we can gain a better view into not only how many queries use different executor types, but how much time each executor spends running queries.
```
jonels-msft: I just cut this part because most queries will be adaptive.

admin_guide/cluster_management.rst
```diff
@@ -458,9 +454,16 @@ Long running queries can hold locks, queue up WAL, or just consume a lot of syst
 
 The timeout is specified in milliseconds.
 
-.. note::
+To customize the timeout per query, use ``SET LOCAL`` in a transaction:
```
jonels-msft: Woo hoo!

develop/api_guc.rst
```diff
@@ -336,30 +339,82 @@ Note that it may be useful to use :code:`error` during development testing, and
...
+Each backend opens connections to the workers to query the shards. At the end
+of the transaction, the configured number of connections is kept open to speed
+up subsequent commands.  Increasing this value will reduce the latency of
+multi-shard queries, but will also increase overhead on the workers.
+
+The default value is 1. A larger value such as 2 might be helpful for clusters
+that use a small number of concurrent sessions, but it's not wise to go much
+further (e.g. 16 would be too high).
+
+citus.propagate_set_commands (enum)
```
onderkalaci: This is generic across executors, not directly related to adaptive executor

admin_guide/cluster_management.rst
```diff
@@ -354,37 +354,53 @@ Results:
...
+  -[ RECORD 3 ]-+----------------------------------------------
+  queryid       | 1301170733886649828
+  userid        | 10
+  dbid          | 13340
+  query         | SELECT * FROM foo WHERE id = $1
+  executor      | adaptive
+  partition_key | 42
+  calls         | 1
+
+We can see that Citus uses the adaptive executor most commonly to run queries. This executor fragments the query into constituent queries to run on relevant nodes, and combines the results on the coordinator node. In the case of the second query (filtering by the distribution column ``id = $1``), Citus determined that it needed the data from just one node. Lastly, we can see that the ``insert into foo select…`` statement ran with the insert-select executor which provides flexibility to run these kind of queries.  The insert-select executor must sometimes pull results to the coordinator and push them back down to workers.
```
onderkalaci: > The insert-select executor must sometimes pull results to the coordinator and push them back down to workers.

in fact, the insert-select executor **always** pulls the results to the coordinator. The `INSERT ..SELECT` pushdown goes through the adaptive executor:

```SQL
EXPLAIN INSERT INTO test  SELECT * FROM test;
┌─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
│                                                       QUERY PLAN                                                        │
├─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ Custom Scan (Citus Adaptive)  (cost=0.00..0.00 rows=0 width=0)                                                          │
│   Task Count: 32                                                                                                        │
│   Tasks Shown: One of 32                                                                                                │
│   ->  Task                                                                                                              │
│         Node: host=localhost port=9700 dbname=postgres                                                                  │
│         ->  Insert on test_102105 citus_table_alias  (cost=0.00..61.00 rows=13 width=4)                                 │
│               ->  Seq Scan on test_102105 test  (cost=0.00..61.00 rows=13 width=4)                                      │
│                     Filter: ((worker_hash(a) >= '-2147483648'::integer) AND (worker_hash(a) <= '-2013265921'::integer)) │
└─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
(8 rows)
```

The same query, but added `LIMIT 10`, which requires coordinator `INSERT .. SELECT` because the results needs to be `merge`d on the coordinator:
```SQL
 EXPLAIN INSERT INTO test  SELECT * FROM test LIMIT 10;
┌──────────────────────────────────────────────────────────────────────────────────────────────────┐
│                                            QUERY PLAN                                            │
├──────────────────────────────────────────────────────────────────────────────────────────────────┤
│ Custom Scan (Citus INSERT ... SELECT via coordinator)  (cost=0.00..0.00 rows=0 width=0)          │
│   ->  Limit  (cost=0.00..0.00 rows=0 width=0)                                                    │
│         ->  Custom Scan (Citus Adaptive)  (cost=0.00..0.00 rows=0 width=0)                       │
│               Task Count: 32                                                                     │
│               Tasks Shown: One of 32                                                             │
│               ->  Task                                                                           │
│                     Node: host=localhost port=9700 dbname=postgres                               │
│                     ->  Limit  (cost=0.00..0.14 rows=10 width=4)                                 │
│                           ->  Seq Scan on test_102105 test  (cost=0.00..35.50 rows=2550 width=4) │
└──────────────────────────────────────────────────────────────────────────────────────────────────┘
(9 rows)

```

admin_guide/cluster_management.rst
onderkalaci: Makes sense

develop/api_guc.rst
```diff
@@ -336,30 +339,82 @@ Note that it may be useful to use :code:`error` during development testing, and
...
-************************************************
+When the individual tasks of a multi-shard query take very little time, they
+can often be finished over a single (often already cached) connection. To avoid
+redundantly opening additional connections, the executor waits between
+connection attempts for the configured number of milliseconds. At the end of
+the interval, it increases the number of connections it is allowed to open next
+time.
 
-Sets the maximum number of concurrent connections to the database server. The default is typically 100 connections, but might be less if your kernel settings will not support it (as determined during initdb). The real time executor maintains an open connection for each shard to which it sends queries. Increasing this configuration parameter will allow the executor to have more concurrent connections and hence handle more shards in parallel. This parameter has to be changed on the workers as well as the coordinator, and can be done only during server start.
+For long queries (those taking >500ms), slow start simply adds latency, but for
```
onderkalaci: maybe instead of `simply adds` we could use `might add` ?

develop/api_guc.rst
```diff
@@ -336,30 +339,82 @@ Note that it may be useful to use :code:`error` during development testing, and
   HINT:  Queries are split to multiple tasks if they have to be split into several queries on the workers.
   STATEMENT:  select * from foo;
 
-Real-time executor configuration
-$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$
+Adaptive executor configuration
+$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$
 
-The Citus query planner first prunes away the shards unrelated to a query and then hands the plan over to the real-time executor. For executing the plan, the real-time executor opens one connection and uses two file descriptors per unpruned shard. If the query hits a high number of shards, then the executor may need to open more connections than max_connections or use more file descriptors than max_files_per_process.
+citus.executor_slow_start_interval (integer)
```
onderkalaci: If there is no alpha-numeric (or anything else) ordering in the names, I'd suggest to start with `max_adaptive_executor_pool_size ` which is the most important setting.

develop/api_guc.rst
```diff
@@ -336,30 +339,82 @@ Note that it may be useful to use :code:`error` during development testing, and
...
+the interval, it increases the number of connections it is allowed to open next
+time.
 
-Sets the maximum number of concurrent connections to the database server. The default is typically 100 connections, but might be less if your kernel settings will not support it (as determined during initdb). The real time executor maintains an open connection for each shard to which it sends queries. Increasing this configuration parameter will allow the executor to have more concurrent connections and hence handle more shards in parallel. This parameter has to be changed on the workers as well as the coordinator, and can be done only during server start.
+For long queries (those taking >500ms), slow start simply adds latency, but for
+short queries it's faster. The default value is 10ms.
 
-max_files_per_process (integer)
-*******************************************************
+citus.max_adaptive_executor_pool_size (integer)
```
onderkalaci: Do you think would it make sense to mention that via `max_adaptive_executor_pool_size `, the users could do `resource management`. For example, if you have some queries that you wouldn't want to utilize too many connections/cores/memory etc., you can lower the value within the session (e.g., this is a session level GUC)

develop/api_guc.rst
```diff
@@ -336,30 +339,82 @@ Note that it may be useful to use :code:`error` during development testing, and
...
+Determines which SET commands are propagated from the coordinator to workers.
+The default value for this parameter is 'none'.
+
+The supported values are:
+
+* **none:** no SET commands are propagated.
+
+* **local:** only SET LOCAL commands are propagated.
+
+citus.force_max_query_parallelization (boolean)
```
onderkalaci: Would it make sense to mention the use case for this GUC we discussed in the meeting?

```SQL
BEGIN;

-- transaction starts with a very lightweight query,
-- which triggers very little connections to be used
-- by the executor
SELECT count(*) FROM table WHERE filter = x;

-- now, we've a very resource intensive query
-- that we want the executor to utilize as much resource/connection
-- as possible
-- however, the executor already decided the resource at the first
-- command in the transaction block
SELECT ... very .. complex..SQL;
````

So, to help with that case:
```SQL
BEGIN;
SET citus.force_max_query_parallelization TO ON;

SELECT count(*) FROM table WHERE filter = x;

SELECT ... very .. complex..SQL;
````
Now, both queries use as much resources as possible although the first command doesn't necessarily need it.

File: develop/api_guc.rst [hunk 14]
```diff
@@ -395,6 +450,31 @@ citus.partition_buffer_size (integer)
...
+
+max_files_per_process (integer)
+*******************************************************
+
+Sets the maximum number of simultaneously open files for each server process and defaults to 1000. The real-time executor requires two file descriptors for each shard it sends queries to. Increasing this configuration parameter will allow the executor to have more open file descriptors, and hence handle more shards in parallel. This change has to be made on the workers as well as the coordinator, and can be done only during server start.
+
+.. note::
+  Along with max_files_per_process, one may also have to increase the kernel limit for open file descriptors per process using the ulimit command.
+
+citus.enable_repartition_joins (boolean)
```
onderkalaci: `citus.enable_repartition_joins ` applies for adaptive executor as well. So, it is a more generic executor setting.

admin_guide/cluster_management.rst
jonels-msft: Yep, I do that above in the query

```sql
SELECT sum(calls),
         partition_key IS NOT NULL AS single_tenant
  FROM citus_stat_statements
  GROUP BY 2;
```

jonels-msft:
@onderkalaci I've addressed your first round of reviews.

develop/reference_workarounds.rst
```diff
@@ -5,18 +5,18 @@ SQL Support and Workarounds
...
 
 Even cross-node queries (used for parallel computations) support most SQL features. However some SQL features are not supported for queries which combine information from multiple nodes.
 
 **Limitations for Cross-Node SQL Queries:**
 
 * `Window functions <https://www.postgresql.org/docs/current/static/tutorial-window.html>`_ are supported only when they include the distribution column in PARTITION BY.
-* `SELECT … FOR UPDATE <https://www.postgresql.org/docs/current/static/sql-select.html#SQL-FOR-UPDATE-SHARE>`_ work in :ref:`router_executor` queries only
-* `TABLESAMPLE <https://www.postgresql.org/docs/current/static/sql-select.html#SQL-FROM>`_ work in :ref:`router_executor` queries only
+* `SELECT … FOR UPDATE <https://www.postgresql.org/docs/current/static/sql-select.html#SQL-FOR-UPDATE-SHARE>`_ work in single-shard queries only
```
onderkalaci: Should we use `single-tenant` for consistency instead of `single-shard`?

develop/reference_processing.rst
onderkalaci: I'm not sure whether we should count `push-pull` as a separate executor.

I don't know how to express this well, but I can explain to you and you can probably come-up with a better explanation of it.

`pull-push` executor is actually not a different executor. It is calling the current executor to execute `some parts` of the query like ctes or some subqueries. 

For example,

```SQL
WITH distinct_users AS (SELECT DISTINCT user_id FROM users_table)
SELECT count(*) FROM events_table WHERE user_id IN (SELECT user_id FROM distinct_users);
```

For example, for the above query, the execution happens in the following order:

1)  The query in `distinct_users` CTE is executed via adaptive executor as if it is a completely separate query. 
2) The results of the  `distinct_users` CTE is first pulled to the coordinator as usual query execution, and copied (`pushed`) to the workers as intermediate result. 
3) The final query `SELECT count(*) FROM events_table WHERE user_id IN (SELECT user_id FROM read_intermediate_result('9_1'::text, 'binary'::citus_copy_format) intermediate_result(user_id integer)) distinct_users)` is executed by adaptive executor. Basically, treating the intermediate result as a reference table, knowing that it exits on all worker nodes.

So, in that sense, I'd probably frame `pull-push` execution as a `tool` that adaptive executor (not task-tracker!) can use to execute more complex SQL that otherwise cannot be executed by Citus.

But, if framing it as a separate executor would make it easier to explain to the users, that's also not too bad.

jonels-msft:
@onderkalaci I removed the phrase talking about push-pull as if it was a distinct executor. (Some time later I should schedule time to get a walk-through of the actual codebase so that I become more familiar with some of these concepts.)

Any other things you want me to revise before we merge?

develop/api_guc.rst
```diff
@@ -336,30 +339,111 @@ Note that it may be useful to use :code:`error` during development testing, and
...
+Although increasing parallelism for a single query is generally good, it also
+puts connection pressure on the workers in a production environment.  This GUC
+provides a knob to trade off parallelism vs connection pressure.
+
+The default value is 16, however it's hard to make a good static choice
+manually. In the future the adaptive executor may adjust this GUC on the fly,
+shrinking the pool when workers are busy, and growing it when they are idle.
+
+This is a session-level GUC, and thus can be used to limit the resource usage
+for individual sessions. Lowering the the value can limit the queries' use of
```
onderkalaci: `the the` -> `the`

PR #413, opened 2017-07-19
Document how to use external log destinations

begriffs:
Fixes #345

PR #881, opened 2019-10-08
Do not advise TLS off for any logging service

fdr:
Do not advise TLS off for any logging service

This encourages people to send their query texts and the like as
plaintext over the Internet!  A quizzical customer pointed this out,
for a service where the only thing necessary to do was to change the
port and check the box, Logentries.

I since have doubled-back and verified:

* LogDNA
* Logentries / Rapid7 insightOps (acquired name)
* Loggly
* Papertrail

Two of the other providers require their certificate be installed;
thus, this patch must come after
https://github.com/citusdata/cloudplane/pull/957 is deployed.
Logentries uses GoDaddy and does not give cert bundle guidence,
Papertrail provides a cert bundle, but it is also a major vendor and
handled by the Mozilla bundle.

I suppose in principle it might be nice if the customer could define
the CA expected, or we only installed necessary certs for the vendor
selected, but, alas: complexity, and the current situation is much
worse.

Unfortunately, so far, we have been unable to turn up any rationale
for why we were so lackadaisical about TLS in our guidance: it simply
always was and passed without detectable remark in pull requests,
commit texts, comments, and emails.

The two times this could have been detected were when the CloudPlane
stuff was committed -- I was the reviewer, and could have asked why
TLS was optional -- and when the manual was updated to include the
three TLS-free provider directives.  Unlike obliged SOC2 procedures
nowadays, CloudCon's commit of this was unilateral. It was in Aug
2016, and the first customers were Apr 2016 or so.

Some hyperlinks from when Mistakes Were Made:

https://github.com/citusdata/citus_docs/pull/413
https://github.com/citusdata/cloudcon/pull/45
https://github.com/citusdata/cloudplane/pull/145


fdr:
Haven't yet verified two of the vendors. LogEntries has been verified by a customer. I'll write back soon.

fdr:
Testing done and documented.

fdr:
Addendum: one thing I noticed that every service that was doing TLS also used "legacy" BSD Syslog as given by their syslog-ng set-up steps. Not sure if that's necessary or they can actually accept new-style IETF syslog, but I did not see fit to push the envelope here. Until next year...

cloud/logging.rst
```diff
@@ -43,26 +43,26 @@ Replace :code:`<token>` with the custom token listed in the provider's web site.
...
-| Port             | 19493                   |
-+------------------+-------------------------+
-| TLS              | Yes                     |
-+------------------+-------------------------+
-| Protocol         | IETF Syslog             |
-+------------------+-------------------------+
-| Message Template |                         |
-+------------------+-------------------------+
++------------------+---------------------------+
+| Hostname         | logs<N>.papertrailapp.com |
```
jonels-msft: How should people determine the value for *N*? I had listed a specific value because it was what I had tested.

cloud/logging.rst
```diff
@@ -83,16 +83,16 @@ Replace :code:`<token>` with the custom token listed in the provider's web site.
 | Message Template | ``<${PRI}>1 ${ISODATE} ${HOST} ${PROGRAM} ${PID} ${MSGID} [<token>@41123] $MSG\\n`` |
 +------------------+-------------------------------------------------------------------------------------+
 
-**Logentries**
+**Logentries** / **Rapid7 insightOps**
 
 +------------------+------------------------------------+
-| Hostname         | data.logentries.com                |
+| Hostname         | us.data.logs.insight.rapid7.com    |
```
jonels-msft: Was Logentries acquired or something? Why is their hostname now rapid7?

cloud/logging.rst
fdr: It's given by papertrail. In my case it was `6`

cloud/logging.rst
fdr: Yeah, I do think so.

cloud/logging.rst
jonels-msft: > In my case it was 6

I'm wondering whether we should suggest a particular number so that our users don't need to take extra steps to find a number, or whether that number is actually specific to each person's account, and must vary.

> papertrail has no authentication whatsoever

Papertrail leaves no papertrail. 😉

jonels-msft:
Nice commit message by the way. I should make mine more descriptive like that.

fdr:
Not everything has a [deep reason](https://github.com/citusdata/cloudplane/commit/17a9f412cffc8d10c7e2fb7166ee01c27e6514f3). This one has a bit more history and ramification than most.[ This one is more normal](https://github.com/citusdata/cloudplane/commit/a3500a5b66a9a99d74b95d7d31bb196cf0a923a0) among the non-trivial messages.

I don't know if this is a model or anything, but I don't write like that all the time.

cloud/logging.rst
fdr: I'm not sure. Maybe we could ask. If it's load balanced, e.g. paper trail customers are less than 30000-60000 ports or so, then I guess, though papertrail may not like the pile-up. If it's part of the routing, then it won't work.

Thing is, if you look at it, the host:port is the *only disambiguating information they have* about who is who.

jonels-msft:
(Just cherry-picked this from master onto v8.3)

PR #885, opened 2019-10-30
Fix faq.rst (shard_count setting)

yhuelf:
It doesn't seem possible to  set a different shard count for two tables of a same co-location group.

### DIFF
--- a/faq/faq.rst
+++ b/faq/faq.rst
@@ -69,7 +69,7 @@ Remember that -- even with this extensive SQL coverage -- data modeling can have
 How do I choose the shard count when I hash-partition my data?
 --------------------------------------------------------------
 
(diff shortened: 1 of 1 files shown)


faq/faq.rst
```diff
@@ -69,7 +69,7 @@ Remember that -- even with this extensive SQL coverage -- data modeling can have
 How do I choose the shard count when I hash-partition my data?
 --------------------------------------------------------------
 
-One of the choices when first distributing a table is its shard count. This setting can be set differently for each table, and the optimal value depends on use-case. It is possible, but difficult, to change the count after cluster creation, so use these guidelines to choose the right size.
+One of the choices when first distributing a table is its shard count. This setting can be set differently for each co-location group, and the optimal value depends on use-case. It is possible, but difficult, to change the count after cluster creation, so use these guidelines to choose the right size.
```
jonels-msft: Good idea. Also, how about making it a link to explain the concept? Use a snippet like this:

```rst
:ref:`co-location group <colocation>`
```

PR #889, opened 2019-11-06
Object propagation

### LINKED ISSUE #887: More detailed section about object propagation from coordinator to workers

jonels-msft:
Fixes #887

### DIFF
--- a/develop/api_metadata.rst
+++ b/develop/api_metadata.rst
@@ -247,8 +247,8 @@ Distributed object table
 The citus.pg_dist_object table contains a list of objects such as types and
 functions that have been created on the coordinator node and propagated to
 worker nodes. When an administrator adds new worker nodes to the cluster, Citus
-automatically creates copies of the distributed objects on the new nodes, in
-the correct order to satisfy object dependencies.
+automatically creates copies of the distributed objects on the new nodes (in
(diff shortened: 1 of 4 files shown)


develop/api_metadata.rst
```diff
@@ -239,6 +239,72 @@ The pg_dist_node table contains information about the worker nodes in the cluste
...
+            color stoplight;
+    BEGIN
+            SELECT *
+              FROM unnest(enum_range(NULL::stoplight)) INTO color
+             ORDER BY random() LIMIT 1;
+            RETURN color;
+    END;
+    $$ LANGUAGE plpgsql VOLATILE;
+
+    SELECT create_distributed_function('intersection()');
```
jonels-msft: Maybe a function with distribution argument would have made a better example... The above is simply a way to get that regular function on the workers.

faq/faq.rst
```diff
@@ -109,23 +109,26 @@ Certain commands, when run on the coordinator node, do not get propagated to the
 
 * ``CREATE ROLE/USER``
 * ``CREATE FUNCTION``
```
thanodnl: I am in doubt if we should keep this here. We don't have tables in here either.
Any strong reasons to keep `CREATE FUNCTION` in this list?

PR #892, opened 2019-12-18
Highlight create_distributed_function

### LINKED ISSUE #886: Showcase create_distributed_function

First round: update the MT use case and/or tutorial.

In the performance section, we would like to demonstrate how using this makes a performance impact.

jonels-msft:
Fixes #886


jonels-msft:
@marcocitus one difficulty I ran into is that our tutorial is supposed to be able to run on Docker or a simple single-server installation. However when I do that I get this error:

```
ERROR:  cannot colocate function "delete_campaign" and table "campaigns"
DETAIL:  Citus currently only supports colocating function with distributed tables that are created using streaming replication model.
HINT:  When distributing tables make sure that citus.replication_model = 'streaming'
```

I had to run it on a PaaS such as Hyperscale that has streaming replication set up. Is there a workaround?

serprex:
Issue on why streaming is required: https://github.com/citusdata/citus/issues/3126

jonels-msft:
@marcocitus says we should be good to suggest people enable streaming replication in our tutorial, so I made the change.

PR #895, opened 2020-01-13
9.1 features

### LINKED ISSUE #893: Draining of nodes

When removing a node from the cluster you have to take care, since there are shards on that node.
Draining of nodes is used to transfer all the shards to other nodes in the cluster, after which you can safely remove the node from the cluster.

- Scaling down clusters
- Upgrading nodes to a bigger size and removing the smaller nodes

1. `rebalance_table_shards`http://docs.citusdata.com/en/v8.3/develop/api_udf.html#rebalance-table-shards
   - [x] document new `drain_only` argument of 
   - [x] document that providing a table is not necessary (anymore), providing no table will rebalance all existing colocation groups.
2. http://docs.citusdata.com/en/v8.3/develop/api_udf.html#cluster-management-and-repair-functions
   - [x] document `master_drain_node` 
3. http://docs.citusdata.com/en/v8.3/develop/api_udf.html#metadata-configuration-information
   - [x] document `master_set_node_property`

## How does this work? (devs)

There's a new column `shouldhaveshards` in `pg_dist_node`. If it's false, but there are shards on the machine the rebalancer will move them away from there. 

### Example sql

There’s two flows of draining nodes that could be useful for you:
Remove multiple nodes:
1.	Run this for each node that you want to remove:
    ```sql
    SELECT * from master_set_node_property(node_hostname, node_port, 'shouldhaveshards', false);
    ```
2.	`SELECT * FROM rebalance_table_shards(drain_only := true);`
3.	Wait until it finishes
4.	Kill all the nodes

Remove  a single node (using usability UDF for single node case):
1.	`SELECT * from master_drain_node(node_hostname, node_port);`
2.	Wait until the command finishes
3.	Kill the node

No

- PR: https://github.com/citusdata/citus-enterprise/pull/372
- regression tests: https://github.com/citusdata/citus-enterprise/pull/372/files#diff-b698af22c0ed49061162e6429e66df1bR490-R567

jonels-msft:
Fixes #893

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 _build
+.venv
--- a/develop/api_metadata.rst
+++ b/develop/api_metadata.rst
@@ -205,38 +205,40 @@ Worker node table
 
 The pg_dist_node table contains information about the worker nodes in the cluster. 
 
-+----------------+----------------------+---------------------------------------------------------------------------+
-|      Name      |         Type         |       Description                                                         |
-+================+======================+===========================================================================+
-| nodeid         |         int          | | Auto-generated identifier for an individual node.                       |
-+----------------+----------------------+---------------------------------------------------------------------------+
-| groupid        |         int          | | Identifier used to denote a group of one primary server and zero or more|
-|                |                      | | secondary servers, when the streaming replication model is used. By     |
-|                |                      | | default it is the same as the nodeid.                                   | 
-+----------------+----------------------+---------------------------------------------------------------------------+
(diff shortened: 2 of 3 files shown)


jonels-msft:
Should we mention `worker_create_schema`, `worker_repartition_cleanup`, `get_rebalance_table_shards_plan` and `alter_role_if_exists`? I thought they looked somewhat internal, so didn't cover them.

JelteF:
They are all pretty internal, although get_rebalance_table_shards_plan might be made useful for users in the future.

jonels-msft:
Great suggestions, thanks.

PR #908, opened 2020-03-10
CTE pushdown

### LINKED ISSUE #901: CTE Pushdown in Citus 9.2

Before Citus 9.2, all the CTEs are executed through pull-push execution. That adds considerable amount of unnecessary execution times.

In some (or many?) cases, some or all of the CTEs in the query can be pushed down similar to how subqueries are pushed down. This could give a huge boost in performance.

This is only valid for multi-shard queries, so we should probably explicitly mention that. In any way, the router queries are always pushed down, so it is not a concern.

The CTE pushdown is most useful in real-time analytics use-case, where CTEs and joins between CTEs/distributed tables follows the rules for (sub)query pushdown.
1) Tables should be colocated
2) The CTE queries should not require any merge steps (e.g., LIMIT or GROUP BY on non-distribution key)
3) Tables and CTEs should be joined on distribution keys 

As long as the above conditions are hold, the CTE is going to be pushed down. In case the CTE does not fit into the above rules, the CTE is executed via pull-push execution as it used to be. So, it is possible to have some CTEs are pushed down whereas some are pulled-pushed.

I think we should do two things.

First, in our docs, we explicitly or implicitly mention that all CTEs are going to use pull-push execution. That's not true anymore, pushdownable CTEs are going to be pushed down. We should find such reference and somehow relax that assumption.

Second, we might consider explicitly mentioning about CTE pushdown here: http://docs.citusdata.com/en/v9.1/sharding/data_modeling.html#co-location-means-better-feature-support 

We might consider having a separate section on this, please see  `Corner cases, gotchas` section where we need to mention about the limitations.

## How does this work? (devs)

Applicable CTEs are inlined as subqueries, and subquery optimizations are applied. This enables CTE pushdown. So, they are not actually pushed down as CTEs, but inlined as subqueries. See below for more details.

### Example sql

There are 3 cases. Common setup SQL

```SQL
CREATE TABLE users_table (user_id int, time timestamp, value_1 int, value_2 int, value_3 float, value_4 bigint);
SELECT create_distributed_table('users_table', 'user_id');

CREATE TABLE events_table (user_id int, time timestamp, event_type int, value_2 int, value_3 float, value_4 bigint);
SELECT create_distributed_table('events_table', 'user_id');



INSERT INTO users_table SELECT (i * random())::int % 10000, timestamp '2014-01-10 20:00:00' +
       random() * (timestamp '2014-01-20 20:00:00' -
                   timestamp '2014-01-10 10:00:00'),(i * random())::int % 10000, (i * random())::int % 10000, (i * random())::int % 10000 FROM generate_series(0, 1000000) i;


INSERT INTO events_table SELECT (i * random())::int % 10000, timestamp '2014-01-10 20:00:00' +
       random() * (timestamp '2014-01-20 20:00:00' -
                   timestamp '2014-01-10 10:00:00'),(i * random())::int % 10000, (i * random())::int % 10000, (i * random())::int % 10000 FROM generate_series(0, 1000000) i;

```

####  1) Case 1: Everything is pushdown-able. Both CTEs are colocated, joined on distribution key. Plus, the queries in CTEs doesn't have that require merge step (LIMIT, OFFSET or GROUP BY on non-distribution key etc)
```SQL
WITH first_pushdownable_cte AS
(
	 SELECT user_id, max(value_1) as max FROM users_table GROUP BY user_id
),
second_pushdownable_cte AS
(
	SELECT user_id, max(value_2) as max FROM events_table GROUP BY user_id
)
SELECT count(*) FROM first_pushdownable_cte JOIN second_pushdownable_cte USING (user_id, max);
```

This is what we send to workers: 
```SQL
SELECT count(*) AS count FROM (SELECT FROM ((SELECT users_table.user_id, max(users_table.value_1) AS max FROM public.users_table_102542 users_table GROUP BY users_table.user_id) first_pushdownable_cte JOIN (SELECT events_table.user_id, max(events_table.value_2) AS max FROM public.events_table_102550 events_table GROUP BY events_table.user_id) second_pushdownable_cte USING (user_id, max))) worker_subquery
```

#### 2) Case 2: The first CTE is still pushdown-able, but the second one is not because it is not joined on distribution key with the first CTE

```SQL
WITH first_pushdownable_cte AS
(
	 SELECT user_id, max(value_1) as max FROM users_table GROUP BY user_id
),
second_non_pushdownable_cte AS
(
	SELECT user_id, max(value_2) as max FROM events_table GROUP BY user_id
)
SELECT count(*) FROM first_pushdownable_cte JOIN second_non_pushdownable_cte USING (max) WHERE first_pushdownable_cte.user_id != second_non_pushdownable_cte.max;
```

This is what we send to workers: 
```SQL
SELECT count(*) AS count FROM (SELECT FROM ((SELECT users_table.user_id, max(users_table.value_1) AS max FROM public.users_table_102542 users_table GROUP BY users_table.user_id) first_non_pushdownable_cte JOIN (SELECT intermediate_result.user_id, intermediate_result.max FROM read_intermediate_result('224_1'::text, 'binary'::citus_copy_format) intermediate_result(user_id integer, max integer)) second_non_pushdownable_cte USING (max)) WHERE (first_non_pushdownable_cte.user_id OPERATOR(pg_catalog.<>) second_non_pushdownable_cte.max)) worker_subquery
```

#### 3) Case 3: None of the CTEs are pushdown-able as they have GROUP BY on non-dist key.

```SQL
WITH first_non_pushdownable_cte AS
(
	 SELECT max(user_id) as max FROM users_table GROUP BY value_1
),
second_non_pushdownable_cte AS
(
	SELECT max(user_id) as max FROM events_table GROUP BY value_2
)
SELECT count(DISTINCT first_non_pushdownable_cte.max) FROM first_non_pushdownable_cte JOIN second_non_pushdownable_cte USING (max) WHERE first_non_pushdownable_cte.max = second_non_pushdownable_cte.max;

```

So, the final query has only intermediate results. This is how it used to work for all CTEs:
```SQL
SELECT count(DISTINCT first_non_pushdownable_cte.max) AS count FROM ((SELECT intermediate_result.max FROM read_intermediate_result('239_1'::text, 'binary'::citus_copy_format) intermediate_result(max integer)) first_non_pushdownable_cte JOIN (SELECT intermediate_result.max FROM read_intermediate_result('239_2'::text, 'binary'::citus_copy_format) intermediate_result(max integer)) second_non_pushdownable_cte USING (max)) WHERE (first_non_pushdownable_cte.max OPERATOR(pg_catalog.=) second_non_pushdownable_cte.max)
```

#### 1) Postgres version 
- This feature is only avaliable in PG 12. Some background: Before PostgreSQL 12, CTEs are considered to be optimization fences on Postgres (and also in Citus). However, in PostgreSQL 12, postgres hackers implemented a feature called [`CTE inlining`](https://www.depesz.com/2019/02/19/waiting-for-postgresql-12-allow-user-control-of-cte-materialization-and-change-the-default-behavior/). In simple words, convert CTEs into subqueries, so that they can benefit from many optimizations that Postgres implements for subqueries.

This approach inspired us since Citus was doing the same. We've a very optimized way of planning subqueries, but not CTEs. So, when Postgres 12 is avaliable, Citus applies the same optimizations as Postgres (a.k.a., inlines CTEs) to apply the optimizations for subqueries.

**Thus, this feature is mostly useful in Postgres 12**.  There are few cases that Citus can benefit from this feature in Postgres 11. But, since they are very limiting, I'd suggest us to position this feature only avaliable on Postgres 12, by referring to CTE inlining in Postgres 12.

#### 2) Controlling CTE inlining:

Citus obeys the `MATERIALIZED` and `NOT MATERIALIZED` keywords as in **[Postgres](https://www.postgresql.org/docs/12/queries-with.html)**.  On Citus jargon, `MATERIALIZED` means pull-push execution, `NOT MATERIALIZED` means try to inline if possible. 

If none is specified, the default action is that, as in Postgres, if the CTE is referenced only once, do `NOT MATERIALIZED`. Else, use `MATERIALIZED` as executing the same CTE twice might be expensive.

On citus, since we support Postgres 11, we have another GUC to control the cte inlining. Users can simply disable all CTE inlining with `set citus.enable_cte_inlining TO false;`. We don't need to expose this GUC to users, sharing as a reference mostly.

- On postgres side, if you search for `CTE Inlining`, you'd find lots of information. Though, I don't know if this item asks for it or not. 

The commits are meant to be self-contained, so could be useful to understand what is going on: https://github.com/citusdata/citus/pull/3161

jonels-msft:
Fixes #901

@onderkalaci, I removed cte_fast, and cte_slow, and added more hints about avoiding push-pull execution.

I know this has a lot less detail than your issue, but I'm wondering how much we need to expose. I also wasn't sure when someone would want to mark their CTEs as MATERIALIZED, so I didn't include that.

Thought I'd just check in at this point and see what you thought.

### DIFF
--- a/develop/api_guc.rst
+++ b/develop/api_guc.rst
@@ -251,7 +251,7 @@ Use the binary copy format to transfer data between coordinator and the workers.
 citus.max_intermediate_result_size (integer)
 $$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$
 
-The maximum size in KB of intermediate results for CTEs and complex subqueries. The default is 1GB, and a value of -1 means no limit. Queries exceeding the limit will be canceled and produce an error message.
+The maximum size in KB of intermediate results for CTEs that are unable to be pushed down to worker nodes for execution, and for complex subqueries. The default is 1GB, and a value of -1 means no limit. Queries exceeding the limit will be canceled and produce an error message.
 
 DDL
 -------------------------------------------------------------------
--- a/performance/performance_tuning.rst
+++ b/performance/performance_tuning.rst
@@ -218,7 +218,7 @@ When using CTEs, or joins between CTEs and distributed tables, you can avoid pus
 * The CTE queries should not require any merge steps (e.g., LIMIT or GROUP BY on a non-distribution key)
 * Tables and CTEs should be joined on distribution keys
 
-Also PostgreSQL 12 or above allows Citus to take advantage of *CTE inlining* to push CTEs down to workers in more circumstances.
+Also PostgreSQL 12 or above allows Citus to take advantage of *CTE inlining* to push CTEs down to workers in more circumstances. The inlining behavior can be controlled with the ``MATERIALIZED`` keyword -- see the `PostgreSQL docs <https://www.postgresql.org/docs/current/queries-with.html>`_ for details.
 
 .. _advanced_performance_tuning:
 
--- a/performance/performance_tuning.rst
+++ b/performance/performance_tuning.rst
@@ -212,19 +212,13 @@ If the query has subqueries or CTEs that exceed this limit, the query will be ca
   DETAIL:  Citus restricts the size of intermediate results of complex subqueries and CTEs to avoid accidentally pulling large result sets into once place.
   HINT:  To run the current query, set citus.max_intermediate_result_size to a higher value or -1 to disable.
 
-If this happens, consider whether you can move limits or filters inside CTEs/subqueries. For instance
+When using CTEs, or joins between CTEs and distributed tables, you can avoid push-pull execution by following these rules:
 
-.. code-block:: sql
+* Tables should be colocated
+* The CTE queries should not require any merge steps (e.g., LIMIT or GROUP BY on a non-distribution key)
+* Tables and CTEs should be joined on distribution keys
 
-  -- It's slow to retrieve all rows and limit afterward
-
-  WITH cte_slow AS (SELECT * FROM users_table)
-  SELECT * FROM cte_slow LIMIT 10;
-
-  -- Limiting inside makes the intermediate results small
-
-  WITH cte_fast AS (SELECT * FROM users_table LIMIT 10)
-  SELECT * FROM cte_fast;
+Also PostgreSQL 12 or above allows Citus to take advantage of *CTE inlining* to push CTEs down to workers in more circumstances.
 
 .. _advanced_performance_tuning:
 
--- a/sharding/data_modeling.rst
+++ b/sharding/data_modeling.rst
@@ -244,6 +244,7 @@ The full list of Citus features that are unlocked by co-location are:
 * Aggregation through INSERT..SELECT
 * Foreign keys
 * Distributed outer joins
+* Pushdown CTEs (requires PostgreSQL >=12)
 
 Data co-location is a powerful technique for providing both horizontal scale and support to relational data models. The cost of migrating or building applications using a distributed database that enables relational operations through co-location is often substantially lower than moving to a restrictive data model (e.g. NoSQL) and, unlike a single-node database, it can scale out with the size of your business. For more information about migrating an existing database see :ref:`Transitioning to a Multi-Tenant Data Model <transitioning_mt>`.
 


onderkalaci:
The changes look good, but I think it'd be nice to mention about MATERIALIZED case, as that's already a Postgres [feature](https://www.postgresql.org/docs/12/queries-with.html) and we're obeying Postgres' behavior during the distributed execution.

We could simply say that: `The inlining behavior can be controlled with MATERIALIZED keyword, see Postgres docs for the details` right after the changes in `performance/performance_tuning.rst`

PR #915, opened 2020-03-30
Add extra information around certificates

### LINKED ISSUE #909: (WIP) How to set up SSL/TLS when running a self hosted Citus

Requirement from security compliance team within Microsoft

1. What does Citus do by default? A: for new clusters self-signed certificates, for old clusters no encryption by default
2. Why are self-signed certificates not secure (and thus also the Citus default)? A: MITM attack is still easily possible
3. What does a user need to do to get `ssl=verify-full` working? A: Get certificates from a CA (either company internal PKI or publicly trusted). Install those certificates. Manage key rotation. Do this for all nodes in the cluster.
4. Mention `hostssl` for denying unencrypted connections 

Postgres docs on setting up SSL/TLS
pg_auto_failover docs on this same topic:
https://github.com/citusdata/pg_auto_failover/blob/release-1.2/docs/security.rst

thanodnl:
Citus has some defaults around newly created clusters (since 8.1) that do setup self signed certificates. It needs to be explicitly stated that self signed certificates only prevent to passive Man-In-The-Middle attacks.

If users upgrade from a Citus cluster that predates the auto self signed certificate setup we have a link to the official postgres documentation on how to setup certificates for postgres, these steps can be followed on the coordinator and the workers to achieve the same. By combining the steps with an actual Certificate Authority the reader could also setup a fully verified TLS connection.

Since 9.2.4 (to be released instead of our wrongfully packaged 9.2.3) it is possible to make sure Citus closes all connections any backend has open after a configuration change by reloading the postgres configuration via `pg_reload_conf()`. This is reflected in the documentation with a warning indicating a database restart is required in versions prior to 9.2.4.

Fixes #909

### DIFF
--- a/admin_guide/cluster_management.rst
+++ b/admin_guide/cluster_management.rst
@@ -555,8 +555,8 @@ Once all files are in place on the nodes the following settings need to be confi
 
 .. code-block:: ini
 
-   # the following settings enable the postgres server to enable ssl and configure the server to present the certifacte
-   # to clients when connecting over tls/ssl
+   # the following settings enable the postgres server to enable ssl and configure the 
+   # server to present the certificate to clients when connecting over tls/ssl
    ssl = on
    ssl_key_file = '/path/to/server.key'
    ssl_cert_file = '/path/to/server.crt'
--- a/admin_guide/cluster_management.rst
+++ b/admin_guide/cluster_management.rst
@@ -480,7 +480,7 @@ Connection Management
 
    Since Citus version 8.1.0 (released 2018-12-17) the traffic between the different nodes in the cluster is encrypted for NEW installations. This is done by using TLS with self-signed certificates. This means that this **does not protect against Man-In-The-Middle attacks.** This only protects against passive eavesdropping on the network.
 
-   For setting up self-signed TLS on installations that were originally created before Citus version 8.1.0 follow the steps in `official postgres documentation <https://www.postgresql.org/docs/current/ssl-tcp.html#SSL-CERTIFICATE-CREATION>`_ together with the citus specific settings described here. Setup should be done on coordinator and workers.
+   Clusters originally created with a Citus version before 8.1.0 do not have any network encryption enabled between nodes (even if upgraded later). To set up self-signed TLS on on this type of installation follow the steps in `official postgres documentation <https://www.postgresql.org/docs/current/ssl-tcp.html#SSL-CERTIFICATE-CREATION>`_ together with the citus specific settings described here, i.e. changing ``citus.node_conninfo`` to ``sslmode=require``. This setup should be done on coordinator and workers.
 
 When Citus nodes communicate with one another they consult a GUC for connection parameters and, in the Enterprise Edition of Citus, a table with connection credentials. This gives the database administrator flexibility to adjust parameters for security and efficiency.
 
--- a/admin_guide/cluster_management.rst
+++ b/admin_guide/cluster_management.rst
@@ -505,7 +505,7 @@ After changing this setting it is important to reload the postgres configuration
 
 .. warning:: 
 
-   Versions before 9.2.4 require a restart for existing connections to be closed.
+   Citus versions before 9.2.4 require a restart for existing connections to be closed.
 
    For these versions a reload of the configuration does not trigger connection ending and subsequent reconnecting. Instead the server should be restarted to enforce all connections to use the new settings.
 
--- a/admin_guide/cluster_management.rst
+++ b/admin_guide/cluster_management.rst
@@ -491,8 +491,7 @@ To set non-sensitive libpq connection parameters to be used for all node connect
   -- key=value pairs separated by spaces.
   -- For example, ssl options:
 
-  ALTER DATABASE foo
-  SET citus.node_conninfo =
+  ALTER SYSTEM SET citus.node_conninfo =
     'sslrootcert=/path/to/citus-ca.crt sslcrl=/path/to/citus-ca.crl sslmode=verify-full';
 
 There is a whitelist of parameters that the GUC accepts, see the :ref:`node_conninfo <node_conninfo>` reference for details. As of Citus 8.1, the default value for node_conninfo is ``sslmode=require``, which prevents unencrypted communication between nodes. If your cluster was originally created before Citus 8.1 the value will be ``sslmode=prefer``. After setting up self-signed certificates on all nodes it's recommended to change this setting to ``sslmode=require``.
--- a/admin_guide/cluster_management.rst
+++ b/admin_guide/cluster_management.rst
@@ -562,9 +562,9 @@ Once all files are in place on the nodes the following settings need to be confi
    ssl_cert_file = '/path/to/server.crt'
 
    # this will tell citus to verify the certificate of the server it is connecting to 
-   citus.node_conninfo = 'sslmode=verify-ca sslrootcert=/path/to/ca.crt sslcrl=/path/to/ca.crl'
+   citus.node_conninfo = 'sslmode=verify-full sslrootcert=/path/to/ca.crt sslcrl=/path/to/ca.crl'
 
-Depending on the policy of the Certificate Authority used you might need or want to change ``sslmode=verify-ca`` in ``citus.node_conninfo`` to ``sslmode=verify-full``. For the difference between the two settings please consult `the official postgres documentation <https://www.postgresql.org/docs/current/libpq-ssl.html#LIBPQ-SSL-SSLMODE-STATEMENTS>`_.
+Depending on the policy of the Certificate Authority used you might need or want to change ``sslmode=verify-full`` in ``citus.node_conninfo`` to ``sslmode=verify-ca``. For the difference between the two settings please consult `the official postgres documentation <https://www.postgresql.org/docs/current/libpq-ssl.html#LIBPQ-SSL-SSLMODE-STATEMENTS>`_.
 
 Lastly, to prevent any user from connecting via an unencrypted connection changes need to be made to ``pg_hba.conf``. Many Postgres installations will have entries allowing ``host`` connections which both allow SSL/TLS connections as well as plain TCP connections. By changing all ``host`` entries with ``hostssl`` entries only encrypted connections will be allowed to authenticate to Postgres. For full documentation on these settings take a look at `the pg_hba.conf file <https://www.postgresql.org/docs/current/auth-pg-hba-conf.html>`_ documentation on the official Postgres documentation.
 
--- a/admin_guide/cluster_management.rst
+++ b/admin_guide/cluster_management.rst
@@ -536,11 +536,11 @@ Changing ``pg_dist_authinof`` does not force any existing connection to reconnec
 Setup Certificate Authority signed certificates
 -----------------------------------------------
 
-This section assumes you have a trusted Certificate Authority that can issue server certificates to you for all nodes in your cluster. It is recommended to work with the security department in your organisation to prevent key material be handled incorrectly. This guide only covers Citus specific configuration that needs to be applied, no best practices on PKI management.
+This section assumes you have a trusted Certificate Authority that can issue server certificates to you for all nodes in your cluster. It is recommended to work with the security department in your organization to prevent key material from being handled incorrectly. This guide covers only Citus specific configuration that needs to be applied, not best practices for PKI management.
 
 For all nodes in the cluster you need to get a valid certificate signed by the *same Certificate Authority*. The following **machine specific** files are assumed to be available on every machine:
  * ``/path/to/server.key``: Server Private Key
- * ``/path/to/server.crt``: Server Certifiacte or Certificate Chain for Server Key, signed by trusted Certifiacte Authority.
+ * ``/path/to/server.crt``: Server Certificate or Certificate Chain for Server Key, signed by trusted Certificate Authority.
 
 Next to these machine specific files you need these cluster or CA wide files available:
  * ``/path/to/ca.crt``: Certificate of the Certificate Authority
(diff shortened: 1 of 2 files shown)


admin_guide/cluster_management.rst
```diff
@@ -487,23 +493,47 @@ To set non-sensitive libpq connection parameters to be used for all node connect
 
   ALTER DATABASE foo
   SET citus.node_conninfo =
-    'sslrootcert=/path/to/citus.crt sslmode=verify-full';
+    'sslrootcert=/path/to/citus-ca.crt sslcrl=/path/to/citus-ca.crl sslmode=verify-full';
```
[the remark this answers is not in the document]
thanodnl: added a section on how to setup with certificates from a trusted CA. Containing a note that one can setup a CA them self in the virtue of a self-signed root certificate.

admin_guide/cluster_management.rst
```diff
@@ -487,23 +493,85 @@ To set non-sensitive libpq connection parameters to be used for all node connect
...
 
 There is a whitelist of parameters that the GUC accepts, see the :ref:`node_conninfo <node_conninfo>` reference for details. As of Citus 8.1, the default value for node_conninfo is ``sslmode=require``, which prevents unencrypted communication between nodes.
```
JelteF:
```suggestion
There is a whitelist of parameters that the GUC accepts, see the :ref:`node_conninfo <node_conninfo>` reference for details. As of Citus 8.1, the default value for node_conninfo is ``sslmode=require``, which prevents unencrypted communication between nodes. If your cluster was originally created before Citus 8.1 the value will be ``sslmode=prefer``. After setting up self-signed certificates on all nodes it's recommended to change this setting to ``sslmode=require``.
```

admin_guide/cluster_management.rst
```diff
@@ -487,23 +493,85 @@ To set non-sensitive libpq connection parameters to be used for all node connect
...
+.. note::
+
+   The Certificate Revocation List is likely to change over time. Work with your security department on a mechanism to get up to date versions of the revocation list on to all nodes in the cluster in a timely manner. A reload of every node in the cluster is required after the revocation list has been updated.
+
+Once all files are in place on the nodes the following settings need to be configured in the Postgres configuration file:
+
+.. code-block:: ini
+
+   # the following settings enable the postgres server to enable ssl and configure the server to present the certifacte
+   # to clients when connecting over tls/ssl
```
JelteF: This comment was not fully readable in the default layout without scrolling right
```suggestion
   # the following settings enable the postgres server to enable ssl and configure the 
   # server to present the certificate to clients when connecting over tls/ssl
```

admin_guide/cluster_management.rst
```diff
@@ -476,6 +476,12 @@ Security
 Connection Management
 ---------------------
 
+.. note::
+
+   Since Citus version 8.1.0 (released 2018-12-17) the traffic between the different nodes in the cluster is encrypted for NEW installations. This is done by using TLS with self-signed certificates. This means that this **does not protect against Man-In-The-Middle attacks.** This only protects against passive eavesdropping on the network.
+
+   For setting up self-signed TLS on installations that were originally created before Citus version 8.1.0 follow the steps in `official postgres documentation <https://www.postgresql.org/docs/current/ssl-tcp.html#SSL-CERTIFICATE-CREATION>`_ together with the citus specific settings described here. Setup should be done on coordinator and workers.
```
JelteF:
```suggestion
   Clusters originally created with a Citus version before 8.1.0 do not have any network encryption enabled between nodes (even if upgraded later). To set up self-signed TLS on on this type of installation follow the steps in `official postgres documentation <https://www.postgresql.org/docs/current/ssl-tcp.html#SSL-CERTIFICATE-CREATION>`_ together with the citus specific settings described here, i.e. changing ``citus.node_conninfo`` to ``sslmode=require``. This setup should be done on coordinator and workers.
```

admin_guide/cluster_management.rst
```diff
@@ -487,23 +493,85 @@ To set non-sensitive libpq connection parameters to be used for all node connect
...
+   # the following settings enable the postgres server to enable ssl and configure the server to present the certifacte
+   # to clients when connecting over tls/ssl
+   ssl = on
+   ssl_key_file = '/path/to/server.key'
+   ssl_cert_file = '/path/to/server.crt'
+
+   # this will tell citus to verify the certificate of the server it is connecting to 
+   citus.node_conninfo = 'sslmode=verify-ca sslrootcert=/path/to/ca.crt sslcrl=/path/to/ca.crl'
+
+Depending on the policy of the Certificate Authority used you might need or want to change ``sslmode=verify-ca`` in ``citus.node_conninfo`` to ``sslmode=verify-full``. For the difference between the two settings please consult `the official postgres documentation <https://www.postgresql.org/docs/current/libpq-ssl.html#LIBPQ-SSL-SSLMODE-STATEMENTS>`_.
```
JelteF: Let's recommend the most secure option.
```suggestion
   citus.node_conninfo = 'sslmode=verify-full sslrootcert=/path/to/ca.crt sslcrl=/path/to/ca.crl'

Depending on the policy of the Certificate Authority used you might need or want to change ``sslmode=verify-full`` in ``citus.node_conninfo`` to ``sslmode=verify-ca``. For the difference between the two settings please consult `the official postgres documentation <https://www.postgresql.org/docs/current/libpq-ssl.html#LIBPQ-SSL-SSLMODE-STATEMENTS>`_.
```

admin_guide/cluster_management.rst
```diff
@@ -487,23 +493,85 @@ To set non-sensitive libpq connection parameters to be used for all node connect
...
+.. code-block:: ini
+
+   # the following settings enable the postgres server to enable ssl and configure the server to present the certifacte
+   # to clients when connecting over tls/ssl
+   ssl = on
+   ssl_key_file = '/path/to/server.key'
+   ssl_cert_file = '/path/to/server.crt'
+
+   # this will tell citus to verify the certificate of the server it is connecting to 
+   citus.node_conninfo = 'sslmode=verify-ca sslrootcert=/path/to/ca.crt sslcrl=/path/to/ca.crl'
```
JelteF: When doing this on a running cluster you will want to split up the `ssl=on` and the `citus.node_conninfo` changes. Otherwise you will get downtime, because not all servers are using the new keys/certificates when accepting connections.

thanodnl: I don't think the documentation becomes more clear by adding this layer on top in the Security section.

@begriffs do we have documentation for zero downtime upgrades? Maybe we can fold this information in there.

thanodnl:
Agree with most of your changes, have applied them.

The only thing I have concerns about is adding the zero downtime considerations to an already complicated piece of documentation. Given the importance for users to understand the changes I think this should be captured elsewhere that we can link from here, to not clutter security considerations with operational considerations.

admin_guide/cluster_management.rst
```diff
@@ -487,23 +493,85 @@ To set non-sensitive libpq connection parameters to be used for all node connect
...
+
+Next to these machine specific files you need these cluster or CA wide files available:
+ * ``/path/to/ca.crt``: Certificate of the Certificate Authority
+ * ``/path/to/ca.crl``: Certificate Revocation List of the Certificate Authority
+
+.. note::
+
+   The Certificate Revocation List is likely to change over time. Work with your security department on a mechanism to get up to date versions of the revocation list on to all nodes in the cluster in a timely manner. A reload of every node in the cluster is required after the revocation list has been updated.
+
+Once all files are in place on the nodes the following settings need to be configured in the Postgres configuration file:
```
JelteF:
```suggestion
Once all files are in place on the nodes the following settings need to be configured in the Postgres configuration file (PostgreSQL needs to be restarted for these settings to be applied):
```

admin_guide/cluster_management.rst
JelteF: Makes sense, this is not required for this for this first version anyway. I did leave one more comment, which notes that you need to restart PostgreSQL for these settings to apply.

admin_guide/cluster_management.rst
thanodnl: I don't think postgres requires a restart. I believe ssl changes are reloaded in postgres since pg10. I think adding a reload is required makes sense.

admin_guide/cluster_management.rst
jonels-msft: We have a section about [upgrading citus](https://docs.citusdata.com/en/v9.2/admin_guide/upgrading_citus.html) itself, but it doesn't talk about downtime. We could add a section in there.

admin_guide/cluster_management.rst
JelteF: Makes sense, I updated my commit suggestion accordingly.

admin_guide/cluster_management.rst
```diff
@@ -487,23 +493,85 @@ To set non-sensitive libpq connection parameters to be used for all node connect
...
-    'sslrootcert=/path/to/citus.crt sslmode=verify-full';
+    'sslrootcert=/path/to/citus-ca.crt sslcrl=/path/to/citus-ca.crl sslmode=verify-full';
+
+There is a whitelist of parameters that the GUC accepts, see the :ref:`node_conninfo <node_conninfo>` reference for details. As of Citus 8.1, the default value for node_conninfo is ``sslmode=require``, which prevents unencrypted communication between nodes. If your cluster was originally created before Citus 8.1 the value will be ``sslmode=prefer``. After setting up self-signed certificates on all nodes it's recommended to change this setting to ``sslmode=require``.
+
+After changing this setting it is important to reload the postgres configuration. Even though the changed setting might be visible in all sessions, the setting is only consulted by Citus when new connections are established. When a reload signal is received citus marks all existing connections to be closed which causes a reconnect after running transactions have been completed.
+
+.. code-block:: postgresql
+
+  SELECT pg_reload_conf();
```
jonels-msft: Is this something that needs to be propagated through `run_command_on_workers()`?

admin_guide/cluster_management.rst
```diff
@@ -487,23 +493,85 @@ To set non-sensitive libpq connection parameters to be used for all node connect
...
 
-There is a whitelist of parameters that the GUC accepts, see the :ref:`node_conninfo <node_conninfo>` reference for details. As of Citus 8.1, the default value for node_conninfo is ``sslmode=require``, which prevents unencrypted communication between nodes.
+.. warning:: 
+
+   Versions before 9.2.4 require a restart for existing connections to be closed.
```
JelteF:
```suggestion
   Citus versions before 9.2.4 require a restart for existing connections to be closed.
```

admin_guide/cluster_management.rst
```diff
@@ -487,23 +493,85 @@ To set non-sensitive libpq connection parameters to be used for all node connect
 
   ALTER DATABASE foo
```
JelteF: What was the reason we have `ALTER DATABASE foo` here. During testing I always used `ALTER SYSTEM`. Does it make sense to change that here?

admin_guide/cluster_management.rst
JelteF: Good question. That depends a bit what you mean. If you only change `citus.node_conninfo` on one node, there's no reason to reload the configuration on the other nodes. Often you probably want to change the configuration on all nodes. So then it would make sense to do something like:
```sql
ALTER SYSTEM SET citus.node_conninfo =
    'sslrootcert=/path/to/root.crt sslcrl=/path/to/root.crl sslmode=verify-full';
select pg_reload_conf();
select run_commands_on_workers($$
ALTER SYSTEM SET citus.node_conninfo =
    'sslrootcert=/path/to/root.crt sslcrl=/path/to/root.crl sslmode=verify-full';
$$)
select run_commands_on_workers($$
select pg_reload_conf();
$$)
```

This assumes these files are in the same place on all nodes though.

I think it's fine to leave this like this for the initial version of this change.

admin_guide/cluster_management.rst
thanodnl: This was already in here, @begriffs is best to answer the question on historical reasons I guess.

admin_guide/cluster_management.rst
thanodnl: I have more generally phrased it and put it below the config block instead. Do you think it is ok like that now?

admin_guide/cluster_management.rst
thanodnl: I think that makes more sense to have a sprint at some point where we pay more attention to the zero downtime operations part of citus.

zero downtime is frankly not a security concern :)

admin_guide/cluster_management.rst
```diff
@@ -487,23 +493,103 @@ To set non-sensitive libpq connection parameters to be used for all node connect
...
+   # the following settings enable the postgres server to enable ssl and configure the 
+   # server to present the certificate to clients when connecting over tls/ssl
+   ssl = on
+   ssl_key_file = '/path/to/server.key'
+   ssl_cert_file = '/path/to/server.crt'
+
+   # this will tell citus to verify the certificate of the server it is connecting to 
+   citus.node_conninfo = 'sslmode=verify-full sslrootcert=/path/to/ca.crt sslcrl=/path/to/ca.crl'
+
+After changing either restart the database or reload the configuration.
```
JelteF:
```suggestion
After changing either restart the database or reload the configuration (a restart is required if a Citus version below 9.2.4 is used).
```

admin_guide/cluster_management.rst
jonels-msft: There's no deep history -- I introduced it in 626fab09. If alter system is preferred then it's fine to use that.

admin_guide/cluster_management.rst
JelteF: I added a small comment updating that wording. I think with that it's okay.

admin_guide/cluster_management.rst
JelteF: Okay, pushed commit that changes it to `ALTER SYSTEM`. That way we're sure it works.

admin_guide/cluster_management.rst
```diff
@@ -485,25 +491,104 @@ To set non-sensitive libpq connection parameters to be used for all node connect
...
+
+To verify the connections from the coordinator to the workers are encrypted you can run the following query. It will show the SSL/TLS version used to encrypt the connection that the coordinator uses to talk to the worker:
+
+.. code-block:: postgresql
+
+  SELECT run_command_on_workers($$
+    SELECT version FROM pg_stat_ssl WHERE pid = pg_backend_pid()
+  $$);
+  
+  ┌────────────────────────────┐
```
jonels-msft: Sphinx says: "Could not lex literal_block as "postgresql". Highlighting skipped."

I'd advise splitting these into two adjacent code blocks, like

```rst
.. code-block:: postgresql

   SELECT foo;

::

   output
```

jonels-msft:
Fixes #909

Issue #897, opened 2020-01-17
Update trigger docs to use %L instead of %s


serprex:
https://github.com/citusdata/citus/issues/2354

The docs they're referring to:

https://github.com/citusdata/citus_docs/blob/85c1de3d92a73509c65361dece94400a5b5d272a/develop/reference_propagation.rst

https://github.com/citusdata/citus_docs/blob/85c1de3d92a73509c65361dece94400a5b5d272a/extra/triggers.rst

Marco's suggesting %L should be used over %s

jonels-msft:
Thanks for this suggestion, I'll update and test the instructions in the next week or so.

PR #919, opened 2020-04-08
Pass identifier to function using %L

jonels-msft:
Fixes https://github.com/citusdata/citus_docs/issues/897


jonels-msft:
@serprex thanks for the review. By the way, I usually use "Squash and merge" rather than creating a merge commit on master. That way the subsequent cherry-pick to e.g. v9.2 is a little cleaner.

Issue #894, opened 2019-12-31
Arbitrary aggregates are supported


serprex:
## Why are we implementing it? (sales eng)

Users like being able to use aggregates

### What are the typical use cases?

https://github.com/citusdata/citus/issues/120

https://github.com/citusdata/citus/issues/3212

## Communication goals (e.g. detailed howto vs orientation)

I don't understand this header

### Good locations for content in docs structure

Wherever we discuss aggregates

## How does this work? (devs)

https://github.com/citusdata/citus/pull/3307
One avenue is that if you're distributing by the distribution column, we can push down all grouping to the worker & that can include arbitrary aggregates

https://github.com/citusdata/citus/pull/3324
The other avenue is that when you're not grouping by the distribution column & your query isn't applicable for partial aggregation, then we can eschew pushing down grouping in order to do it all on the coordinator

### Example sql

`select string_agg(x, 'x') from distributed_table`

`select random_column, sum(distinct value) from distributed_table group by random_column`

`select distribution_key, sum(distinct value) from distributed_table group by distribution_key`

The first two would have grouping on coordinator. The last would have grouping on workers

### Corner cases, gotchas

This could surprise users where `sum(x)` is distributed whereas `sum(distinct x)` is pulling up the entire set if input records. Additionally row-gather disables partial aggregation for the entire query, so `select sum(value1), sum(distinct value2) from distributed_table` will have to pull up intermediate rows for `sum(value1)` too. Also due to partial support for `COMBINEFUNC` use of `stddev(float8)` or `stddev(float4)` will be distributed across workers, whereas `stddev(integer)` will pull up to coordinator

In order to implement [`FILTER`](https://www.postgresql.org/docs/12/sql-expressions.html#SYNTAX-AGGREGATES) when pulling up grouping we only compute the boolean result on workers. So while usually `FILTER` will skip computing the rest of the aggregate for filtered out rows, we compute everything on worker & filter them out on the coordinator

When we push down grouping, we assume the aggregates will exist on workers

`select count(distinct t) from t` fails to be properly pushed down due to type errors (row type of rows from shard tables are different since they're from a different table)

Users can disable pulling up to coordinator with `SET citus.coordinator_aggregation_strategy to 'disabled'`

### Are there relevant blog posts or outside documentation about the concept/feature?

No

### Link to relevant commits and regression tests if applicable

Pushing down grouping: https://github.com/citusdata/citus/pull/3307

[...]

PR #921, opened 2020-04-14
More aggregates supported in some situations

jonels-msft:
Fixes https://github.com/citusdata/citus_docs/issues/894

### DIFF
--- a/develop/reference_sql.rst
+++ b/develop/reference_sql.rst
@@ -12,7 +12,60 @@ In the following sections, we discuss the different types of queries you can run
 Aggregate Functions
 -------------------
 
-Citus supports and parallelizes most aggregate functions supported by PostgreSQL. Citus's query planner transforms the aggregate into its commutative and associative form so it can be parallelized. In this process, the workers run an aggregation query on the shards and the coordinator then combines the results from the workers to produce the final output.
+Citus supports and parallelizes most aggregate functions supported by
+PostgreSQL, including custom user-defined aggregates. Aggregates execute using
+one of three methods, in this order of preference:
+
+1. When the aggregate is grouped by a table's distribution column, Citus can
+   push down execution of the entire query to each worker. All aggregates are
+   supported in this situation and execute in parallel on the worker nodes.
+   (Any custom aggregates being used must be installed on the workers.)
+2. When the aggregate is *not* grouped by a table's distribution column, Citus
+   can still optimize on a case-by-case basis. Citus has internal rules for
+   certain aggregates like sum(), avg(), and count(distinct) that allows it to
+   rewrite queries for *partial aggregation* on workers. For instance, to
+   calculate an average, Citus obtains a sum and a count from each worker,
+   and then the coordinator node computes the final average.
+
+   Full list of the special-case aggregates:
+
+     avg, min, max, sum, count, array_agg, jsonb_agg, jsonb_object_agg,
+     json_agg, json_object_agg, bit_and, bit_or, bool_and, bool_or,
+     every, hll_add_agg, hll_union_agg, topn_add_agg, topn_union_agg,
+     any_value, var_pop(float4), var_pop(float8), var_samp(float4),
+     var_samp(float8), variance(float4), variance(float8) stddev_pop(float4),
+     stddev_pop(float8), stddev_samp(float4), stddev_samp(float8)
+     stddev(float4), stddev(float8)
+
+3. Last resort: pull all rows from the workers and perform the aggregation on
+   the coordinator node. When the aggregate is not grouped on a distribution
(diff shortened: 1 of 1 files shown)


develop/reference_sql.rst
```diff
@@ -12,7 +12,46 @@ In the following sections, we discuss the different types of queries you can run
...
+results from the workers to produce the final output.
+
+Aggregate execution works in two ways:
+
+1. When the aggregate is grouped by a table's distribution column, and there
+   are no aggregates on distinct values of non-distribution columns, Citus can
+   push down execution of the entire query to each worker. All PostgreSQL
+   aggregates and custom aggregates are supported in this situation, and execute
+   in parallel on the worker nodes.
+2. Otherwise, if grouping on a non-distribution column, or aggregating on a
```
serprex, line 28: This is an incomplete set of when the coordinator has to get involved. It's also done for aggregates such as `array_agg(col ORDER BY ...)`, & aggregates we either don't explicitly support or don't support with COMBINEFUNC (not exactly 1 parameter, not pseudotype `STYPE`)

jonels-msft: Is `array_agg` the culprit in your example, or the fact that it's operating on an ordered column?

Can you explain more about the reasoning behind COMBINEFUNC restrictions?

I'm looking for the general idea behind these cases. I can present that idea, and then say "for example ... [cases]"

serprex: The fact that it's operating on an ordered column is what prevents distribution

https://github.com/citusdata/citus/pull/2921 is the PR for COMBINEFUNC. Single argument is used because we have helper functions which have to approximate the signature of the aggregate. Supporting multiple arguments requires implementing generation of the helper functions at runtime, which we decided to not do. Rejecting pseudotype transition state is done because that simplifies serialize/deserialize. We can't really be sure that an `INTERNAL` serialized format is safe to send to a different db (it may contain OIDs etc)

jonels-msft: I updated scenario two. Is my description accurate enough at a high level?

develop/reference_sql.rst
```diff
@@ -12,7 +12,48 @@ In the following sections, we discuss the different types of queries you can run
...
-Citus supports and parallelizes most aggregate functions supported by PostgreSQL. Citus's query planner transforms the aggregate into its commutative and associative form so it can be parallelized. In this process, the workers run an aggregation query on the shards and the coordinator then combines the results from the workers to produce the final output.
+Citus supports and parallelizes most aggregate functions supported by
+PostgreSQL. Citus's query planner transforms the aggregate into its commutative
+and associative form so it can be parallelized. In this process, the workers
+run an aggregation query on the shards and the coordinator then combines the
+results from the workers to produce the final output.
+
+Aggregate execution works in two ways:
+
+1. When the aggregate is grouped by a table's distribution column, and there
```
serprex, line 23: > and there are no aggregates on distinct values of non-distribution columns,

This can be removed. When grouping by the distribution column, aggregates can always be pushed down

develop/reference_sql.rst
serprex: Second scenario seems well covered. Skips over COMBINEFUNC, but that's somewhat being addressed with "Citus supports and parallelizes most aggregate functions supported by PostgreSQL"

I think the mix up in this is that we actually have 3 ways of executing aggregates, 2 of which are new. The two new ways are what you've outlined. The third is partial aggregation, which is what the paragraph before this list is getting at

jonels-msft:
@serprex what do you think about this rewrite?

Should we also include a full list of Citus' "special-case" aggregates? Can you give me that list?

serprex:
This looks good

The aggregates which we have specialized support for are here: https://github.com/citusdata/citus/blob/master/src/include/distributed/multi_logical_optimizer.h#L133-L142

The builtin aggregates which fall under COMBINEFUNC are listed here: https://github.com/citusdata/citus/pull/2921#issuecomment-550061292

(Note that we prefer specialized support to COMBINEFUNC)

develop/reference_sql.rst
```diff
@@ -26,6 +26,17 @@ one of three methods, in this order of preference:
    rewrite queries for *partial aggregation* on workers. For instance, to
    calculate an average, Citus obtains a sum and a count from each worker,
    and then the coordinator node computes the final average.
+
+   Full list of the special-case aggregates:
+
+     invalid, avg, min, max, sum, count, array_agg, jsonb_agg,
```
serprex, line 32: Not `invalid`, that's a bit of a silly string, I actually had to make a PR https://github.com/citusdata/citus/pull/2957 to fix us trying to distribute aggregates named `invalid` in release builds

PR #925, opened 2020-04-20
Remove window function restrictions

### LINKED ISSUE #914: General window functions merged for 9.3
SQL support. We already supported window functions when `PARTITION BY partition_key`, but now we support arbitrary `PARTITION BY`

Generally really useful for reporting

[FAQ can remove window functions as unsupported](https://github.com/citusdata/citus_docs/blob/fb90fb7880fccb1e762226e8bd0d572656e946f2/faq/faq.rst)

## How does this work? (devs)
With recent update to use `standard_planner` we're now able to have window function in the coordinator's plan. Window functions happen at the very end of everything, so there wasn't really a good intuitive reason to not support this

### Example sql
```
SELECT
	value_2,
	value_1,
	array_agg(value_1) OVER (PARTITION BY value_2 ORDER BY value_1 RANGE BETWEEN  UNBOUNDED PRECEDING AND CURRENT ROW),
	array_agg(value_1) OVER (PARTITION BY value_2 ORDER BY value_1 RANGE BETWEEN  UNBOUNDED PRECEDING AND CURRENT ROW EXCLUDE CURRENT ROW)
FROM
	users_table
WHERE
	value_2 > 2 AND value_2 < 6
ORDER BY
	value_2, value_1, 3, 4;
```

Here's a random article Craig wrote about window functions: https://dzone.com/articles/fun-with-sql-window-functions-in-postgres

Postgres documentation is good: https://www.postgresql.org/docs/current/tutorial-window.html

https://github.com/citusdata/citus/pull/3537

jonels-msft:
Fixes #914

### DIFF
--- a/develop/reference_workarounds.rst
+++ b/develop/reference_workarounds.rst
@@ -69,18 +69,30 @@ Remember that the coordinator will send the results in the subquery or CTE to al
 Temp Tables: the Workaround of Last Resort
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-There are still a few queries that are :ref:`unsupported <unsupported>` even with the use of push-pull execution via subqueries. One of them is running window functions that partition by a non-distribution column.
+There are still a few queries that are :ref:`unsupported <unsupported>` even with the use of push-pull execution via subqueries. One of them is using `grouping sets <https://www.postgresql.org/docs/current/queries-table-expressions.html#QUERIES-GROUPING-SETS>`_ on a distributed table.
 
-In our :ref:`real-time analytics tutorial <real_time_analytics_tutorial>` we created a table called :code:`github_events`, distributed by the column :code:`user_id`. If we want to query it and count the events for a preselected of repos, one way to do it in SQL is with a window function. However, partitioning on a non-distribution column does not work:
+In our :ref:`real-time analytics tutorial <real_time_analytics_tutorial>` we
+created a table called :code:`github_events`, distributed by the column
+:code:`user_id`. Let's query it and find the earliest events for a preselected
+set of repos, grouped by combinations of event type and event publicity. A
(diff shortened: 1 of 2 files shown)


jonels-msft:
@serprex we use a window function as an example where we need to use a temp table as a workaround:
https://docs.citusdata.com/en/v9.2/develop/reference_workarounds.html#temp-tables-the-workaround-of-last-resort

What would be a good replacement query for this example?

serprex:
@jonels-msft hard to say. First thought I have is one could make two temp tables to perform a cartesian join. We may be at a point where that part of the documentation can be removed, as any temp table requiring situation would involve creating multiple temp tables to join

jonels-msft:
@onderkalaci suggested recursive CTEs for an unsupported query. Do either of you know a nice example, and how we would fix it with the temp table workaround?

serprex:
```
-- this errors
select repo_id, event_type, event_public, grouping(event_type, event_public), min(created_at)
from github_events
WHERE repo_id IN (8514, 15435, 19438, 21692)
GROUP BY event_id, rollup(event_type, event_public);

CREATE TEMP TABLE results AS (
  select repo_id,  event_type, event_public, created_at
    FROM github_events
       WHERE repo_id IN (8514, 15435, 19438, 21692)
    );

select repo_id, event_type, event_public, grouping(event_type, event_public), min(created_at)
from results
WHERE repo_id IN (8514, 15435, 19438, 21692)
GROUP BY repo_id, rollup(event_type, event_public);
```

I decided to go with `ROLLUP` / `GROUPING` instead since it seemed more practical, recursive CTEs get weird & the temp table will really vary depending on the use case

onderkalaci:
> @onderkalaci suggested recursive CTEs for an unsupported query. Do either of you know a nice example, and how we would fix it with the temp table workaround?

Just a clarification, I also suggested that GROUPING SET is a better example. The example that @serprex shared makes a lot of sense.

PR #928, opened 2020-04-22
Some updates on docs (Real-time executor, role prop., extension prop.)

onurctirtir:
(cc: @halilozanakgul & @hanefi: if they have any more updates or ideas on role propagation changes I made in docs 208e690)
(cc: @onderkalaci: if there is a better suggestion for changes I made in a84e8bb)


performance/performance_tuning.rst
```diff
@@ -241,14 +241,14 @@ There are two configuration parameters which relate to the format in which inter
...
 
-Real Time Executor
+Adaptive Executor
 -------------------------------
 
-If you have SELECT queries which require sub-second response times, you should try to use the real-time executor.
+If you have SELECT queries which require sub-second response times, you should use the adaptive executor.
 
-The real-time executor opens one connection and uses two file descriptors per unpruned shard (Unrelated shards are pruned away during planning). Due to this, the executor may need to open more connections than max_connections or use more file descriptors than max_files_per_process if the query hits a high number of shards.
+The adaptive executor opens one connection and uses two file descriptors per unpruned shard (Unrelated shards are pruned away during planning). Due to this, the executor may need to open more connections than max_connections or use more file descriptors than max_files_per_process if the query hits a high number of shards.
```
onderkalaci, line 249: I'll re-write this part, so please hold on merge for 1 day

faq/faq.rst
```diff
@@ -107,7 +107,7 @@ How do I create database roles, functions, extensions etc in a Citus cluster?
 
 Certain commands, when run on the coordinator node, do not get propagated to the workers:
 
-* ``CREATE ROLE/USER``
+* ``CREATE USER``
```
onurctirtir, line 110:
```suggestion
* ``CREATE ROLE/USER (gets propagated in Citus Enterprise)``
```

onurctirtir: @halilozanakgul how about this one ?

hanefi, line 110: I think we should keep an item for ROLE in here, in case the user does not know enough about the. difference between role and user

onurctirtir: I don't think it is the document's responsibility to identify difference between role and user, as we are generally assuming the reader has enough knowledge on PostgreSQL already ?

performance/performance_tuning.rst
```diff
@@ -181,7 +181,7 @@ General
...
 
-Citus has two executor types for running SELECT queries. The desired executor can be selected by setting the citus.task_executor_type configuration parameter. If your use case mainly requires simple key-value lookups or requires sub-second responses to aggregations and joins, you can choose the real-time executor. On the other hand if there are long running queries which require repartitioning and shuffling of data across the workers, then you can switch to the task tracker executor.
+Citus has two executor types for running SELECT queries. The desired executor can be selected by setting the citus.task_executor_type configuration parameter. If your use case mainly requires simple key-value lookups or requires sub-second responses to aggregations and joins, you can choose the adaptive executor. On the other hand if there are long running queries which require repartitioning and shuffling of data across the workers, then you can switch to the task tracker executor.
```
onderkalaci, line 184: I  suggest we  use  the following instead of your suggestion:

Citus has two executor types for running SELECT queries. The desired executor can be selected by setting the citus.task_executor_type configuration parameter. In general, adaptive executor is the default one and it can handle all different workloads very efficiently. The task-tracker executor is designed to execute long running queries which require repartitioning and shuffling of data across the workers. The only advantage of using task-tracker is its more advanced failure handling for tables with citus.replication_factor > 1. As long as the workload doesn't specifically require task-tracker, we advice to use adaptive executor.

performance/performance_tuning.rst
onderkalaci: I suggest the following (please check the grammer/wording:)


Adaptive Executor
Adaptive executor is the default executor, and it can handle all workloads very efficiently.

The adaptive executor is smart enough to use as less connections as possible to utilize minimal resources on the worker nodes for multi-shard queries. The executor starts using a single connection to a remote node, which is almost always cached so it doesn't incur any additional overhead. If the query execution finishes in `citus.executor_slow_start_interval`, we're done, no more  connections are opened. Else, the executor keeps establishing new connections in every `citus.executor_slow_start_interval`.

With the behaviour explained above, the executor aims to open optimal number of connections to remote nodes. For short running multi-shard queries, like an index-only-scan on the shards, the executor may use only a single connection and execute all the queries on the shards over a single connection. For longer running multi-shard queries, the executor will keep opening connections to parallelize the execution on the remote nodes. If the queries on the shards take long (such as > 500ms), the executor converges to using one connection per shard (or up to `citus.max_adaptive_executor_pool_size`), in order to maximize the parallelism.

faq/faq.rst
hanefi: If we were to choose one of the two, I suggest we go with `ROLE` because 
(a) PostgreSQL docs state that

> CREATE USER is now an alias for CREATE ROLE

(b) The `CREATE ROLE` statement is in the SQL standard,  whereas the `CREATE USER` statement is a PostgreSQL extension

jonels-msft:
@onurctirtir do you want to incorporate Onder's suggestions and then I'll do a final review?

onurctirtir:
> @onurctirtir do you want to incorporate Onder's suggestions and then I'll do a final review?

@jonels-msft Now included Onder's suggestions as well, now ready for final review

performance/performance_tuning.rst
```diff
@@ -181,9 +181,7 @@ General
...
-Other than the above, there are two configuration parameters which can be useful in cases where approximations produce meaningful results. These two parameters are citus.limit_clause_row_fetch_count and citus.count_distinct_error_rate. The former sets the number of rows to fetch from each task while calculating limits while the latter sets the desired error rate when calculating approximate distinct counts. You can learn more about the applicability and usage of these parameters in the user guide sections: :ref:`count_distinct` and :ref:`limit_pushdown`.
+Citus has two executor types for running SELECT queries. The desired executor can be selected by setting the citus.task_executor_type configuration parameter. In general, adaptive executor is the default one and it can handle all different workloads very efficiently. The task-tracker executor is designed to execute long running queries which require repartitioning and shuffling of data across the workers. The only advantage of using task-tracker is its more advanced failure handling for tables with citus.replication_factor > 1. As long as the workload doesn't specifically require task-tracker, we advice to use adaptive executor.
```
jonels-msft, line 184:
```diff
- In general, adaptive executor is the default one and it can handle all different workloads very efficiently.
+ The adaptive executor is the default, and can handle most workloads very efficiently.
```

```diff
- we advice to use
+ we advise using
```

performance/performance_tuning.rst
```diff
@@ -241,14 +239,14 @@ There are two configuration parameters which relate to the format in which inter
 
 However, for certain data types like hll or hstore arrays, the cost of serializing and deserializing data is pretty high. In such cases, using binary format for transferring intermediate data can improve query performance due to reduced CPU usage. There are two configuration parameters which can be used to tune this behaviour, citus.binary_master_copy_format and citus.binary_worker_copy_format. Enabling the former uses binary format to transfer intermediate query results from the workers to the coordinator while the latter is useful in queries which require dynamic shuffling of intermediate data between workers.
 
-Real Time Executor
+Adaptive Executor
```
jonels-msft, line 242: Just a question: is the real-time executor entirely removed now, or merely deprecated?

performance/performance_tuning.rst
```diff
@@ -241,14 +239,14 @@ There are two configuration parameters which relate to the format in which inter
...
 -------------------------------
 
-If you have SELECT queries which require sub-second response times, you should try to use the real-time executor.
+Adaptive executor is the default executor, and it can handle all workloads very efficiently.
 
-The real-time executor opens one connection and uses two file descriptors per unpruned shard (Unrelated shards are pruned away during planning). Due to this, the executor may need to open more connections than max_connections or use more file descriptors than max_files_per_process if the query hits a high number of shards.
+The adaptive executor is smart enough to use as less connections as possible to utilize minimal resources on the worker nodes for multi-shard queries. The executor starts using a single connection to a remote node, which is almost always cached so it doesn't incur any additional overhead. If the query execution finishes in citus.executor_slow_start_interval, we're done, no more connections are opened. Else, the executor keeps establishing new connections in every citus.executor_slow_start_interval.
```
jonels-msft, line 247: How about this:

The adaptive executor conserves database connections to help reduce resource usage on worker nodes during multi-shard queries. When running a query, the executor begins by using a single -- usually precached -- connection to a remote node. If the query finishes within the time period specified by ``citus.executor_slow_start_interval``, no more connections are required.  Otherwise the executor gradually establishes new connections as directed by ``citus.executor_slow_start_interval``.

use_cases/multi_tenant.rst
```diff
@@ -501,6 +501,11 @@ Next we move the data across the network to a new dedicated node. Create a new n
   -- move the shard to your choice of worker (it will also move the
   -- other shards created with the CASCADE option)
 
+  -- note that you should set wal_level for all nodes to be >= logical
```
jonels-msft, line 504: Should we include the steps to do this as part of our instructions? This comment has a lot packed inside.

performance/performance_tuning.rst
jonels-msft: Also we could link to the [adaptive executor configuration](https://docs.citusdata.com/en/v9.2/develop/api_guc.html#adaptive-executor-configuration) section that has lots of details.

use_cases/multi_tenant.rst
onurctirtir: It's actually a requirement for `master_move_shard_placement` but I couldn't decide which one is better (comment or instruction), your call :)

performance/performance_tuning.rst
onurctirtir: afaik removed

onurctirtir: but configuration is still there for backward compatibility

performance/performance_tuning.rst
onurctirtir: done

performance/performance_tuning.rst
onurctirtir: done

jonels-msft:
I changed my review to "approve"

After you merge this PR I'll touch up the adaptive executor link to give it a restructuredtext target.

PR #931, opened 2020-04-27
Update broken message templates

onlined:
Due to changes in message template escaping logic of CloudPlane, message templates were broken. I updated them to be compatible with both old and new logic.


onlined:
I did a test with Loggly (it was successful), but I'm sure the other ones are correct too.

PR #932, opened 2020-04-28
Truncating local records

### LINKED ISSUE #924: Introduce truncate_local_data_after_distributing_table()
Currently we retain the local data after distributing a table. This causes a variety of issues, and can sometimes make migrations harder than needs to be. We used to have [a workaround](https://github.com/citusdata/citus/issues/2889#issue-480885263) that is best described as _obviously terrifying for users_. To remedy this we introduced a UDF that truncates local data after distributing a table.

(i) Create, populate, distribute a table.
(ii) Create a constraint that can not be validated.
(iii) Update the records so that the previous constraint should be valid.
(iv) Still fail to create the constraint as the outdated local records violate the constraint.
(v) Attempt to truncate the local and outdated records that are not easily accessible.

The workaround in (v) is not failsafe, and this UDF is a way better alternative to solve this problem.

Some issues reported by users/SEs:

- Primary key creation fails with duplicate key violation even though there are no duplicate values

  https://github.com/citusdata/citus/issues/1770

  If the outdated local records start to violate constraints, the user ends up in an undesirable place where some set of records they can not access blocks their desired schema changes.

- If we create a table with a nullable column, insert and then distribute. Changing the NOT NULL fails even after updating the columns.

	https://github.com/citusdata/citus/issues/2889

I do not have a clue about what I should put here :(

- [table_management](https://docs.citusdata.com/en/latest/admin_guide/table_management.html)
- [common errors](http://docs.citusdata.com/en/latest/reference/common_errors.html)
	- this one is not so easy, as there are various constraint types and they differ in error messages. If a user wishes to search for the exact match, it may not be easy to list all possible errors
- [reference_ddl](https://docs.citusdata.com/en/latest/develop/reference_ddl.html)

## How does this work? (devs)

This UDF runs a `TRUNCATE .. CASCADE` on the local copies provided that:

1. the parameter is a Citus table
2. there are no foreign key references to the given table

That means, we may end up truncating the local data of other distributed tables as well, but the user will not lose any data. The original data is supposed to be copied already, and the `TRUNCATE .. CASCADE` should affect only inaccessible and stale local rows.

### Example sql

```sql
SELECT truncate_local_data_after_distributing_table($$public.dist$$)
```

### Corner cases, gotchas

-

No

https://github.com/citusdata/citus/pull/3752/files has two commits:
1. https://github.com/citusdata/citus/pull/3752/commits/d535121f8de97dec3feabac84aa62131f26a52f9 introduces `truncate_local_data_after_distributing_table()`. See changes in `multi_truncate.sql` for regression tests.

2. https://github.com/citusdata/citus/pull/3752/commits/0c5d0cfee97282d93330244c50f923f28095b39c adds a notice message to help truncate local data after distribution

jonels-msft:
Fixes #924

### DIFF
--- a/develop/api_udf.rst
+++ b/develop/api_udf.rst
@@ -73,7 +73,7 @@ Truncate all local rows after distributing a table, and prevent constraints from
 
   ERROR:  cannot truncate a table referenced in a foreign key constraint by a local table
 
-Truncatating local coordinator node table data is safe for distributed tables because their rows, if they have any, are copied to worker nodes during distribution.
+Truncating local coordinator node table data is safe for distributed tables because their rows, if they have any, are copied to worker nodes during distribution.
 
 Arguments
 ************************
--- a/develop/api_udf.rst
+++ b/develop/api_udf.rst
@@ -67,7 +67,7 @@ For more examples, see :ref:`ddl`.
 truncate_local_data_after_distributing_table
 $$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$
 
-Truncate all local rows after distributing a table, and prevent constraints from failing due to outdated local records. The truncation cascades to tables having a foreign key to the designated table. If the referring tables are themselves not distributed then truncation is forbidden until they are, to protect referential integrity:
+Truncate all local rows after distributing a table, and prevent constraints from failing due to outdated local records. The truncation cascades to tables having a foreign key to the designated table. If the referring tables are not themselves distributed then truncation is forbidden until they are, to protect referential integrity:
 
 ::
 
--- a/develop/api_udf.rst
+++ b/develop/api_udf.rst
@@ -62,6 +62,37 @@ This example informs the database that the github_events table should be distrib
 
 For more examples, see :ref:`ddl`.
 
+.. _truncate_local_data_after_distributing_table:
+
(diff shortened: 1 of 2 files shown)


jonels-msft:
I added some error messages in the text, although didn't put them in the common error messages section. Should still be discoverable through docs search.

One thing I might want to do is include Metin's example of how to reproduce the problem of local constraints on stale data.

Other than that, does this look good?

develop/api_udf.rst
```diff
@@ -62,6 +62,37 @@ This example informs the database that the github_events table should be distrib
...
+truncate_local_data_after_distributing_table
+$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$
+
+Truncate all local rows after distributing a table, and prevent constraints from failing due to outdated local records. The truncation cascades to tables having a foreign key to the designated table. If the referring tables are themselves not distributed then truncation is forbidden until they are, to protect referential integrity:
+
+::
+
+  ERROR:  cannot truncate a table referenced in a foreign key constraint by a local table
+
+Truncatating local coordinator node table data is safe for distributed tables because their rows, if they have any, are copied to worker nodes during distribution.
```
hanefi, line 76: s/Truncatating/Truncating/

Issue #926, opened 2020-04-21
Introduce `citus.max_shared_pool_size`


onderkalaci:
## Why are we implementing it? (sales eng)

Prior to 9.3, running high concurrency multi-shard commands was problematic with Citus. The idea is that, a single backend on the coordinator can initiate upto `#shards/workeNoderCount` connections per worker node.

It was making *almost* impossible to have more than few dozens of concurrent multi-shard queries. For example, say shard_count = 32 and worker count = 2. In this case, each distributed execution can establish 16 connections per worker node. Say, we have 20 backends running distributed queries. This ends-up 16 * 20 = 320 connections on each worker. As this is > max_connections, some queries would fail. As you increase the concurrency, more and more queries would start to fail.

### What are the typical use cases?

Citus now can be used with higher concurrency. 

## How does this work? (devs)

We introduced a new GUC named `citus.max_shared_pool_size`.  It sets the maximum number of connections allowed per worker node **across all the backends from the coordinator**. So, basically throttle the number of connections going to per worker node.

There are two ways of throttling happening: 
- If a query execution has already established at least one connection to a remote node (meaning that it can continue executing the query over that connection), the other connections are qualified as `optional`. Meaning that, if the total connections per worker node has already reached `citus.max_shared_pool_size` across backends, the executor wouldn't establish any more connections to the node for the query execution.

- If query execution has not established any connections yet (meaning that it can NOT execute the query), the backend waits until it gets a connection. 

Setting to -1 disables connections throttling. Setting to 0 makes it auto-adjust, meaning equal to max_connections on the coordinator. The default value is `0`.

### Corner cases, gotchas

- As a rule of thumb, the value should be at most equal to the `max_connections` on the worker nodes. Because, the idea is to prevent any connection establishment failures on the workers due to reaching `max_connections`.  

- There are certain operations that does not obey `citus.max_shared_pool_size`, most importantly `COPY` and repartition joins. That's why, it might be useful to increase the `max_connections` on the workers a bit further compared to `max_connections` on the coordinator. This could give some extra space for the connections required for COPY and repartition queries on the workers.

### Link to relevant commits and regression tests if applicable

https://github.com/citusdata/citus/pull/3692

PR #933, opened 2020-04-30
max_shared_pool_size reference

jonels-msft:
Fixes https://github.com/citusdata/citus_docs/issues/926

@onderkalaci, is this enough explanation?

### DIFF
--- a/develop/api_guc.rst
+++ b/develop/api_guc.rst
@@ -470,7 +470,7 @@ each worker proportional to the number of shards it accesses (in particular,
 #shards/#workers). Running dozens of multi-shard queries at once can easily hit
 worker nodes' ``max_connections`` limit, causing queries to fail.
 
-The default value, 0, caps the connections to the coordinator's
+The default value, 0, caps the connections to the coordinator's own
 ``max_connections``, which isn't guaranteed to match that of the workers (see
 the note below). The value -1 disables throttling.
 
--- a/develop/api_guc.rst
+++ b/develop/api_guc.rst
@@ -457,6 +457,23 @@ time.
 For long queries (those taking >500ms), slow start might add latency, but for
 short queries it's faster. The default value is 10ms.
 
+citus.max_shared_pool_size (integer)
+************************************
+
+Specifies the maximum number of connections that the coordinator node, across
+all simultaneous sessions, is allowed to make per worker node. PostgreSQL must
+allocate fixed resources for every connection and this GUC helps ease
+connection pressure on workers.
+
+Without connection throttling, every multi-shard query creates connections on
+each worker proportional to the number of shards it accesses (in particular,
+#shards/#workers). Running dozens of multi-shard queries at once can easily hit
+worker nodes' ``max_connections`` limit, causing queries to fail.
+
+The default value, 0, caps the connections to the coordinator's
+``max_connections``, which isn't guaranteed to match that of the workers, but
+is a good estimate. The value -1 disables throttling.
+
 citus.max_cached_conns_per_worker (integer)
 *******************************************
 
--- a/develop/api_guc.rst
+++ b/develop/api_guc.rst
@@ -471,8 +471,16 @@ each worker proportional to the number of shards it accesses (in particular,
 worker nodes' ``max_connections`` limit, causing queries to fail.
 
 The default value, 0, caps the connections to the coordinator's
-``max_connections``, which isn't guaranteed to match that of the workers, but
-is a good estimate. The value -1 disables throttling.
+``max_connections``, which isn't guaranteed to match that of the workers (see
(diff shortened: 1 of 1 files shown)


develop/api_guc.rst
```diff
@@ -457,6 +457,31 @@ time.
...
+all simultaneous sessions, is allowed to make per worker node. PostgreSQL must
+allocate fixed resources for every connection and this GUC helps ease
+connection pressure on workers.
+
+Without connection throttling, every multi-shard query creates connections on
+each worker proportional to the number of shards it accesses (in particular,
+#shards/#workers). Running dozens of multi-shard queries at once can easily hit
+worker nodes' ``max_connections`` limit, causing queries to fail.
+
+The default value, 0, caps the connections to the coordinator's own
```
onderkalaci, line 473: Minor note:  I think we don't need to explicitly mention about the default value `0`, as that is  an internal value, the users don't  see it  at all. Because, we override the value when showing  to the users: https://github.com/citusdata/citus/blob/c6f5d5fe885b35d6ac7fb865db30222c945b3d49/src/backend/distributed/shared_library_init.c#L1595-L1612

```SQL
postgres=# show max_connections ;
┌─────────────────┐
│ max_connections │
├─────────────────┤
│ 100             │
└─────────────────┘
(1 row)

Time: 0.419 ms
postgres=# show citus.max_shared_pool_size ;
┌────────────────────────────┐
│ citus.max_shared_pool_size │
├────────────────────────────┤
│ 100                        │
└────────────────────────────┘
(1 row)

Time: 0.549 ms
postgres=# 

``` 
So, maybe  `The default value, 0, ` -> `By default, the value is automatically set equal to the coordinator's own ...`

jonels-msft:
@onderkalaci after moving this GUC to the top of its section, I noticed the next one listed is `max_adaptive_executor_pool_size` which seems to share a similar function. Assuming that one sets a limit per session, while `max_shared_pool_size` provides a limit across *all* sessions? Should these settings be used in combination, or does `max_shared_pool_size` kind of obsolete the older GUC?

onderkalaci:
>  Should these settings be used in combination, or does max_shared_pool_size kind of obsolete the older GUC?

Good point. I'd say `max_shared_pool_size` will become the main GUC that users typically interact as it provides a global limit. 

But  `max_adaptive_executor_pool_size` can still useful for:
- preventing a single backend to get all the resources (as a backend is limited to open `max_adaptive_executor_pool_size` connections at most)
- Providing resource management. You may have  some low priority sessions with low `max_adaptive_executor_pool_size` and high priority sessions with higher values

So, I think it is not deprecated at this moment.


But, we should probably update that section. See my suggestion below for `max_adaptive_executor_pool_size` docs:

> The maximum number of connections per worker node used by the adaptive executor to execute a multi-shard command.

we  can add `multi-shard command by this session.`

> The adaptive executor opens multiple connections per worker node when running multi-shard commands to parallelize the command across multiple cores on the worker. This setting specifies the maximum number of connections it will open. (The number of connections is also bounded by the number of shards on the node.)

Above section looks fine.

> Although increasing parallelism for a single query is generally good, it also puts connection pressure on the workers in a production environment. This GUC provides a knob to trade off parallelism vs connection pressure.

I think instead of the above, we should  refer to `max_shared_pool_size` by saying that the overall connection pressure can be adjusted with `max_shared_pool_size` and this is more for resource management per session. So, basically, merge/move the last sentence , starting with `This is a session-level GUC`.. , here.

> The default value is 16, however it's hard to make a good static choice manually. In the future the adaptive executor may adjust this GUC on the fly, shrinking the pool when workers are busy, and growing it when they are idle.

We should only keep `The default value is 16`, because actually `shared_pool_size` already implements this.

> This is a session-level GUC, and thus can be used to limit the resource usage for individual sessions. Lowering the value can limit the queries' use of connections, cores, or memory.

jonels-msft:
What do you think about this latest change? I shortened the description for max_adaptive_executor_pool_size, contrasting it with max_shared_pool_size. We talk about connection pressure there, so I thought we could reduce duplication.

develop/api_guc.rst
```diff
@@ -417,29 +417,46 @@ By default, an INSERT INTO … SELECT statement that cannot be pushed down will
...
+citus.max_shared_pool_size (integer)
+************************************
+
+Specifies the maximum number of connections that the coordinator node, across
+all simultaneous sessions, is allowed to make per worker node. PostgreSQL must
+allocate fixed resources for every connection and this GUC helps ease
+connection pressure on workers.
+
+Without connection throttling, every multi-shard query creates connections on
+each worker proportional to the number of shards it accesses (in particular,
```
onderkalaci, line 431: very minor: `#shards/#workers` -> `up to #shards/#workers`

PR #938, opened 2020-05-27
Add a Makefile for easy command running

JelteF:
This also changes the command that was shown in the README in two ways:
1. `-a` is removed. This makes the build quicker by doing incremental builds.
2. `-W` is added. This changes warnings into errors, so they don't
   accidentily slip in.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,5 @@
+docs:
+	sphinx-build -b html -W -n . _build
+
+clean:
+	rm -rf _build
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,10 @@
+.PHONY: docs watch clean
+
 docs:
 	sphinx-build -b html -W -n . _build
+
+watch:
+	sphinx-autobuild -b html -W -n . _build
 
 clean:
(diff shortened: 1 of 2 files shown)


jonels-msft:
Could also add a `watch` target:

```makefile
watch:
	sphinx-autobuild -b html -W -n . _build
```

As a separate observation -- the makefile is really more of a shell script because it's all "phony" targets (in GNU terminology). it would be nice if the Makefile could capture the dependency between the source and output files to determine when to rebuild things, but I guess the sphinx-build tool itself manages the incremental build.

That said, I hesitate to mark the targets with `.PHONY` either because it's not conformant with [POSIX make](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/make.html).

JelteF:
Done, I also marked them as `.PHONY` (I always forget doing that). Worst case if it's not supported and nothing bad happens. It seems to be supported by both GNU, OpenBSD and FreeBSD make, so I think it's also a really common extension.

jonels-msft:
Thank you, this will be more convenient! LGTM

PR #939, opened 2020-06-03
No longer a need to manually propagate grant select

jonels-msft:
Verified with Citus 9.2 on Hyperscale that "grant select on all tables in schema" works all by itself and does not require run_command_on_workers.


jonels-msft:
Just verified it's valid on Citus Cloud as well.

PR #943, opened 2020-06-25
Update realtime_analytics.rst

AureaMohammadAlavi:
Because the current minute (`date_trunc('minute', now())`) has not passed yet, records that are inserted in the current minute but after the function execution will never be aggregated into `http_request_1min` table. We need to assign one minute before the current minute (`date_trunc('minute', now() - interval '1 minute')`) to `curr_rollup_time`.


jonels-msft:
Thank you!

PR #945, opened 2020-07-15
update vpc peering section to reflect inter-region vpc support limitation

### LINKED ISSUE #944: inter-region vpc peering in citus cloud is not supported
Docs should raise the caveat that vpc peering is only supported with same region vpcs for clarity to customers. Depending on how we decide to handle [the inter-region vpc peering](https://github.com/citusdata/cloudplane/issues/1538) we may need to change again.

kileri:
added a note on inter-region vpc peering support status. closes #944


jonels-msft:
Thanks!

PR #946, opened 2020-07-22
Document recent EXPLAIN changes.

pykello:
This includes:
1. Query texts are logged when the VERBOSE flag is specified.
2. EXPLAIN ANALYZE contains statistics about intermediate result files.


jonels-msft:
@pykello thanks a lot!

I scanned to see if we have other examples of verbose or analyze that need updating, but it looks good. Merging!

PR #955, opened 2020-09-11
pg_auto_failover in enterprise install

jonels-msft:


jonels-msft:
Fixes https://github.com/citusdata/hyperscale-planning/issues/22

Issue #953, opened 2020-09-04
Instructions for creating a new database are slightly misleading


marcocitus:
Citus requires nodes to be up and have the Citus extension installed at master_add_node time such that reference tables, types and roles can be replicated during master_add_node and tables can be created immediately after. That requirement is now enforced even if the cluster is empty.

The current instructions for CREATE DATABASE seem to suggest that you can do the CREATE EXTENSION citus on the workers after adding the nodes, which is causing some confusion.
http://docs.citusdata.com/en/v9.4/admin_guide/cluster_management.html#creating-a-new-database

It might better to simply state that on all workers you run:
```sql
CREATE DATABASE newbie;
\c newbie
CREATE EXTENSION citus;
```

and on the coordinator you run:
```sql
CREATE DATABASE newbie;
\c newbie
CREATE EXTENSION citus;
SELECT * from master_add_node('node-name', 5432);
SELECT * from master_add_node('node-name2', 5432);
```

This was raised by @ivyazmitinov in https://github.com/citusdata/citus/issues/3879

ivyazmitinov:
I'd also add that setting `citus.enable_alter_role_propagation = off` is required when a password of user differs on the coordinator and workers nodes. It is a common use-case for a security based on `.pgpass` file, as described [here](https://docs.citusdata.com/en/v9.4/admin_guide/cluster_management.html?highlight=.pgpass#increasing-worker-security).

jonels-msft:
Thanks for reporting. We'll get it prioritized

PR #956, opened 2020-09-11
Change order of commands to create db

jonels-msft:
Fixes https://github.com/citusdata/citus_docs/issues/953


jonels-msft:
I tested this and it works great.

PR #970, opened 2021-01-11
Docs updates on range distr. and copy from worker in append distr.

naisila:
DESCRIPTION: Docs updates on range distr. and copy from worker in append distr.

I provided solutions to some problems I mentioned in release 9.5 doc testing from _Quick Tutorials_, _Develop_ and _FAQ_ sections:

* We don't support COPY from worker nodes on append-distributed tables anymore, so I removed those sections. See https://github.com/citusdata/citus/pull/3261

* As also discussed with @onderkalaci, we don't explain range-based distribution on the docs, so it's better to remove all mentions of range-distributed tables on the docs.

* Lastly, `node_capacity_function` was erroring out due to type mismatch. I simply changed the return type from `boolean` to `int`.
```SQL
-- example of node_capacity_function

CREATE FUNCTION v2_node_double_capacity(nodeidarg int)
    RETURNS boolean AS $$
    SELECT
        (CASE WHEN nodename LIKE '%.v2.worker.citusdata.com' THEN 2 ELSE 1 END)
    FROM pg_dist_node where nodeid = nodeidarg
    $$ LANGUAGE sql;
```
```
ERROR:  return type mismatch in function declared to return boolean
DETAIL:  Actual return type is integer.
CONTEXT:  SQL function "v2_node_double_capacity"
```

### DIFF
--- a/develop/api_metadata.rst
+++ b/develop/api_metadata.rst
@@ -436,9 +436,9 @@ Here are examples of functions that can be used within new shard rebalancer stra
       -- example of node_capacity_function
 
       CREATE FUNCTION v2_node_double_capacity(nodeidarg int)
-          RETURNS int AS $$
+          RETURNS real AS $$
           SELECT
-              (CASE WHEN nodename LIKE '%.v2.worker.citusdata.com' THEN 2 ELSE 1 END)
+              (CASE WHEN nodename LIKE '%.v2.worker.citusdata.com' THEN 2.0::float4 ELSE 1.0::float4 END)
           FROM pg_dist_node where nodeid = nodeidarg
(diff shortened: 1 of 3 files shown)


jonels-msft (approved):
Thanks for the fixes, and catching the range distribution language I overlooked in 8feb7d6a6f9fb2c257dff946c62a6dc715a1a3c7

Assuming you tested the new v2_node_double_capacity, the changes look good to me, feel free to squash and merge when ready.

naisila:
I tested `v2_node_double_capacity`, it should return `real` type instead of `int`. I am changing the function accordingly.

PR #976, opened 2021-02-04
Add explain analyze sort method

agedemenli:
Adds the new configuration variable `explain_analyze_sort_method` to docs

### DIFF
--- a/develop/api_guc.rst
+++ b/develop/api_guc.rst
@@ -472,3 +472,17 @@ citus.explain_all_tasks (boolean)
 ************************************************
 
 By default, Citus shows the output of a single, arbitrary task when running `EXPLAIN <http://www.postgresql.org/docs/current/static/sql-explain.html>`_ on a distributed query. In most cases, the explain output will be similar across tasks. Occasionally, some of the tasks will be planned differently or have much higher execution times. In those cases, it can be useful to enable this parameter, after which the EXPLAIN output will include all tasks. This may cause the EXPLAIN to take longer.
+
+.. _explain_analyze_sort_method:
+
+citus.explain_analyze_sort_method (enum)
+************************************************
+
+Determines the sort method of the tasks in the output of EXPLAIN ANALYZE.
+The default value of citus.explain_analyze_sort_method is ``execution-time``.
+
+The supported values are:
+
+* **execution-time:** sort by execution time.
+
+* **taskId:** sort by task id.
--- a/performance/performance_tuning.rst
+++ b/performance/performance_tuning.rst
@@ -152,6 +152,9 @@ Differences in worker execution can be caused by tuning configuration difference
 
   Note that when citus.explain_all_tasks is enabled, EXPLAIN plans are retrieved sequentially, which may take a long time for EXPLAIN ANALYZE.
 
+Citus, by default, sorts tasks by execution time in descending order. If `citus.explain_all_tasks` is disabled, then Citus shows the single longest-running task. Please note that this functionality can be used only with EXPLAIN ANALYZE, since regular EXPLAIN doesn't execute the queries, and therefore doesn't know any execution times.
+To change the sort order, you can use :ref:`explain_analyze_sort_method`.
+
 .. _scaling_out_performance:
 
 Scaling Out Performance
--- a/use_cases/multi_tenant.rst
+++ b/use_cases/multi_tenant.rst
@@ -11,7 +11,7 @@ Multi-tenant Applications
 
 If you're building a Software-as-a-service (SaaS) application, you probably already have the notion of tenancy built into your data model. Typically, most information relates to tenants / customers / accounts and the database tables capture this natural relation.
 
-For SaaS applications, each tenant's data can be stored together in a single database instance and kept isolated from and invisible to other tenants. This is efficient in three ways. First application improvements apply to all clients. Second, sharing a database between tenants uses hardware efficiently. Last, it is much simpler to manage a single database for all tenants than a different database server for each tenant.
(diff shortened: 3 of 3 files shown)


jonels-msft:
Thanks! Could you also add a reference section about the GUC in develop/api_guc.rst ?

performance/performance_tuning.rst
```diff
@@ -152,6 +152,14 @@ Differences in worker execution can be caused by tuning configuration difference
 
   Note that when citus.explain_all_tasks is enabled, EXPLAIN plans are retrieved sequentially, which may take a long time for EXPLAIN ANALYZE.
 
+For troubleshooting and detecting the bottlenecks, you can use this configuration variable:
+
+.. code-block:: postgresql
+
+  SET citus.explain_analyze_sort_method = 'execution-time';
+
+This will allow you to see all tasks sorted by execution time, in descending order, if ``citus.explain_all_tasks`` is on. i.e You can see the longest running task on top. If ``citus.explain_all_tasks`` is not enabled, then you will see only the longest running task. Please note that this functionality can only be used with ``EXPLAIN ANALYZE``, since ``EXPLAIN`` itself doesn't execute the queries, and therefore doesn't have any execution time. The default value of ``citus.explain_analyze_sort_method`` is ``execution-time``. Also, running ``SET citus.explain_analyze_sort_method = 'taskId';`` will allow you to get the output sorted by task ids.
```
naisila, line 161: Nice GUC :)
Maybe after adding a reference section about the GUC in develop/api_guc.rst as @jonels-msft suggested, the last two sentences can be removed from here and be a part of the reference section.
We can also omit the quotation marks around citus.explain_all_tasks, EXPLAIN and EXPLAIN ANALYZE, as they are usually omitted in other paragraphs as well.

develop/api_guc.rst
```diff
@@ -472,3 +472,15 @@ citus.explain_all_tasks (boolean)
 ************************************************
 
 By default, Citus shows the output of a single, arbitrary task when running `EXPLAIN <http://www.postgresql.org/docs/current/static/sql-explain.html>`_ on a distributed query. In most cases, the explain output will be similar across tasks. Occasionally, some of the tasks will be planned differently or have much higher execution times. In those cases, it can be useful to enable this parameter, after which the EXPLAIN output will include all tasks. This may cause the EXPLAIN to take longer.
+
+citus.explain_analyze_sort_method (enum)
+************************************************
+
+Sorts the tasks in the output of EXPLAIN ANALYZE by execution time.
```
naisila, line 479: How about 
Determines the sort method of the tasks in the output of EXPLAIN ANALYZE.

develop/api_guc.rst
```diff
@@ -472,3 +472,15 @@ citus.explain_all_tasks (boolean)
...
+The default value of citus.explain_analyze_sort_method is ``execution-time``. Also, running ``SET citus.explain_analyze_sort_method = 'taskId';`` will allow you to get the output sorted by task ids.
```
naisila, line 480: We can remove the second sentence since you are also explaining it in the supported values

develop/api_guc.rst
agedemenli: better. thank you

naisila (approved):

performance/performance_tuning.rst
```diff
@@ -152,6 +152,14 @@ Differences in worker execution can be caused by tuning configuration difference
 
   Note that when citus.explain_all_tasks is enabled, EXPLAIN plans are retrieved sequentially, which may take a long time for EXPLAIN ANALYZE.
 
+For troubleshooting and detecting the bottlenecks, you can use this configuration variable:
+
+.. code-block:: postgresql
+
+  SET citus.explain_analyze_sort_method = 'execution-time';
```
jonels-msft, line 159: The reference section (and Citus source) says that execution-time is the default value, so do we need to tell people to set the GUC this way? Perhaps we simply mention that the results of explain analyze will be sorted by the length of their execution time, and that the sort order is configurable (linking to the reference section).

develop/api_guc.rst
```diff
@@ -472,3 +472,15 @@ citus.explain_all_tasks (boolean)
...
+************************************************
+
+Determines the sort method of the tasks in the output of EXPLAIN ANALYZE.
+The default value of citus.explain_analyze_sort_method is ``execution-time``.
+
+The supported values are:
+
+* **execution-time:** sort by execution time.
+
+* **taskId:** sort by task id.
```
jonels-msft, line 486: Weird that one is hyphenated and the other is camel case. I guess it's minor, but should that be updated in Citus before release?

jonels-msft (changes requested):
Added a couple questions inline with the code.

jonels-msft (changes requested):
Thanks, almost there. Added a few comments.

performance/performance_tuning.rst
```diff
@@ -152,6 +152,9 @@ Differences in worker execution can be caused by tuning configuration difference
 
   Note that when citus.explain_all_tasks is enabled, EXPLAIN plans are retrieved sequentially, which may take a long time for EXPLAIN ANALYZE.
 
+For troubleshooting and detecting the bottlenecks, Citus, by default, allows you to see all tasks sorted by execution time, in descending order, if citus.explain_all_tasks is on. i.e You can see the longest running task on top. If citus.explain_all_tasks is disabled, then you will see only the longest running task. Please note that this functionality can only be used with EXPLAIN ANALYZE, since EXPLAIN itself doesn't execute the queries, and therefore doesn't have any execution time.
```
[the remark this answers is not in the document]
agedemenli: better. thank you

jonels-msft:
Thank you!

PR #977, opened 2021-02-04
Docs updates from release test 10.0

naisila:
DESCRIPTION: Docs updates after release testing for Citus 10.0

I tested the _Use-Case Guides_, _Administer_ and _Troubleshoot_ sections. 
(I went out of these sections only for the `However` to `However,` changes, with a find-replace in the whole repo)

Mainly some outdated outputs, partitioned table description, some small typos.

I also put the UTF8 encoding warning since I experienced issues with my database setup because it had SQL_ASCII encoding.

### DIFF
--- a/admin_guide/cluster_management.rst
+++ b/admin_guide/cluster_management.rst
@@ -539,10 +539,10 @@ After this INSERT, any query needing to connect to node 123 as the user jdoe wil
 
 .. code-block:: postgresql
 
-  -- update user jdoe to use certificate authenticaion
+  -- update user jdoe to use certificate authentication
   UPDATE pg_dist_authinfo
-  WHERE nodeid = 123 AND rolename = 'jdoe'
-  SET authinfo = 'sslcert=/path/to/user.crt sslkey=/path/to/user.key';
+  SET authinfo = 'sslcert=/path/to/user.crt sslkey=/path/to/user.key'
+  WHERE nodeid = 123 AND rolename = 'jdoe';
 
 This changes the user from using a password to use a certificate and keyfile while connecting to node 123 instead. Make sure the user certificate is signed by a certificate that is trusted by the worker you are connecting to and authentication settings on the worker allow for certificate based authentication. Full documentation on how to use client certificates can be found in `the postgres libpq documentation <https://www.postgresql.org/docs/current/libpq-ssl.html#LIBPQ-SSL-CLIENTCERT>`_.
 
--- a/admin_guide/cluster_management.rst
+++ b/admin_guide/cluster_management.rst
@@ -32,14 +32,14 @@ Real-Time Analytics Use-Case
 
 In the :ref:`rt_blurb` use-case, shard count should be related to the total number of cores on the workers. To ensure maximum parallelism, you should create enough shards on each node such that there is at least one shard per CPU core. We typically recommend creating a high number of initial shards, e.g. **2x or 4x the number of current CPU cores**. This allows for future scaling if you add more workers and CPU cores.
 
-However keep in mind that for each query Citus opens one database connection per shard, and these connections are limited. Be careful to keep the shard count small enough that distributed queries won't often have to wait for a connection. Put another way, the connections needed, ``(max concurrent queries * shard count)``, should generally not exceed the total connections possible in the system, ``(number of workers * max_connections per worker)``.
+However, keep in mind that for each query Citus opens one database connection per shard, and these connections are limited. Be careful to keep the shard count small enough that distributed queries won't often have to wait for a connection. Put another way, the connections needed, ``(max concurrent queries * shard count)``, should generally not exceed the total connections possible in the system, ``(number of workers * max_connections per worker)``.
 
 .. _prod_size:
 
 Initial Hardware Size
 =====================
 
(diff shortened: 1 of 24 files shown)


agedemenli (approved):
LGTM

admin_guide/cluster_management.rst
```diff
@@ -410,7 +410,6 @@ So far the information in this view doesn't give us anything we couldn't already
 
 ::
 
-  .
```
jonels-msft, line 413: The reason this dot is here is to set the indentation level of the code block. If the dot isn't present, Sphinx will consider the first line to be the one starting with `   sum` and will count three spaces of indentation. The horizontal line of dashes is less indented, however, and would end the code block.

The dot thing is a hack, no doubt about that. I just don't know a better way. Maybe you can think of one?

admin_guide/table_management.rst
```diff
@@ -165,7 +165,6 @@ contain the same data, and compare:
 
 ::
 
-  .
```
jonels-msft, line 168: The reason this dot is here is to set the indentation level of the code block. If the dot isn't present, Sphinx will consider the first line to be the one starting with `   row_size` and will count three spaces of indentation. The horizontal line of dashes is less indented, however, and would end the code block.

The dot thing is a hack, no doubt about that. I just don't know a better way. Maybe you can think of one?

articles/hll_count_distinct.rst
```diff
@@ -62,7 +62,7 @@ Basically, to be able to observe k leading zeros, we need 2k+1 trials
 k in a data set, HLL concludes that there are approximately 2k+1
 distinct elements.
 
-This is pretty straightforward and simple estimation method. However; it
+This is pretty straightforward and simple estimation method. However, it
```
jonels-msft, line 65: Note that files under `articles/` are verbatim copies of blog posts that management wanted to put here to keep visible longer. There may be more stylistic problems lurking. I haven't changed them beyond rewriting as reStructuredText.

reference/common_errors.rst
```diff
@@ -24,16 +14,17 @@ Caused when the the coordinator node is unable to connect to a worker.
...
+          TCP/IP connections on port 5432?
+
 
 Resolution
 ~~~~~~~~~~
 
 To fix, check that the worker is accepting connections, and that DNS is correctly resolving.
 
-Canceling the transaction since it was involved in a distributed deadlock
+Cancelling the transaction since it was involved in a distributed deadlock
```
jonels-msft, line 27: I think this is an American vs British spelling difference. You've chosen the British form, which might not match the rest of the document as well.

reference/common_errors.rst
```diff
@@ -58,8 +49,7 @@ We can see this in action by distributing rows across worker nodes, and then run
 
 ::
 
-  ERROR:  40P01: canceling the transaction since it was involved in a distributed deadlock
-  LOCATION:  ProcessInterrupts, postgres.c:2988
+  ERROR:  canceling the transaction since it was involved in a distributed deadlock
```
jonels-msft, line 52: Has the error message changed to omit 40P01? If so, I wonder if there are other messages changed in a similar way...

use_cases/timeseries.rst
```diff
@@ -166,7 +167,7 @@ Now whenever maintenance runs, partitions older than a month are automatically d
 Archiving with Columnar Storage
 -------------------------------
 
-Some applications have data logically divides into a small updatable part and a
+Some applications have data logically divided into a small updatable part and a
```
jonels-msft, line 170: Oh, what I meant was they have "data that logically divides." My typo was dropping "that."

use_cases/timeseries.rst
```diff
@@ -213,7 +214,7 @@ Next, download sample data:
   wget http://examples.citusdata.com/github_archive/github_events-2015-01-01-{0..5}.csv.gz
   gzip -d github_events-2015-01-01-*.gz
 
-And load it:
+And load it (note that this data requires the database to have UTF8 encoding):
```
jonels-msft, line 217: There are other places where we load sample data with \COPY. Can you check if they need a similar warning?

jonels-msft:
@naisila thanks for the great changes. I left a few comments of things to change/check.

Also, I learned my lesson to add commas after "however." 😆

admin_guide/table_management.rst
naisila: Oh sorry, I hadn't realized that! I will put the dot back for now, and will let you know if I find a better way.

reference/common_errors.rst
naisila: Oh sorry about that, it's not that I prefer the British form, but I had never encountered "canceling" before. I will change it back to "canceling" in that case.

reference/common_errors.rst
naisila: Probably it does. Will check it as well

use_cases/timeseries.rst
naisila: Will change that.

use_cases/timeseries.rst
naisila: Will do.

jonels-msft:
Thank you!

PR #979, opened 2021-02-08
Foreign keys between reference and local tables

onurctirtir:


develop/api_guc.rst
```diff
@@ -277,6 +277,17 @@ Specifies whether to automatically propagate DDL changes from the coordinator to
 
   For a list of DDL propagation support, see :ref:`ddl_prop_support`.
 
+.. _enable_local_ref_fkeys:
+
+citus.enable_local_reference_table_foreign_keys (boolean)
+$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$
+
+Enables foreign keys between reference tables and local tables when coordinator is added to metadata (see :ref:`citus_add_node`).
+When enabled, Citus supports foreign keys between reference tables and local tables by applying a conversion to local tables (i.e adds local tables to Citus metadata) when defining such foreign keys.
```
onurctirtir, line 286: looked into deployment, seems that I should add some new lines like we did in `citus.all_modifications_commutative`

onderkalaci: Do we have any docs listing the limitations of such foreign keys (e.g., ON DELETE CASCADE not supported in some cases)?

develop/api_guc.rst
```diff
@@ -277,6 +277,17 @@ Specifies whether to automatically propagate DDL changes from the coordinator to
...
+Likely, Citus would automatically remove such tables from Citus metadata when user removes such foreign keys.
```
onderkalaci, line 287: `removes such fkey` -> `drops `

Also, maybe omit `likely` from the start?

>  which are mostly the same limitations that are already applying to reference tables and distributed tables.

Do we have a link for such limitations on the docs? If so, should we add here?

develop/api_udf.rst
```diff
@@ -101,15 +101,17 @@ Undistributing moves all data from shards back into a local table on the
...
+If `cascade_via_foreign_keys` is set to `false` (by default), you must drop its
+foreign key constraints, or those in other tables pointing to it before undistributing.
 
 A common use for this function is to :ref:`change_dist_col`.
 
 Arguments
 ************************
 
 **table_name:** Name of the distributed or reference table to undistribute.
+**cascade_via_foreign_keys:** (Optional) When set to "true," also undistributes other tables that are connected to input table via a foreign key graph.
```
onderkalaci, line 114: `via a foreign key graph.` seems a little technical. Can you rewrite the sentence in a little more user friendly way?

develop/api_guc.rst
onurctirtir: I couldn't find an exact place where we such limitations but in different sections of http://docs.citusdata.com/en/v9.5/develop/reference.html, maybe I should refer this section directly ?

Also, I think I should remove `distributed tables` from
`that are already applying to reference tables and distributed tables`
(so it becomes ~ `that are already applying to reference tables`)

develop/api_guc.rst
onurctirtir: I think I should mention such limitations in http://docs.citusdata.com/en/v9.5/develop/reference_ddl.html?highlight=Indices#adding-removing-constraints

onurctirtir: and give a reference to this section

develop/reference_ddl.rst
```diff
@@ -246,11 +246,15 @@ Foreign keys may be created in these situations:
...
 * between two local (non-distributed) tables,
 * between two reference tables,
+* between reference tables and local tables (by default enabled, via :ref:`enable_local_ref_fkeys`),
 * between two :ref:`colocated <colocation>` distributed tables when the key includes the distribution column, or
 * as a distributed table referencing a :ref:`reference table <reference_tables>`
 
 Foreign keys from reference tables to distributed tables are not supported.
 
+For the foreign keys between reference tables and local tables, any `referential action <https://www.postgresql.org/docs/current/ddl-constraints.html#DDL-CONSTRAINTS-FK>`_ is allowed for foreign keys from local tables to reference tables.
+However, Citus only supports `RESTRICT` action or no action for foreign keys from reference tables to local tables.
```
onderkalaci, line 256: I think mentioning not support for `ON DELETE/UPDATE CASCADE` could be more useful as they are much more commonly known by the users (as opposed to the meaning of RESTIRCT/NO ACTION`

develop/api_guc.rst
```diff
@@ -277,6 +277,20 @@ Specifies whether to automatically propagate DDL changes from the coordinator to
...
+
+citus.enable_local_reference_table_foreign_keys (boolean)
+$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$
+
+Enables foreign keys between reference tables and local tables when coordinator is added to metadata (see :ref:`citus_add_node`).
+
+When enabled, Citus supports foreign keys between reference tables and local tables by applying a conversion to local tables (i.e adds local tables to Citus metadata) when defining such foreign keys.
+Citus would automatically remove such tables from Citus metadata when user drops such foreign keys.
+
+Note that foreign keys between reference tables and local tables might bring some limitations to related local tables, which are mostly the same limitations that are already applying to reference tables (see :ref:`sql_reference`).
```
onderkalaci, line 290: Should we mention that dropping the foreign keys eliminates the limitations?

develop/api_udf.rst
```diff
@@ -101,8 +101,9 @@ Undistributing moves all data from shards back into a local table on the
 coordinator node (assuming the data can fit), then deletes the shards.
 
 Citus will not undistribute tables that have -- or are referenced by -- foreign
-keys. Before undistributing a table you must drop its foreign key constraints,
-or those in other tables pointing to it.
+keys unless `cascade_via_foreign_keys` option is set to `true`.
+If `cascade_via_foreign_keys` is set to `false` (by default), you must drop its
+foreign key constraints, or those in other tables pointing to it before undistributing.
```
onderkalaci, line 106: I think it'd be good to mention what happens when `cascade_via_foreign_keys ` = true.  That is a case where users might be careful. `local < ref < dist` case, we'd even undistribute the `dist`, right?

onurctirtir: true, I will explicitly mention that too

jonels-msft:
@onurctirtir let me know when you've covered everything you wanted to and verified its correctness with @onderkalaci. I'll then give the PR a style review.

develop/api_udf.rst
```diff
@@ -129,6 +137,25 @@ This example distributes a ``github_events`` table and then undistributes it.
   -- undo that and make it local again
   SELECT undistribute_table('github_events');
 
+
+.. _remove_local_tables_from_metadata:
+
+remove_local_tables_from_metadata
+$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$
+
+remove_local_tables_from_metadata() function removes local tables from Citus metadata
```
jonels-msft, line 146: This section doesn't tell me much. I have two questions:

* Why/when would someone need to call remove_local_tables_from_metadata() manually? Doesn't Citus take care of this when the foreign keys are dropped?
* What are the disadvantages of having tables stored in the metadata? Why should a user care?

develop/reference.rst
```diff
@@ -1,3 +1,5 @@
+.. _sql_reference:
```
jonels-msft, line 1: This section is so high-level, I'm not sure why we want to link here. Is there a more specific place you had in mind?

develop/api_guc.rst
```diff
@@ -277,6 +277,21 @@ Specifies whether to automatically propagate DDL changes from the coordinator to
 
   For a list of DDL propagation support, see :ref:`ddl_prop_support`.
 
+.. _enable_local_ref_fkeys:
+
+citus.enable_local_reference_table_foreign_keys (boolean)
+$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$
+
+Enables foreign keys between reference tables and local tables when coordinator is added to metadata (see :ref:`citus_add_node`).
```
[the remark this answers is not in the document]
jonels-msft: Oh also, are you sure we need to citus_add_node the coordinator? Does single-node Citus remove that requirement?

develop/reference_ddl.rst
```diff
@@ -246,11 +246,15 @@ Foreign keys may be created in these situations:
 
 * between two local (non-distributed) tables,
 * between two reference tables,
+* between reference tables and local tables (by default enabled, via :ref:`enable_local_ref_fkeys`),
 * between two :ref:`colocated <colocation>` distributed tables when the key includes the distribution column, or
 * as a distributed table referencing a :ref:`reference table <reference_tables>`
 
 Foreign keys from reference tables to distributed tables are not supported.
 
+For the foreign keys between reference tables and local tables, any `referential action <https://www.postgresql.org/docs/current/ddl-constraints.html#DDL-CONSTRAINTS-FK>`_ is allowed for foreign keys from local tables to reference tables.
```
jonels-msft, line 255: Might simplify like this:

```rst
Citus supports all `referential actions <foo>`_ on foreign keys from local to reference tables,
but does not support support ``ON DELETE/UPDATE CASCADE`` in the reverse direction (reference
to local).
```

jonels-msft (changes requested):
Added some suggestions for the phrasing, and some questions.

develop/api_guc.rst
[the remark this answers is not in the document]
onurctirtir: >Citus must convert the plain table into a "citus-local" table

In 9.5, we were calling naming such tables as "citus-local tables" but in 10.0, we decided to call them as "local tables added to metadata" since we actually don't want to draw a picture as if we introduced another table type. So, I think we should not use "citus-local" thing anywhere.

>Also, what are the limitations exactly?

The answer to this question would be a little bit long. This is because, for example any DDL limitation applying to reference tables (except creating/dropping triggers) do also apply to those local tables that are added to metadata.
But I can tell that we use more or less the same components to process queries & utility commands on those local tables with reference tables.

> Does single-node Citus remove that requirement?

In Citus 10.0, we implicitly add coordinator only if `pg_dist_node` is empty. So I think we should tell that user should explicitly add coordinator to metadata to be able to define such foreign keys.

develop/api_udf.rst
```diff
@@ -101,8 +101,9 @@ Undistributing moves all data from shards back into a local table on the
 coordinator node (assuming the data can fit), then deletes the shards.
 
 Citus will not undistribute tables that have -- or are referenced by -- foreign
-keys. Before undistributing a table you must drop its foreign key constraints,
-or those in other tables pointing to it.
+keys unless `cascade_via_foreign_keys` option is set to `true`.
```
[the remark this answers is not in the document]
onurctirtir: agreed, will address that

develop/api_udf.rst
```diff
@@ -111,6 +112,13 @@ Arguments
 
 **table_name:** Name of the distributed or reference table to undistribute.
 
+**cascade_via_foreign_keys:** (Optional) When set to "true," also undistributes other tables that are chained to input table with foreign keys.
```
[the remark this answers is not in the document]
onurctirtir: makes sense

develop/api_udf.rst
onurctirtir: >Doesn't Citus take care of this when the foreign keys are dropped?

Yes, Citus actually does that as long as `enable_local_reference_foreign_keys` guc is set to on (by default).
So we thought that it might to better to mention this udf in docs in case that user disabled the guc or something went wrong.
What do you think ?

develop/reference_ddl.rst
onurctirtir: makes sense

develop/reference.rst
onurctirtir: Then I think giving links to those two sections should be sufficient:
http://docs.citusdata.com/en/v9.5/develop/reference_workarounds.html & http://docs.citusdata.com/en/v9.5/develop/reference_ddl.html#modifying-tables

develop/api_udf.rst
jonels-msft: OK, so something like this might provide more motivation:

```rst
The remove_local_tables_from_metadata() function removes local tables
from Citus metadata that no longer need to be there. (See
:ref:`enable_local_ref_fkeys`.)

Usually if a local table is in Citus' metadata, there's a reason, such as
the existence of foreign keys between the table and a reference table.
However, if ``enable_local_reference_foreign_keys`` is disabled, Citus
will no longer manage metadata in that situation, and unnecessary
metadata can persist.
```

develop/api_guc.rst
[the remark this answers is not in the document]
jonels-msft: OK, want to add my suggested text but with the new terminology for citus-local?

jonels-msft:
@onurctirtir let me know when you've made the next round of updates and I can look at it one last time and probably merge it!

develop/api_guc.rst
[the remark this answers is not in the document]
onurctirtir: sure

onurctirtir:
> @onurctirtir let me know when you've made the next round of updates and I can look at it one last time and probably merge it!

addressed all your reviews now, thanks for all the feedback @jonels-msft

PR #982, opened 2021-02-12
Single node citus

jonels-msft:
Terminology and installation sections


jonels-msft:
In the interest of time I'm just going to merge this. We can make improvements later if needed.

PR #983, opened 2021-02-17
Fix the workaround given for local/dist table joins

onurctirtir:
Fixes #978

PR #985, opened 2021-02-25
Add doc for citus.local_table_join_policy

SaitTalhaNisanci:
Add doc for citus.local_table_join_policy

### DIFF
--- a/develop/api_guc.rst
+++ b/develop/api_guc.rst
@@ -225,51 +225,63 @@ Planner Configuration
 .. _local_table_join_policy:
 
 citus.local_table_join_policy (enum)
-$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$
+$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$
 
-This GUC determines how Citus moves data when doing a join between local and distributed tables. Customizing the join policy can help reduce the amount of data sent between worker nodes during the join. 
+This GUC determines how Citus moves data when doing a join between local and
+distributed tables. Customizing the join policy can help reduce the amount of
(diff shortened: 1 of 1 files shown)


develop/api_guc.rst
```diff
@@ -222,6 +222,19 @@ The default value for this GUC is 'on'.
 Planner Configuration
 ------------------------------------------------
 
+citus.local_table_join_policy (enum)
+$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$$
+
+Sets how citus should plan local and distributed table joins. There are four modes available:
```
onderkalaci, line 228: should we very briefly mention why would anyone need to change this? Also, maybe move `auto` to top as that's the default?

jonels-msft (changes requested):
More explanation of the terminology would be helpful. Especially the mechanics of conversion. I wasn't sure if that refers to rewriting the query in some way, or making some kind of temporary copy of tables across nodes...

develop/api_guc.rst
```diff
@@ -222,6 +222,19 @@ The default value for this GUC is 'on'.
 Planner Configuration
 ------------------------------------------------
 
+citus.local_table_join_policy (enum)
```
jonels-msft, line 225: We should add a link target above this, like:

```rst
.. _local_table_join_policy:
```

That way Onur can link to it from https://github.com/citusdata/citus_docs/pull/983

jonels-msft:
@SaitTalhaNisanci thanks for the clarification! I'll make a few edits on this branch and get it merged.

SaitTalhaNisanci:
@jonels-msft, let me know if there is anything we can clarify more because if that isn't clear to you, it will not be clear to users for sure :)

PR #989, opened 2021-03-02
Add citus.local_shared_pool_size

onderkalaci:
With Citus 10, we add a new GUC which implements connection throttling
for local node. This commit adds the GUC to the reference.


develop/api_guc.rst
```diff
@@ -406,6 +406,22 @@ the note below). The value -1 disables throttling.
   on the coordinator. This gives extra space for connections required for
   repartition queries on the workers.
 
+citus.local_shared_pool_size (integer)
+**************************************
+
+Specifies the maximum number of connections, across all simultaneous
+sessions, is allowed to make to the local node. This setting is relevant
+when the local node has shards, such as Citus on Single Node.
```
onderkalaci, line 414: What is the syntax to give a reference to`Citus on Single Node` section?

jonels-msft (dismissed):
Sorry @onderkalaci ! I added a review comment long ago, but never submitted the review!

I thought you weren't responding but it was my fault.

develop/api_guc.rst
jonels-msft:
```rst
:ref:`Citus on Single Node <development>`
```

or

```rst
:ref:`development`
```

if you want to use the text of the linked header, in this case `Single-Node Citus`.

ihalatci:
updated in the new docs, will be live soon

PR #993, opened 2021-03-02
Updates to workarounds section

marcocitus:
We support joins between local tables and reference tables in Citus 10, but are missing outer join limitations.


jonels-msft (changes requested):
Nice, the workarounds are going away!

Can you rebase this PR? I merged some other ones that updated a few of the workarounds. Instead of removing some of the workarounds, the PRs noted that older versions of citus had trouble, and the new easy way to do those things.

Once you rebase, I'll do a check to see whether other docs pages link to any of the removed workarounds.

marcocitus:
Rebased. I removed the local-distributed table join parts, because it did not seem like the right section for documenting the GUC.

PR #1000, opened 2021-04-05
Update Citus customer metrics data

clairegiordano:


jonels-msft:
Thanks Claire, I overlooked this one for a while.

PR #1003, opened 2021-05-12
Preserve keyboard focus on menu item during navigation

jonels-msft:
The downside is that it moves focus away from "Skip to main content."


mulander:
I'm closing due to age and conflicts, feel free to re-open.

PR #1015, opened 2021-09-27
Make clear that in PG14 we encode things as binary by default

JelteF:
In Citus 10.2 we fixed some bugs in binary result encoding and Postgres
fixed some bugs in PG14. So we decided to make binary encoding the
default for PG14. This updates the docs to say so.

It also removes a note about binary encoding resulting in more data.
This is true in some rare cases, but usually people binary encoding will
result in smaller data.


jonels-msft:
Thanks @JelteF

PR #1016, opened 2021-09-27
More citus_worker_stat_activity queries from Sai

jonels-msft:


jonels-msft:
Just wanted to see a preview

PR #1020, opened 2021-10-21
Second round of accessibility fixes

jonels-msft:


jonels-msft:
@missusmy thanks again!

PR #1026, opened 2022-01-07
Fixes invalid column_name minute for http_request in insert statements

gurkanindibay:
http_request table does not have minute column but in insert statements we are trying to use minute column . I changed it into date_trunc('minute', ingest_time) which was used in other insert statements in the same document


jonels-msft:
Thanks for noticing and fixing.

PR #1028, opened 2022-01-07
Fixes columnar recovery text

gurkanindibay:


admin_guide/table_management.rst
```diff
@@ -255,9 +255,8 @@ Gotchas
   so inserting one row per transaction will put single rows into their own
   stripes. Compression and performance of single row stripes will be worse than
   a row table. Always insert in bulk to a columnar table.
-* If you mess up and columnarize a bunch of tiny stripes, there is no way to
-  repair the table. The only fix is to create a new columnar table and copy
-  data from the original in one transaction:
+* If you mess up and columnarize a bunch of tiny stripes, you can use VACUUM command to be able to
```
jonels-msft, line 258: How about this:

> If you mess up and columnarize a bunch of tiny stripes, you can consolidate them
> and re-compress by running `VACUUM` on the table.

jonels-msft (changes requested):
What do you think about my suggestion? If you like it, can you make that change?

mulander:
@gurkanindibay is this still relevant? I'm closing, but please re-open if you are willing to continue with the suggestions provided by Joe.

PR #1030, opened 2022-02-11
switch to msft cookie consent

missusmy:


jonels-msft:
Please check:

* [ ] the popup appears for a user in the EU
* [ ] it causes no errors/warnings in the JS console for EU and non-EU users

PR #1039, opened 2022-03-30
fix installtion step

rikonaka:
Need to add the `shared_preload_libraries` before create extension.
```
root@citus3:~# sudo -i -u postgres psql -c "CREATE EXTENSION citus;"
ERROR:  Citus can only be loaded via shared_preload_libraries
HINT:  Add citus to shared_preload_libraries configuration variable in postgresql.conf in master and workers. Note that citus should be at the beginning of shared_preload_libraries.
```

### DIFF
--- a/installation/multi_node_debian.rst
+++ b/installation/multi_node_debian.rst
@@ -69,10 +69,14 @@ Before starting the database let's change its access permissions. By default the
   # and make it start automatically when computer does
   sudo update-rc.d postgresql enable
 
-You must add the Citus extension to **every database** you would like to use in a cluster. The following example adds the extension to the default database which is named `postgres`.
+You must add the Citus extension to **every database** you would like to use in a cluster. The following example adds the extension to the default database which is named `postgres`. To tell Postgres to use this extension you'll need to add it to a configuration variable called shared_preload_libraries:
 
 ::
 
+  # add shared_preload_libraries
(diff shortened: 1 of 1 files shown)


installation/multi_node_debian.rst
```diff
@@ -69,10 +69,14 @@ Before starting the database let's change its access permissions. By default the
   # and make it start automatically when computer does
   sudo update-rc.d postgresql enable
 
-You must add the Citus extension to **every database** you would like to use in a cluster. The following example adds the extension to the default database which is named `postgres`.
+You must add the Citus extension to **every database** you would like to use in a cluster. The following example adds the extension to the default database which is named `postgres`. To tell Postgres to use this extension you'll need to add it to a configuration variable called shared_preload_libraries:
 
 ::
 
+  # add shared_preload_libraries
+  sudo echo "shared_preload_libraries='citus'" >> /etc/postgresql/14/main/postgresql.conf
```
jonels-msft, line 77: Previously, in step 2 of "Steps to be executed on all nodes," we do this like so:

```bash
sudo pg_conftool 14 main set shared_preload_libraries citus
```

That worked when I tested it earlier, can you double check and see if it works for you?

jonels-msft (changes requested):
Thanks for your suggestion. Can you check and see if my comment below fixes the issue for you?

PR #1047, opened 2022-06-02
Fix wrong attribute names in API docs

paulslaby:


jonels-msft (changes requested):
Thanks for fixing the incorrect names!

Just one question about a seemingly accidental change.

PR #1050, opened 2022-06-15
Briefly mention local tables in metadata

onderkalaci:


develop/api_udf.rst
```diff
@@ -263,6 +263,47 @@ reference table
...
+tables (see :ref:`ddl` and :ref:`citus_sql_reference`).
+
+If you drop the foreign keys, Citus will automatically remove such local tables from metadata,
+which eliminates such limitations on those tables.
+
+Arguments
+************************
+
+**table_name:** Name of the table on the coordinator which needs to be accessed from workers.
+
```
onurctirtir, line 290: should we mention cascadeViaForeignKeys option ?

onderkalaci: When a table is added via this UDF, even if fkey is dropped, it stays on the metadata.

What do you think should we mention regarding cascadeViaForeignKeys?

onderkalaci: Ah, I forgot about `cascade_via_foreign_keys` parameter, now I got what you mean.

onurctirtir (approved):

jonels-msft:
Thanks!

I made a few edits and it looks good.

PR #1055, opened 2022-06-27
multi-node ubuntu changes

francisjodi:


installation/multi_node_debian.rst
```diff
@@ -59,15 +59,21 @@ Before starting the database let's change its access permissions. By default the
...
   Your DNS settings may differ. Also these settings are too permissive for some environments, see our notes about :ref:`worker_security`. The PostgreSQL manual `explains how <http://www.postgresql.org/docs/current/static/auth-pg-hba-conf.html>`_ to make them more restrictive.
+ Based on your DNS you can add another line with your IP address to allow the host unrestricted access, if you do not have an IP address that begins with 10.*.*.*. Otherwise add the line given for IPV4 address space. 
 
 **4. Start database servers, create Citus extension**
 
 ::
 
   # start the db server
   sudo service postgresql restart
+   or
```
jonels-msft, line 70: Can you explain more about why we present alternatives here? When would someone choose to restart postgresql vs postgresql@14-main ?

installation/multi_node_debian.rst
```diff
@@ -59,15 +59,21 @@ Before starting the database let's change its access permissions. By default the
...
 
   # start the db server
   sudo service postgresql restart
+   or 
+  sudo systemctl restart postgresql@14-main
   # and make it start automatically when computer does
   sudo update-rc.d postgresql enable
+  
+.. note:: 
+Ensure you are starting the right postges version. You can check this by running ``ps -ef | grep Postgres``
```
jonels-msft, line 76: Could the `ps -ef ...` be woven into the restart commands above? Ideally we could have the script itself detect the right service to restart.

jonels-msft (changes requested):
Thanks for your contribution!

Is there a way to make the command for starting the db server do the right thing automatically, rather than having the user grep through process list?

installation/multi_node_debian.rst
francisjodi: The idea is to ensure that someone does not start the server on another Postgres server. I had this issue where it started on a 9.5 server instead of 14. Someone would use postgres@14-main to ensure connection to postgres14, another person can use restart Postgres if they are sure their server is already on 14. Which would be the note about ps - ef. 
We could eliminate the option entirely and allow users to only use one.

installation/multi_node_debian.rst
francisjodi: Yes, that would be the best thing to do to ensure users are starting the right thing automatically.

jonels-msft (changes requested):
Hi @francisjodi, are you interested in submitting a change to automate the script's restart logic, by means of `ps -ef ...` or otherwise?

mulander:
This PR had no activity since Oct 2022. @francisjodi, please feel free to re-open if you are willing to continue with the changes requested by Joe.

PR #1059, opened 2022-08-29
Use authenticated access for multi-node Citus install

eradman:
Users may elect to use trust authentication to a local area network, but
there is probably no compelling reason to allow this, even for a trial
install.

* Merge worker_security section with debian and rhel installation
  instructions
* There should be no need to change the default rules for localhost


jonels-msft:
Thanks for helping to improve security,. I'll review this next week.

PR #1061, opened 2022-09-19
Fix missing double quotes

junland:
Came across this while reading the docs. Just missing a double quote at the end of the line for multi-node installations.


jonels-msft:
Good catch, thank you!

PR #1063, opened 2022-09-23
Add pdf and epub format

agedemenli:


jonels-msft:
Thanks for researching this and sending the PR!

PR #1065, opened 2022-10-15
Updates to link to Azure Cosmos DB for PostgreSQL, and to update how …

clairegiordano:
…we define/explain what Citus is.

### DIFF
--- a/_templates/page.html
+++ b/_templates/page.html
@@ -47,7 +47,7 @@
   {# Open Graph metadata #}
   <meta property="og:image" content="https://docs.citusdata.com/en/stable/_images/citus-docs-og-logo.png" />
   <meta property="og:title" content="{{ title }} - {{ docstitle }}" />
-  <meta property="og:description" content="Docs for the Citus extension to Postgres. Citus distributes your data &amp; queries across multiple nodes in a cluster, so your database can scale and your queries are fast." />
+  <meta property="og:description" content="Docs for the Citus database, an open source extension to Postgres. Citus gives you the Postgres you love at any scale—from a single node to a large distributed cluster. Available as 100% open source &amp; as a managed service with Azure Cosmos DB for PostgreSQL." />
 {% endblock %}
 
 {% block header %}
--- a/installation/citus_cloud.rst
+++ b/installation/citus_cloud.rst
@@ -1,23 +1,18 @@
 .. _multi_node_cloud:
 
-Managed Deployment
+Managed Service
 ==================
 
 One of the easiest ways to get started with Citus is to spin up a cluster in
-the cloud on Azure. Citus is available as a built-in deployment option in the
-`Azure Database for PostgreSQL - Hyperscale (Citus)
-<https://docs.microsoft.com/azure/postgresql/hyperscale/>`_ managed
-service.
+the cloud on Azure. The Citus managed service is available as
+`Azure Cosmos DB for PostgreSQL
+<https://learn.microsoft.com/azure/cosmos-db/postgresql/introduction/>`_, formerly known as Hyperscale (Citus) in Azure Database for PostgreSQL.
 
-To get started, you can take advantage of our Hyperscale (Citus) `quickstart
+To get started, you can take advantage of our Citus on Azure `quickstart
 documentation
-<https://docs.microsoft.com/azure/postgresql/quickstart-create-hyperscale-portal>`_,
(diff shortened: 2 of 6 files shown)


installation/citus_cloud.rst
```diff
@@ -1,23 +1,18 @@
...
-the cloud on Azure. Citus is available as a built-in deployment option in the
-`Azure Database for PostgreSQL - Hyperscale (Citus)
-<https://docs.microsoft.com/azure/postgresql/hyperscale/>`_ managed
-service.
+the cloud on Azure. The Citus managed service is available as
+`Azure Cosmos DB for PostgreSQL
+<https://learn.microsoft.com/azure/cosmos-db/postgresql/introduction/>`_, formerly known as Hyperscale (Citus) in Azure Database for PostgreSQL.
 
-To get started, you can take advantage of our Hyperscale (Citus) `quickstart
+To get started, you can take advantage of our Citus on Azure `quickstart
```
jonels-msft, line 11: Might be best to call it the "Azure Cosmos DB for PostgreSQL quickstart" rather than "Citus on Azure quickstart." The naming has been confusing and variable as it is. 😉

installation/citus_cloud.rst
```diff
@@ -1,23 +1,18 @@
...
 documentation
-<https://docs.microsoft.com/azure/postgresql/quickstart-create-hyperscale-portal>`_,
+<https://learn.microsoft.com/azure/cosmos-db/postgresql/quickstart-create-portal>`_,
 or go straight to the `Azure Portal
-<https://portal.azure.com/#create/Microsoft.PostgreSQLServer>`_ to provision a
-Postgres server. (And if you don’t yet have an Azure subscription, just `create
-a free Azure account <https://azure.microsoft.com/free/>`_ first.)
+<https://portal.azure.com/#view/Microsoft_Azure_DocumentDB/CreatePostgreSQL.ReactView>`_ to provision an
+Azure Cosmos DB for PostgreSQL cluster. (And if you don't yet have an Azure account, just `create
+a free Azure account <https://azure.microsoft.com/free/>`_—or take advantage of the `free trial <https://cosmos.azure.com/try/>`_.)
```
jonels-msft, line 17: Rather than all these options, what if we sent them to just the free trial? Sounds like it would be the most enticing option if I were a new user.

jonels-msft (changes requested):
Great changes, thanks! Appreciate you jumping in to send a PR.

Some questions and suggestions below.

_templates/page.html
```diff
@@ -47,7 +47,7 @@
   {# Open Graph metadata #}
   <meta property="og:image" content="https://docs.citusdata.com/en/stable/_images/citus-docs-og-logo.png" />
   <meta property="og:title" content="{{ title }} - {{ docstitle }}" />
-  <meta property="og:description" content="Docs for the Citus extension to Postgres. Citus distributes your data &amp; queries across multiple nodes in a cluster, so your database can scale and your queries are fast." />
+  <meta property="og:description" content="Docs for the Citus database, an open source extension to Postgres. Citus gives you the Postgres you love at any scale—from a single node to a large distributed cluster. Available as 100% open source &amp; as a managed service with Azure Cosmos DB for PostgreSQL." />
```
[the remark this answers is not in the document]
clairegiordano: For HTML descriptions I reco they be of 250-310 chars or so in length, but front loaded to the first 150 chars so that the most important bits show up in Google SERP. (Last I checked, Google sometimes chooses its own meta description anyway so fine tuning this too much can be a waste of time, as Google will serve up what it deems best for the description in SERP.) 

I will take a peek we should reorder the contents of the HTML description so it's more front loaded, but whether keeping the a longer description (to benefit slack users, etc) is worth it. Stay tuned.

installation/citus_cloud.rst
clairegiordano: will do, although I propose we keep the connection to "Citus on Azure" in the sentence while at the same time referring to the quickstart as you suggest. I believe the Citus on Azure connection is important because the new name doesn't make it intuitive that "Azure Cosmos DB for PostgreSQL" gets you to Citus. And the user is reading the Citus open source docs at this point.

_templates/page.html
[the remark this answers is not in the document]
clairegiordano: I updated the HTML description, it's still longer than you've requested but that's because I believe the extra description will serve us well in non-Google SERP scenarios. See what you think!

jonels-msft (approved):
Nice!

PR #1075, opened 2023-04-07
Remove "outer join with ref table" limitation

onurctirtir:
This should be a part of 11.2 page but I forgot about opening a PR for this.


develop/reference_workarounds.rst
```diff
@@ -32,30 +31,6 @@ multi-tenant use cases. <when_to_use_citus>`
...
-.. _pull_push_workaround:
-
-Work around limitations using CTEs
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-
-When a SQL query is unsupported, one way to work around it is using CTEs, which use what we call pull-push execution.
-
-.. code-block:: sql
-
-  SELECT * FROM ref LEFT JOIN dist USING (id) WHERE dist.value > 10;
```
jonels-msft, line 44: Although this particular query will be supported, are there any other types of queries that would require the CTE workaround? If so, we should keep this section and update the query inside.

jonels-msft (commented):
Glad the SQL support keeps improving.

I have a question about removing one of the sections.

jonels-msft:
Looks good, thanks for updating the example workaround too!

PR #1076, opened 2023-04-10
remove wrong optimized aggs

aykut-bozkurt:
Removes some of miswritten aggregates from optimized aggragetes.


jonels-msft:
Thanks!

PR #1091, opened 2023-05-03
Pin requests lib for urllib+ssl compatibility

jonels-msft:
Python environment issues on RTD


jonels-msft:
Built properly on Heroku and honored version restriction, so merging.

PR #1120, opened 2024-04-08
Update table_management.rst

thecatontheflat:


thecatontheflat:
@microsoft-github-policy-service agree

admin_guide/table_management.rst
```diff
@@ -252,18 +252,14 @@ Gotchas
...
+  and then back to :ref:`columnar` again like so:
 
   .. code-block:: postgresql
-
-    BEGIN;
-    CREATE TABLE foo_compacted (LIKE foo) USING columnar;
-    INSERT INTO foo_compacted SELECT * FROM foo;
-    DROP TABLE foo;
-    ALTER TABLE foo_compacted RENAME TO foo;
-    COMMIT;
```
JelteF, line 266: I think we should list both options. This one is faster with lots of data, because you only rewrite the table once. But the new approach is clearly easier from a command perspective, which makes it nice when messing around.

onurctirtir:
I think a table-rewrite (VACUUM FULL) should also work.

thecatontheflat:
@onurctirtir Since I'm not sure if that works too (potentially even a simpler solution), could you please let me know if I should add it to the docs?

thecatontheflat:
Updated the docs to include `VACUUM (FULL)` option as well as the original method

PR #1123, opened 2024-05-14
Change one thousand, to ten thousand

andyatkinson:
The batch size in the SQL is 10000, not 1000


mulander:
Merging in, thanks for spotting it! :)

PR #1124, opened 2024-08-17
Remove references to master_get_table_metadata

hanefi:
This UDF is removed in v11.0 release. This commit removes the references to this UDF from the documentation.

Needs to be backported to all branches starting from 11.0.

See https://github.com/citusdata/citus/pull/5389


mulander:
Thanks @hanefi, I'm going to merge this into master so it's at least present for the upcoming 13.0 release.

PR #1134, opened 2025-04-10
Update create_distributed_table with info about truncate_local_data_after_distributing_table

ivyazmitinov:
Currently, there is no mention about the way local data is handled after `create_distributed_table`. 
This PR adds necessary information, which may be crucial for some users.


hanefi (approved):
Thanks for your contribution. IMO we should merge and backport this to earlier releases of Citus as well.

PR #1135, opened 2025-07-03
fix: the foreign key violation on the documentation is fixed.

kuzeyardic:
I changed the places of the query, because it wasn't working due to a foreign key violation it is tested and working fine with the altered version.


kuzeyardic:
The violation is fixed.

ihalatci:
updated in the new docs, will be live soon

Issue #870, opened 2019-07-22
Please list shard_max_size as not relevant to hash partitioning


colton-citus:
Link:

http://docs.citusdata.com/en/v8.2/develop/api_guc.html#citus-shard-max-size-integer 

I had a customer get confused about shard_max_size and think it applied to hash partitioning. Please update this page to mention that it most likely doesn't apply to their use case and is deprecated and only relevant to append distribution.

R3gardless:
@onderkalaci 
Hi, would it be okay to remove the `shard_max_size` configuration from the documentation starting with version 11 and above?Thanks!

PR #1136, opened 2025-07-19
docs: remove citus.shard_max_size GUC

R3gardless:
According to #870,
Citus 11.0 and later no longer support the `citus.shard_max_size` GUC.

To avoid confusion, I believe it's better to remove this GUC from the documentation.

Also, I link `citus.shard_count` references to configuration documents.

If there's any issue with this change, please let me know.

Thanks!


ihalatci:
removed from the new docs, will be live soon

PR #1118, opened 2024-03-18
Add docs for citus_schema_move

onurctirtir:

PR #1138, opened 2025-08-27
Add docs for citus_schema_move in v12.1 docs

Green-Chan:
`citus_schema_move()` function was added in v12.1.0, but the docs for v12.1 is missing it.

This is a cherry-pick of #1118


ihalatci:

PR #1139, opened 2025-10-21
Simplify backup instruction for Citus metadata

ihalatci:
steps need to be run on all nodes, call out to old coordinator is misleading.


eaydingol (approved):

ihalatci:

PR #1140, opened 2025-12-25
Upgrades to Citus 14 and citus_stat_counters

onurctirtir:
Locally built by using "make docs" and everything is looking good.


ihalatci:

